PR #7, opened 2019-05-15
Add Http3Transport

pthatcherg:
Replace PooledHttpTransport and reference https://tools.ietf.org/html/draft-vvv-webtransport-http3-00

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -272,7 +272,7 @@ An Http3Transport is a WebTransport that provides QUIC streams and datagrams
 with slightly more overhead vs. a QuicTransport.  It has the advantage that HTTP
 and non-HTTP traffic can share the same network port and congestion control
 context, and it may be pooled with other transports such that the transport may
-be connected more quickly (by resuing an existing HTTP/3 connection).
+be connected more quickly (by reusing an existing HTTP/3 connection).
 
 ## Alternative designs considered
 
--- a/explainer.md
+++ b/explainer.md
@@ -208,7 +208,7 @@ async function readStreamUntilFin(stream) {
 }
 ```
 
-## Example of requesting over HTTP and receiving media pushed out-of-order and unreliably over the same network connection
+## Example of requesting over pooled HTTP and receiving media pushed out-of-order and unreliably over the same network connection
 
 ```javascript
 const mime = 'video/webm; codecs="opus, vp09.00.10.08"';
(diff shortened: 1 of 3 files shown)


explainer.md
```diff
@@ -253,10 +268,11 @@ A QuicTransport is a WebTransport that maps directly to QUIC streams and
...
 
-A PooledHttpTransport is a WebTransport that provides different subset of these
-capabilities depending on the underlying HTTP protocol (HTTP/3 providing the
-widest support).  It has the advantages that HTTP and non-HTTP traffic can share
-the same network port and congestion control context.
+An Http3Transport is a WebTransport that provides QUIC streams and datagrams
+with slightly more overhead vs. a QuicTransport.  It has the advantage that HTTP
+and non-HTTP traffic can share the same network port and congestion control
+context, and it may be pooled with other transports such that the transport may
+be connected more quickly (by resuing an existing HTTP/3 connection).
```
shampson: resuing --> reusing

index.html
```diff
@@ -1811,6 +1811,125 @@ <h3>Interface <dfn>ReceiveStream</dfn></h3>
...
+              internal slot representing a sequence of <code><a>OutgoingStream</a></code>
+              objects, initialized to empty.
+            </li>
+            <li>Let <var>transport</var> have a <dfn>[[\IncomingStreams]]</dfn>
+              internal slot representing a sequence of <code><a>IncomingStream</a></code>
+              objects, initialized to empty.
+            </li>
+            <li>
+                Let <var>transport</var> have a <dfn>[[\WebTransportState]]</dfn>
+                internal slot, initialized to <code>"connecting"</code>.
```
shampson: Is the "connecting" redundant? You mentioned it above that the transport object has a "connecting" state.

index.html
```diff
@@ -1811,6 +1811,125 @@ <h3>Interface <dfn>ReceiveStream</dfn></h3>
...
+            </li>
+            <li>Let <var>transport</var> have a <dfn>[[\ReceivedDatagrams]]</dfn>
+                internal slot representing a queue of <code>Uint8Array</code>, initialized to empty.
+            </li>
+            <li>Let <var>transport</var> have a <dfn>[[\ReceiveDatagramsPromise]]</dfn>
```
shampson: What about mentioning the ReadyToSendDatagramsPromise? This isn't included in the QuicTransport either. We also don't mention these internal slots in the methods which should probably be done.

index.html
```diff
@@ -1811,6 +1811,125 @@ <h3>Interface <dfn>ReceiveStream</dfn></h3>
...
+                internal slot representing a <code>Promise&lt;sequence&lt;Uint8Array&gt;&gt;?</code>,
+                initialized to null.
+            <li>Run these steps in parallel:
+              <ol>
+                <li>
+                  Either establish an HTTP/3 connection or reuse an existing HTTP/3
```
shampson: Do we need to specify how we establish an HTTP/3 connection or reuse one here?

explainer.md
pthatcherg: Fixed

index.html
pthatcherg: I'll punt to the IETF doc :)

index.html
pthatcherg: Good catch; removed

index.html
pthatcherg: Let's make it an issue to follow up on

PR #16, opened 2019-06-19
Update the explainer to use WHATWG streams

pthatcherg:


explainer.md
```diff
@@ -217,13 +187,16 @@ mediaSource.onsourceopen = (e) => {
...
   
   const transport = new Http3Transport("/video");
-  while (transport.state == "connecting" || transport.state == "connected") {
-    const datagrams = await transport.receiveDatagrams();
-    for (let data of datagrams) {
-      if (data) {
-        const chunk = ccontainerizeMedia(data);
-        sourceBuffer.appendBuffer(chunk);
+  if (transport) {
+    await fetch('http://example.com/babyshark');
```
steveanton: Is this intended to be used?

explainer.md
```diff
@@ -117,45 +119,20 @@ mediaSource.onsourceopen = (e) => {
   // App-specific request
   const mediaRequest = Uint8Array.from([1, 2, 3, 4]);
   const requestStream = await transport.createSendStream();
-  requestStream.write(mediaRequest);
+  const writer = requestStream.writable.getWriter();
+  await writer.ready();
```
steveanton: I don't think this is needed. If it is, then it should be `await writer.ready` (property not a method).

explainer.md
```diff
@@ -117,45 +119,20 @@ mediaSource.onsourceopen = (e) => {
...
   transport.onreceivestream = (e) => {
-    const size = await readStreamUint32(e.stream);
-    if (!size || size > 512000) {
-      return;
-    }
-    const chunk = await readStreamBytes(e.stream, size);
-    if (!!chunk) {
-      sourceBuffer.appendBuffer(chunk);
+    const reader = e.stream.readable.getReader();
+    while (!reader.closed()) {
```
steveanton: `closed` is a promise, so instead I think instead you want something like:

```
while (true) {
  const {value: chunk, done} = await reader.read();
  if (done) break;
  sourceBuffer.appendBuffer(chunk);
}
```

explainer.md
```diff
@@ -217,13 +187,16 @@ mediaSource.onsourceopen = (e) => {
...
-    const datagrams = await transport.receiveDatagrams();
-    for (let data of datagrams) {
-      if (data) {
-        const chunk = ccontainerizeMedia(data);
-        sourceBuffer.appendBuffer(chunk);
+  if (transport) {
+    await fetch('http://example.com/babyshark');
+    const reader = transport.receiveDatagrams.getReader();
+    while (true) {
+      const {datagram, done} = await reader.read();
```
steveanton: `const {value: datagram, done} = await ...`

explainer.md
```diff
@@ -117,45 +119,20 @@ mediaSource.onsourceopen = (e) => {
...
 
   transport.onreceivestream = (e) => {
-    const size = await readStreamUint32(e.stream);
-    if (!size || size > 512000) {
-      return;
-    }
-    const chunk = await readStreamBytes(e.stream, size);
-    if (!!chunk) {
-      sourceBuffer.appendBuffer(chunk);
+    const reader = e.stream.readable.getReader();
```
shampson: Do you want to still read the size & use a ReadableStreamBYOBReader?
https://streams.spec.whatwg.org/#byob-reader-read

explainer.md
```diff
@@ -182,20 +159,13 @@ transport.onbidirectionalstream = (e) => {
...
-    const buffer = new Uint8Array(stream.readableAmount);
-    const read = stream.readInto(readBuffer);
-    buffers.push(buffer);
-    bufferedSize += read.amount;
-    finished = read.finished;
-  }
-  if (!finished) {
-    // Stream was aborted
-    return null;
+  while (!reader.closed()) {
```
shampson: Steve's comment about closed() being a promise applies here as well.

shampson:
Did you want to incorporate a WritableStream in the sendDatagram() example as well?

lgrahl:
> Did you want to incorporate a WritableStream in the sendDatagram() example as well?

If there is no strict (MTU-like) size limitation in the abstract API, this should also use a stream, yes.

explainer.md
pthatcherg: Removed

explainer.md
pthatcherg: Done

explainer.md
pthatcherg: Done

explainer.md
pthatcherg: Do you mean the result?  The idea here is that the result comes in datagrams, not in the body of the response.

explainer.md
pthatcherg: Done

explainer.md
pthatcherg: I'd rather keep the example simple.

explainer.md
```diff
@@ -74,14 +74,15 @@ encrypted and congestion-controlled communication.
...
   // App-specific encoded game state
   const gameState = getGameState();
   const encodeGameState = encodeGameState(gameState);
-  try {
-    transport.sendDatagram(encodedGameState);
-  } catch(err) {
+  if (datagramWritableStream.ready) {
+    datagramWritableStream.getWriter().write(encodedGameState);
+  } else {
     // Ignore; just keep sending anyway
```
shampson: I think you want to remove this else and the comment. The comment is old and was referring to sending without being ready.

explainer.md
```diff
@@ -74,14 +74,15 @@ encrypted and congestion-controlled communication.
...
+const datagramWritableStream = transport.sendDatagrams();
 
 setInterval(() => {
   // App-specific encoded game state
   const gameState = getGameState();
   const encodeGameState = encodeGameState(gameState);
-  try {
-    transport.sendDatagram(encodedGameState);
-  } catch(err) {
+  if (datagramWritableStream.ready) {
```
shampson: ready is a promise attribute. I think there's two ways to do this.
1. await the ready promise and if it resolves before a delta 100 ms you write(). 
2. don't write unless the last promise has resolved 

2. is more simple but 1. gives more accuracy (makes sure you aren't buffering a datagram).
https://streams.spec.whatwg.org/#default-writer-ready

explainer.md
```diff
@@ -94,12 +95,14 @@ const host = 'example.com';
...
 
-setInterval(() => {
+setInterval(async () => {
   // App-specific encoded game state
   const gameState = getGameState();
   const encodeGameState = encodeGameState(gameState);
   const stream = await transport.createSendStream();
-  stream.write({data: encodedGameState, finished: true});
+  const writer = stream.writable.getWriter();
+  writer.write(encodedGameState);
```
shampson: Do you want to do something similar here for waiting for ready?

explainer.md
```diff
@@ -211,13 +216,20 @@ mediaSource.onsourceopen = async (e) => {
...
-mediaSource.onsourceopen = (e) => {
+mediaSource.onsourceopen = async (e) => {
   const sourceBuffer = mediaSource.addSourceBuffer(mime);
   
   const transport = new Http3Transport("https://example.com/video");
-  transport.onunidirectionalstream = async (e) => {
-    const chunk = await readStreamUntilFin(e.stream);
-    sourceBuffer.appendBuffer(chunk);
+  const streamsReader = transport.receiveStreams().getReader();
+  while (true) {
```
shampson: Instead of using a while loop for this would it be cleaner to use a function? I'm not sure, so I'll let you decide.

async function appendDataToBuffer(streamResult) {
  if (streamResult.done) {
    return;
  }
  (async () => {
    const chunk = await readStreamUntilFin(e.stream);
    sourceBuffer.appendBuffer(chunk);
  })
  streamsReader.read().then(appendDataToBuffer(result);
}
streamsReader.read().then(appendDataToBuffer(result));

explainer.md
pthatcherg: Done

explainer.md
pthatcherg: I'm not sure if it's needed, but I'll add it for now.

explainer.md
pthatcherg: I like the idea of the function.  I moved the while loop to it.

shampson:
lgtm modulo the last comment about the .ready.

explainer.md
pthatcherg: Why don't we just say that if you call write when there is back pressure you get dropped (unlike a stream, where you get queued)?  Then we can just call .write() without checking .ready.

Issue #21, opened 2019-06-22
Stats for congestion control and bandwidth estimation


vasilvv:
One of the use cases we are interested in is media streaming; when streaming media, the application can often decide to change the amount of data it sends based on how much bandwidth it expects to have available.  Since all of the transports we define are congestion-controlled, we already naturally have to make some form of a guess regarding how much data the path can handle (even though it can be as rudimentary as CWND/RTT).

We should provide an API that lets the underlying transport library expose this kind of data to the Web application.  My intuitive idea would be `estimateBytesAvailable(time)`, or even `estimateBytesAvailable(time, p)`, for models that accept the target probability of not oversending (for fancier models that accept such parameter, e.g. [Sprout](https://www.usenix.org/system/files/conference/nsdi13/nsdi13-final113.pdf)).

cc @keithw, who is an expert on this topic and might have a much better idea of how this API should look.

pthatcherg:
I agree that providing information about the transport is important for the application to adapt to network conditions.  The app may want estimated bandwidth and RTT.  Currently the impl in Chromium provides three pieces of info easily in the C++ that could be exposed up to JS:

https://cs.chromium.org/chromium/src/net/third_party/quiche/src/quic/quartc/quartc_session.h?g=0&l=130

They are: bandwidth_estimate, pacing_rate (the rate at which the congestion controller wants to send right now, which may be higher or lower than the estimated bandwidth), and RTT, and they seem to be fired whenever we receive an ack:

https://cs.chromium.org/chromium/src/net/third_party/quiche/src/quic/core/quic_sent_packet_manager.cc?g=0&l=297

Every ack might be a bit too frequent, but perhaps an event throttled to some reasonable frequency would make sense (rather than polling).

keithw:
Thank you for looping me in. I've skimmed the current WebTransport draft (at least the DatagramTransport-related parts) and pondered your question for a bit.

I think what you're trying to do here is super-interesting, and the `readyToSendDatagram` API where the application effectively gets an upcall when the congestion controller is willing to send one new datagram is a cool idea. The challenge is that of course often datagrams are not independent to the application, and an app that cares about the latency of individual messages doesn't want to have a 5-datagram-long message and get blocked after sending the first 4 datagrams. (If it knew that was going to happen, the app would have decided to send a shorter message.)

[...]

PR #43, opened 2019-07-23
Adds getStats() method to QuicTransportBase.

shampson:
This basically is just reusing what was implemented in the RTCQuicTransport origin trial. Feel free to suggest removing/editing certain stats. The main motivation for this PR is:
1) Get basic stats in the spec
2) Have a place to point for at least some basic congestion control information exposed in the API.

See issues:
- https://github.com/w3c/webrtc-quic/issues/66 (Initial suggestion for P2P QUIC transport stats)
- https://github.com/WICG/web-transport/issues/21 (bandwidth prediction)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -535,7 +535,7 @@ dictionary QuicTransportStats {
   unsigned long numIncomingStreamsCreated;
   unsigned long long bytesReceived;
   unsigned long long packetsReceived;
-  DOMHighResTimeStamp minRttUs;
+  DOMHighResTimeStamp minRtt;
   unsigned long numReceivedDatagramsDropped;
 };
 </pre>
@@ -546,22 +546,22 @@ The dictionary SHALL have the following attributes:
 :: The `timestamp` for when the stats are gathered, relative to the
 UNIX epoch (Jan 1, 1970, UTC).
 : <dfn for="QuicTransportStats" dict-member>bytesSent</dfn>
-:: The amount of bytes sent on the QUIC connection, including retransmissions.
+:: The number of bytes sent on the QUIC connection, including retransmissions.
 Does not include UDP or any other outer framing.
 : <dfn for="QuicTransportStats" dict-member>packetsSent</dfn>
-:: The amount of packets sent on the QUIC connection, including retransmissions.
+:: The number of packets sent on the QUIC connection, including retransmissions.
 : <dfn for="QuicTransportStats" dict-member>numOutgoingStreamsCreated</dfn>
-:: The amount of outgoing QUIC streams created on the QUIC connection.
+:: The number of outgoing QUIC streams created on the QUIC connection.
 : <dfn for="QuicTransportStats" dict-member>numIncomingStreamsCreated</dfn>
-:: The amount of incoming QUIC streams created on the QUIC connection.
+:: The number of incoming QUIC streams created on the QUIC connection.
 : <dfn for="QuicTransportStats" dict-member>bytesReceived</dfn>
-:: The amount of total bytes received on the QUIC connection, including
+:: The number of total bytes received on the QUIC connection, including
 duplicate data for streams. Does not include UDP or any other outer framing.
 : <dfn for="QuicTransportStats" dict-member>packetsReceived</dfn>
-:: The amount of total packets received on the QUIC connection, including
+:: The number of total packets received on the QUIC connection, including
 packets that were not processable.
-: <dfn for="QuicTransportStats" dict-member>minRttUs</dfn>
-:: The minimum RTT.
+: <dfn for="QuicTransportStats" dict-member>minRtt</dfn>
+:: The minimum RTT observed on the entire connection.
 : <dfn for="QuicTransportStats" dict-member>numReceivedDatagramsDropped</dfn>
 :: The number of datagrams that were dropped, due to too many datagrams buffered
 between calls to {{receiveDatagrams()}}.
--- a/index.bs
+++ b/index.bs
@@ -478,7 +478,7 @@ The dictionary SHALL have the following attributes:
 
 # `QuicTransportBase` Interface #  {#quic-transport-base}
 
-The `QuicTransportBase` is the base interface for {{QuicTransport}}. Most of
+The `QuicTransportBase` is the base for {{QuicTransport}}. Most of
 the functionality of a `QuicTransport` is in the base class to allow for other
 subclasses (such as a p2p variant) to share the same interface.
 
@@ -535,8 +535,7 @@ dictionary QuicTransportStats {
   unsigned long numIncomingStreamsCreated;
   unsigned long long bytesReceived;
   unsigned long long packetsReceived;
-  unsigned long long minRttUs;
-  unsigned long estimatedBandwidthBps;
+  DOMHighResTimeStamp minRttUs;
   unsigned long numReceivedDatagramsDropped;
 };
 </pre>
@@ -546,10 +545,9 @@ The dictionary SHALL have the following attributes:
 : <dfn for="QuicTransportStats" dict-member>timestamp</dfn>
 :: The `timestamp` for when the stats are gathered, relative to the
 UNIX epoch (Jan 1, 1970, UTC).
-
 : <dfn for="QuicTransportStats" dict-member>bytesSent</dfn>
 :: The amount of bytes sent on the QUIC connection, including retransmissions.
-Does not include UDP/IP/ethernet framing.
+Does not include UDP or any other outer framing.
 : <dfn for="QuicTransportStats" dict-member>packetsSent</dfn>
 :: The amount of packets sent on the QUIC connection, including retransmissions.
 : <dfn for="QuicTransportStats" dict-member>numOutgoingStreamsCreated</dfn>
@@ -558,15 +556,12 @@ Does not include UDP/IP/ethernet framing.
 :: The amount of incoming QUIC streams created on the QUIC connection.
 : <dfn for="QuicTransportStats" dict-member>bytesReceived</dfn>
 :: The amount of total bytes received on the QUIC connection, including
-duplicate data for streams. Does not include UDP/IP/ethernet framing.
+duplicate data for streams. Does not include UDP or any other outer framing.
 : <dfn for="QuicTransportStats" dict-member>packetsReceived</dfn>
 :: The amount of total packets received on the QUIC connection, including
 packets that were not processable.
 : <dfn for="QuicTransportStats" dict-member>minRttUs</dfn>
-:: The minimum RTT in microseconds.
-: <dfn for="QuicTransportStats" dict-member>estimatedBandwidthBps</dfn>
-:: The estimated bandwidth in bits per second determined by the QUIC
-connection.
+:: The minimum RTT.
 : <dfn for="QuicTransportStats" dict-member>numReceivedDatagramsDropped</dfn>
 :: The number of datagrams that were dropped, due to too many datagrams buffered
 between calls to {{receiveDatagrams()}}.
--- a/index.html
+++ b/index.html
@@ -2034,7 +2034,7 @@ frame.</p>
      <p>The reason for closing the <code class="idl"><a data-link-type="idl" href="#webtransport" id="ref-for-webtransport④">WebTransport</a></code>.</p>
    </dl>
    <h2 class="heading settled" data-level="8" id="quic-transport-base"><span class="secno">8. </span><span class="content"><code>QuicTransportBase</code> Interface</span><a class="self-link" href="#quic-transport-base"></a></h2>
-   <p>The <code>QuicTransportBase</code> is the base interface for <code class="idl"><a data-link-type="idl" href="#quictransport" id="ref-for-quictransport">QuicTransport</a></code>. Most of
+   <p>The <code>QuicTransportBase</code> is the base for <code class="idl"><a data-link-type="idl" href="#quictransport" id="ref-for-quictransport">QuicTransport</a></code>. Most of
 the functionality of a <code>QuicTransport</code> is in the base class to allow for other
 subclasses (such as a p2p variant) to share the same interface.</p>
    <h3 class="heading settled" data-level="8.1" id="quic-transport-base-overview"><span class="secno">8.1. </span><span class="content">Overview</span><a class="self-link" href="#quic-transport-base-overview"></a></h3>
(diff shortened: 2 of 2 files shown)


shampson:
It seems I don't have contributor's access. Could someone help me out? The error says:
_"shampson did not make IPR commitments for this group."_

shampson:
@pthatcherg @aboba PTAL. I don't have access to add you as reviewers unfortunately. Thanks!

aboba:
The approach used in [ORTC](http://draft.ortc.org/#statistics-api) for handling per-object stats is a bit different. Using that approach, we would have something like this (could also be used if stats were desired for the IceTransport or QuicStream). 

interface QuicTransport : RTCStatsProvider {
}

interface RTCStatsProvider {
    Promise<RTCStatsReport> getStats();
};

[Exposed=Window]
interface RTCStatsReport {
    readonly maplike<DOMString, object>;
};

dictionary RTCStats {
    DOMHighResTimeStamp timestamp;
    RTCStatsType        type;
    DOMString           id;
};

A new RTCStatsType Enum would probably be needed, such as "quic-transport".

shampson:
@aboba I would prefer not to use this approach for a few reasons:
- It's a much more complex model targeted at monolithic API surfaces (like PeerConnection) that include a lot of other sub-objects - streams, tracks, transports, senders, receivers, etc.
- It's a ORTC/WebRTC specific approach with lots of "RTC"
- I don't see a reason not to use the simple approach. What's wrong with the following code for getting stats on different objects?
```javascript
const transport = new QuicTransport(host, port);
const stream = await transport.createSendStream();
const streamStats = await stream.getStats();
const transportStats = await transport.getStats();
```

I'm open to doing it the ORTC way but I'd like to better understand the benefits and why we would want to do this.

aboba:
The major difference from what you are proposing is that `getStats()` returns an `RTCStatsReport` and QuicTransportStats inherits the timestamp, id and type from `RTCStats`.  This allows the same format to be used for `getStats()` called on any object, potentially making it easier for services such as callstats.io to process the stats.  There is no assumption that there is an `RTCPeerConnection` - each object would have its own `getStats()` method.

shampson:
I was more calling out `RTCPeerConnection` as an example of an object where this stats model you’re suggesting makes more sense, because it has a bunch of underlying objects associated with it that you would also like to pull the stats from. We could use this model and remove the “RTC” stuff.

So the main advantage of the ORTC/WebRTC model is the self identification of the stats, that allows pulling the same stats from multiple objects. In our case that would mean you could get QUIC stream stats, ICE stats, as well as QUIC connection stats all from a QuicTranpsort.getStats() call.

My opinion is that this model is over complicated for the WebTransport/QuicTransport case, because what we’re trying to do is expose a lower level object. I think pulling the stats from the objects directly is sufficient.

Maybe @vr000m would like to weigh in.

The main goal of this PR is to get some initial stats we can reference for discussions in open issues like the one around congestion control. WDYT of landing it then continuing the discussion around the stats model in a separate issue?

aboba:
In the ORTC model, each object has a `getStats()` method and calling this returns the statistics relevant to that object.  However, the type may still be useful, in case multiple reports are to be returned for a given object, or the statistics are stored for later retrieval.  As an example, in WebRTC-ICE the RTCIceTransport does not have a `getStats` method, but the ORTC RTCIceTransport does have such a method because it inherits from RTCStatsProvider.

index.html
```diff
@@ -1017,6 +1017,206 @@ <h3>Interface Definition</h3>
...
+          </div>
+        </dd>
+    </section>
+    <section id="quictransportstats*">
+      <h3><dfn>QuicTransportStats</dfn> Dictionary</h3>
+      <p>The <code>QuicTransportStats</code> dictionary includes information
+      QUIC connection level stats.
+      <div>
+        <pre class="idl">
+dictionary QuicTransportStats {
```
vasilvv: This is a rich set of stats, and it's unclear to me how much of them we want.  I assume this is roughly the QuicConnectionStats object that gQUIC exposes.  I feel like we should be more conservative about what we expose, since it's not necessarily the case that other browsers' QUIC implementation tracks all of those.

index.html
```diff
@@ -1017,6 +1017,206 @@ <h3>Interface Definition</h3>
...
+            "idlMemberType"><a>DOMHighResTimeStamp</a></span></dt>
+            <dd>
+              <p>The <code>timestamp</code> for when the stats are gathered, relative to the
+              UNIX epoch (Jan 1, 1970, UTC)</p>
+            </dd>
+            <dt><dfn><code>bytesSent</code></dfn> of type <span class=
+            "idlMemberType"><a>unsigned long long</a></span></dt>
+            <dd>
+              <p>The amount of bytes sent on the QUIC connection, including
+              retransmissions.</p>
```
vasilvv: We should be more precise about how we count bytes.  Does this include UDP/IP/Ethernet framing?

index.html
```diff
@@ -1017,6 +1017,206 @@ <h3>Interface Definition</h3>
...
+            <dt><dfn><code>packetsLost</code></dfn> of type <span class=
+            "idlMemberType"><a>unsigned long long</a></span></dt>
+            <dd>
+              <p>The amount of packets abandoned as lost determined by the loss detection
+              algorithm.</p>
+            </dd>
+            <dt><dfn><code>packetsDropped</code></dfn> of type <span class=
+            "idlMemberType"><a>unsigned long long</a></span></dt>
+            <dd>
+              <p>The amount of packets dropped.</p>
```
vasilvv: What's meant by "dropped" here?

aboba:
@shampson Can you get nominated as a WICG member so your PRs can pass the IPR check?

index.html
shampson: Yes you're correct, this is what I called out in the first PR comment - these are roughly the stats from the origin trial, which comes from gQUIC stats, & I'm happy to remove/edit. We could just keep a minimum for now and add later. WDYT of the following?

```
DOMHighResTimeStamp timestamp;
unsigned long long bytesSent;
unsigned long numOutgoingStreamsCreated;
unsigned long numIncomingStreamsCreated;
unsigned long long bytesReceived;
unsigned long estimatedBandwidthBps;
unsigned long long minRttUs;
unsigned long numReceivedDatagramsDropped;
```

index.html
shampson: You're right we should specify this. I don't think this includes overhead. Comes from the WriteResult in quiche, tested here:
https://cs.chromium.org/chromium/src/net/quic/platform/impl/quic_socket_utils_test.cc?g=0&l=137

I updated the comment. Correct me if I'm wrong.

index.html
shampson: Ignored packets. In quiche this is with unrecognized packet headers, or connection IDs. Could change to packetsIgnored?

shampson:
@aboba Victor helped me resolve the issue, thank you. Thanks for looking over the PR @vasilvv. Let's decide on the minimum stats we want to add now. I made a suggestion in a comment. I'd also like to discuss as a group the need for the ORTC model for stats.

p.s. I snuck in the `numReceivedDatagramsDropped` stat for datagrams that are dropped, as discussed here: [Issue 124- Alternatively using stats for dropped datagrams](https://github.com/w3c/webrtc-quic/issues/124)

vr000m:
At the high level,

1. I do not want to traverse getStats to figure out the relationship, if that was explicit somewhere then the getStats could be called on the individual objects. That would avoid get back a mega object.
2. want to getStats through the lifecycle of the object.

shampson:
I created an [issue](https://github.com/WICG/web-transport/issues/45) for the stats API design. I think this PR is valuable to land without this discussion being resolved, so that we have a place to point to for other discussions like congestion control & datagram feedback. 

I'd be in favor of landing the simple stats API now (with a subset of the current metrics), then following up in the issue. How does that sound?

vr000m:
I am okay with the way forward as long as we are open to make breaking changes when we have that discussion

shampson:
Breaking changes are fine to make right now and will be until the spec becomes more mature and is begins the process to ship. Development hasn't even begun for the WebTransport API in an origin trial. Since the RTCQuicTransport API shares the same base object, it is dependent on this as well before shipping.

I'd like to resolve this before we have a WebTransport origin trial, which is intended as a way to experiment with APIs that **don't** become baked into the browser.

index.html
shampson: Removed this.

index.html
shampson: I updated to include a subset of the original PR. I'm resolving this conversation, please comment if you think there are others that should be removed. Thanks!

shampson:
@aboba are you okay landing this change, given the conversation about statistics API requirements has been moved out in to a separate issue? 

If so, could you and @vasilvv please give this a look and see if the given statistics seems reasonable? Thanks!

aboba:
@shampson Yes, am ok with landing this change and creating a new issue for API design.

shampson:
@vasilvv would you mind looking over this again?

index.html
```diff
@@ -1017,6 +1016,120 @@ <h3>Interface Definition</h3>
...
+            <dt><dfn><code>numIncomingStreamsCreated</code></dfn> of type <span class=
+            "idlMemberType"><a>unsigned long</a></span></dt>
+            <dd>
+              <p>The amount of incoming QUIC streams created on the QUIC connection.</p>
+            </dd>
+            <dt><dfn><code>bytesReceived</code></dfn> of type <span class=
+            "idlMemberType"><a>unsigned long long</a></span></dt>
+            <dd>
+              <p>The amount of total bytes received on the QUIC connection, including
+              duplicate data for streams. Does not include UDP/IP/ethernet framing.</p>
```
vasilvv: Same as above.

index.html
```diff
@@ -1017,6 +1016,120 @@ <h3>Interface Definition</h3>
...
+            <dt><dfn><code>packetsReceived</code></dfn> of type <span class=
+            "idlMemberType"><a>unsigned long long</a></span></dt>
+            <dd>
+              <p>The amount of total packets received on the QUIC connection, including
+              packets that were not processable.</p>
+            </dd>
+            <dt><dfn><code>minRttUs</code></dfn> of type <span class=
+            "idlMemberType"><a>unsigned long long</a></span></dt>
+            <dd>
+              <p>The minimum RTT in microseconds.</p>
```
vasilvv: What time units we use elsewhere in the spec?  `timestamp` is a `DOMHighResTimeStamp`, which is milliseconds represented as `double`.  We should try to be consistent.

index.html
```diff
@@ -1017,6 +1016,120 @@ <h3>Interface Definition</h3>
...
+            </dd>
+            <dt><dfn><code>estimatedBandwidthBps</code></dfn> of type <span class=
```
vasilvv: I'd remove this for now, since "bandwidth estimate" is ambiguous (at least until we figure out #21).

vasilvv: Alternatively, we can provide a reference to pacing rate, https://tools.ietf.org/html/draft-ietf-quic-recovery-22#section-7.8

shampson:
Okay, I've resolved all outstanding comments and merged with the bikshed PR, so the change is in index.bs now. @vasilvv thanks a lot for updating the spec to bikeshed, the merge was clumsy, but  it's so much easier to edit!

I don't have the ability to merge this, could an owner go ahead and land the PR? @pthatcherg @aboba 
Thanks!

index.bs
```diff
@@ -504,6 +505,67 @@ QuicTransportBase includes DatagramTransport;
...
+<pre class="idl">
+dictionary QuicTransportStats {
+  DOMHighResTimeStamp timestamp;
+  unsigned long long bytesSent;
+  unsigned long long packetsSent;
+  unsigned long numOutgoingStreamsCreated;
+  unsigned long numIncomingStreamsCreated;
+  unsigned long long bytesReceived;
+  unsigned long long packetsReceived;
+  DOMHighResTimeStamp minRttUs;
```
pthatcherg: Does this need the "Us" suffix since it has the DOMHighResTimeStamp type?  "timestamp" doesn't have it.

index.bs
```diff
@@ -504,6 +505,67 @@ QuicTransportBase includes DatagramTransport;
...
+: <dfn for="QuicTransportStats" dict-member>numIncomingStreamsCreated</dfn>
+:: The amount of incoming QUIC streams created on the QUIC connection.
+: <dfn for="QuicTransportStats" dict-member>bytesReceived</dfn>
+:: The amount of total bytes received on the QUIC connection, including
+duplicate data for streams. Does not include UDP or any other outer framing.
+: <dfn for="QuicTransportStats" dict-member>packetsReceived</dfn>
+:: The amount of total packets received on the QUIC connection, including
+packets that were not processable.
+: <dfn for="QuicTransportStats" dict-member>minRttUs</dfn>
+:: The minimum RTT.
```
pthatcherg: Over what time window?  The entire connection?

index.bs
```diff
@@ -504,6 +505,67 @@ QuicTransportBase includes DatagramTransport;
...
+:: The amount of packets sent on the QUIC connection, including retransmissions.
+: <dfn for="QuicTransportStats" dict-member>numOutgoingStreamsCreated</dfn>
+:: The amount of outgoing QUIC streams created on the QUIC connection.
+: <dfn for="QuicTransportStats" dict-member>numIncomingStreamsCreated</dfn>
+:: The amount of incoming QUIC streams created on the QUIC connection.
+: <dfn for="QuicTransportStats" dict-member>bytesReceived</dfn>
+:: The amount of total bytes received on the QUIC connection, including
+duplicate data for streams. Does not include UDP or any other outer framing.
+: <dfn for="QuicTransportStats" dict-member>packetsReceived</dfn>
+:: The amount of total packets received on the QUIC connection, including
```
pthatcherg: All of these seem like "number of" would sound better than "amount of".

index.bs
shampson: The entire connection. It looks like QUICHE does this for the entire time the connection has been open:
https://cs.chromium.org/chromium/src/net/third_party/quiche/src/quic/core/congestion_control/rtt_stats.cc?dr=CSs&g=0&l=43

shampson: I updated to "The minimum RTT observed on the entire connection." Leaving this comment open and leaving it to you to resolve or suggest something different. I think this is the most straight forward thing to do, but maybe in the future  a given time window would be better.

shampson:
Thanks for also giving this a look @pthatcherg

PR #66, opened 2019-10-22
Simplify example

chicoxyzzy:


yoavweiss:
Marked as non substantive for IPR from ash-nazg.

PR #68, opened 2019-10-25
Fix one link to QuicTransportBase

ricea:
Also remove some blank lines.


yoavweiss:
@ricea - can you join the WICG and link your GH account with your W3C one to appease the IPR bots?

ricea:
@yoavweiss Sorry for the delay. I have joined WICG and linked the accounts. It should work now if you rerun the check.

PR #87, opened 2019-12-13
Use URL to identify QuicTransport

### LINKED ISSUE #26: Provide a way to specify a subprotocol in client-server
WebSockets have a subprotocol mechanism which works similar to ALPN (see https://tools.ietf.org/html/rfc6455#section-1.9).  Perhaps we should provide web applications using this API the same functionality directly via ALPN, e.g., if the client asks for "foo" and "bar", we provide ALPNs "wq:foo" and "wq:bar".

vasilvv:
The URL scheme for QuicTransport is now defined in
https://tools.ietf.org/html/draft-vvv-webtransport-quic-01#section-6

Fixes #26

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -61,6 +61,8 @@ Boilerplate: omit conformance
 <pre class="link-defaults">
 spec:streams; type:interface; text:ReadableStream
 spec:html; type:dfn; for:/; text:origin
+spec:url; type:dfn; text:scheme
+spec:url; type:dfn; text:fragment
 </pre>
 <pre class="anchors">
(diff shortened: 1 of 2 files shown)


vasilvv:
@ricea Could you take a look at this?  I assume this is what we currently do in our implementation.

ricea:
lgtm. It matches the implementation: https://cs.chromium.org/chromium/src/third_party/blink/renderer/modules/webtransport/quic_transport.cc?type=cs&sq=package:chromium&g=0&l=86

yutakahirano:
Our implementation removes username and password.

PR #89, opened 2019-12-14
Expose everything to Workers

pthatcherg:
Fixes https://github.com/WICG/web-transport/issues/37


vasilvv:
Adding a note here that we should check the Exposed attribute on our IDL, since right now some of the interfaces and mixins that have it should not have it.

PR #100, opened 2020-02-06
Allow QuicTransport to connect to the servers not using a publicly trusted certificate.

### LINKED ISSUE #18: Consider adding support for certificates not anchored to Web PKI
I've been talking to some folks who are currently using RTC Data Channels but are interested in using WebTransport.  One of the use cases that came up is a situation where a website spins up an ephemeral VM and wants the user to talk to that VM directly.  Provisioning a real browser-trusted certificate can be a lot of pain in this scenario.  If we don't provide a way to deal with this, we'd regress over what the Data Channels currently provide.

We can do this with QuicTransport, since the API we designed is sufficiently low-level, i.e. we don't risk accidentally leaking the channel with altered trust model into random parts of the web ecosystem; nor do we send any sensitive data like cookies as the part of the handshake.

I've thought about this a bunch, and my current conclusion is that the best way to do this would be to essentially copy the WebRTC mechanism where you just specify the expected certificate hash.  Here are some alternatives, and why I believe they are bad:

* **Allowing users to just ignore the certificate**.  This is bad in an obvious way: it encourages users to not do any authentication at all, steering them towards bad practices.  It's also bad in a non-obvious way: if the users do decide to do their own authentication, they would have to bind the underlying cryptographic channel to it, which would require us to support TLS secret exporters.
* **Allowing users to have a custom certificate callback**.  This was my original instinct, however, on the reflection, I am not quite happy with it.  My main concern is that QUIC connections can be intercepted by an enterprise MITM firewall, and the certificate in that firewall could contain potentially sensitive information.

vasilvv:
Fixes #18.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -521,21 +521,33 @@ defined in [[!QUIC-DATAGRAM]].
 
 <pre class="idl">
 dictionary QuicTransportConfiguration {
-  sequence&lt;RTCDtlsFingerprint&gt; server_certificates;
+  sequence&lt;RTCDtlsFingerprint&gt; server_certificate_fingerprints;
 };
 </pre>
 
 <dfn dictionary>QuicTransportConfiguration</dfn> is a dictionary of parameters
 that determine how QuicTransport connection is established and used.
 
-: <dfn for="QuicTransportConfiguration" dict-member>server_certificates</dfn>
+: <dfn for="QuicTransportConfiguration" dict-member>server_certificate_fingerprints</dfn>
 :: If non-empty, the user agent SHALL deem a server certificate trusted if and
    only if it matches one of the fingerprints defined in
-   {{QuicTransportConfiguration/server_certificates}}.  The user agent SHALL
-   ignore any fingerprint that uses an unknown {{RTCDtlsFingerprint/algorithm}}
-   or has a malformed {{RTCDtlsFingerprint/value}}.  If empty, the user agent
-   SHALL use certificate verification procedures it would use for normal
-   [=fetch=] operations.
+   {{QuicTransportConfiguration/server_certificate_fingerprints}} and satisfies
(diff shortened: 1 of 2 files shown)


vasilvv:
cc @aboba @ricea @yutakahirano

ricea:
lgtm

yutakahirano:
lgtm

aboba:
Recommend that server_certificates be renamed to fingerprints, since server_certificates is not actually a sequence of certificates, but a sequence of fingerprints.

vasilvv:
Renamed to `server_certificate_fingerprints`.  I also added a requirement for certificates to expire, after some discussion with folks who were worried this API would lead people to accidentally pin certificates without ability to rotate them afterwards.

vasilvv:
Per our previous discussion, I've described how exactly a certificate is matched against the fingerprint.

aboba:
[RFC 8122](https://tools.ietf.org/html/rfc8122) contains the latest text relating to DTLS fingerprint matching, in [Section 5.1](https://tools.ietf.org/html/rfc8122#section-5.1): 

" An endpoint MUST, as a minimum, calculate a fingerprint using both
   the 'SHA-256' hash function algorithm and the hash function used to
   generate the signature on the certificate for each possible
   certificate.  Including the hash from the signature algorithm ensures
   interoperability with strict implementations of RFC 4572.
   Either of these fingerprints MAY be omitted if the endpoint includes
   a hash with a stronger hash algorithm that it knows that the peer
   supports, if it is known that the peer does not support the hash
   algorithm, or if local policy mandates use of stronger algorithms...

   An endpoint MUST select the set of fingerprints that use its most
   preferred hash function (out of those offered by the peer) and verify
   that each certificate used matches one fingerprint out of that set.

   If a certificate does not match any such fingerprint, the endpoint
   MUST NOT establish the TLS connection."

PR #114, opened 2020-05-11
Add an example for receiving from ReceiveStreams.

jianjunz:
Section 16 only has examples for datagrams. Receiving chunks from `ReceiveStream`s is a little bit different from receiving datagrams. It requires developers to have a reader to get `ReceiveStream`s first, then another reader to read chunks from each `ReceiveStream`s.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -961,9 +961,8 @@ while (true) {
     const {value: chunk, done} = await chunkReader.read();
     if(done){
       break;
-    } else {
-      // Process the data.
     }
+    // Process the data.
   }
 }
 </pre>
--- a/index.bs
+++ b/index.bs
@@ -939,6 +939,35 @@ while (true) {
 }
 </pre>
 
+## Receiving from ReceiveStream ##  {#example-receiving-from-receivestream}
+
+*This section is non-normative.*
+
+Reading from ReceiveStreams can be achieved by calling
+{{UnidirectionalStreamsTransport/receiveStreams()}} method, then getting the
+reader for each {{ReceiveStream}}.
+
+<pre class="example" highlight="js">
+const transport = getTransport();
+const receiveStreamReader = transport.receiveStreams().getReader();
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -939,6 +939,35 @@ while (true) {
...
+    break;
+  }
+  // Process ReceiveStreams created by remote endpoint.
+  const chunkReader = receiveStream.readable.getReader();
+  while(true){
+    const {value: chunk, done} = await chunkReader.read();
+    if(done){
+      break;
+    } else {
+      // Process the data.
```
ricea, line 965: I suggest you don't put this inside an `else` section, to match the pattern above and avoid extra indentation.

You could also consider writing `processTheData(chunk)` to make it clearer where the data is.

jianjunz: Thanks for your suggestions. New commits were added.

index.bs
```diff
@@ -939,6 +939,34 @@ while (true) {
...
+const transport = getTransport();
+const receiveStreamReader = transport.receiveStreams().getReader();
+while (true) {
+  const {value: receiveStream, done} = await receiveStreamReader.read();
+  if (done) {
+    break;
+  }
+  // Process ReceiveStreams created by remote endpoint.
+  const chunkReader = receiveStream.readable.getReader();
+  while(true){
```
ricea, line 960: Please put a space between `while` and `(true)`. Also between `if` and `(done)` below.

jianjunz: Fixed. Thanks.

ricea:
Non-owner lgtm.

jianjunz:
Could someone please help me merge this PR if it looks good to you. Thanks.

vasilvv:
Since this is an example, I've marked this as non-substantive from IPR perspective.

The example LGTM.  Thanks for providing it!

jianjunz:
Thanks for your review.

PR #134, opened 2020-09-15
Merge QuicTransport and Http3Transport interfaces into a single WebTransport interface

### LINKED ISSUE #129: Merge QuicTransport and Http3Transport into a single class?
Currently we have two separate constructors for QuicTransport and Http3Transport:

```js
let transport1 = new QuicTransport("quic-transport://example.org/endpoint");
let transport2 = new Http3Transport("https://example.org/endpoint");
```

Notice that the transport information is redundantly encoded twice: once in the class name, and once in the scheme.

I propose that we merge those into a single constructor that dispatches based on the scheme:

```js
let transport1 = new WebTransport("quic-transport://example.org/endpoint");  // QuicTransport 
let transport2 = new WebTransport("https://example.org/endpoint");  // Http{2,3}Transport
```
This has a nice property of increasing transport agility (since URLs are typically data, transport can be changed with no code changes involved), and will probably simplify client implementation too (by reducing code duplication).

vasilvv:
Fixes #129.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -521,13 +521,13 @@ To <dfn>compute a certificate fingerprint</dfn>, do the following:
 
 <div algorithm="verify a certificate fingerprint">
 To <dfn>verify a certificate fingerprint</dfn>, do the following:
-1. Let |fprs| be the input array of fingerprints.
-1. Let |ref_fpr| be the [=compute a certificate fingerprint|computed
+1. Let |fingerprints| be the input array of fingerprints.
+1. Let |referenceFingerprint| be the [=compute a certificate fingerprint|computed
    fingerprint=] of the input certificate.
-1. For every fingerprint |fpr| in |fprs|:
-   1. If {{RTCDtlsFingerprint/algorithm}} of |fpr| is equal to "sha-256", and
-      {{RTCDtlsFingerprint/value}} of |fpr| is equal to |ref_fpr|, the
-      certificate is valid.  Return true.
+1. For every fingerprint |fingerprint| in |fingerprints|:
+   1. If {{RTCDtlsFingerprint/algorithm}} of |fingerprint| is equal to "sha-256",
+      and {{RTCDtlsFingerprint/value}} of |fingerprint| is equal to
+      |referenceFingerprint|, the certificate is valid.  Return true.
 1. Return false.
 
 </div>
--- a/index.bs
+++ b/index.bs
@@ -988,4 +988,4 @@ were provided by Robin Raymond.
 
 The {{WebTransport}} interface is based on the `QuicTransport` interface
 initially described in the [W3C ORTC CG](https://www.w3.org/community/ortc/),
-and have been adapted for use in this specification.
+and has been adapted for use in this specification.
--- a/index.bs
+++ b/index.bs
@@ -346,19 +346,98 @@ interface mixin DatagramTransport {
    but large enough to avoid dropping packets when for the stream is not read
    for short periods of time (due to the reader being paused).
 
-# `WebTransport` Mixin #  {#web-transport}
+# `WebTransport` Interface #  {#web-transport}
 
-The `WebTransport` includes the methods common to all transports, such as
-state, state changes, and the ability to close the transport.
+The `WebTransport` provides a unified interface to all client-server transports
+that are supported by the WebTransport API.  It implements all of the transport
+mixins ({{UnidirectionalStreamsTransport}}, {{BidirectionalStreamsTransport}},
+{{DatagramTransport}}), as well as stats and transport state information.
 
 <pre class="idl">
-interface mixin WebTransport {
+[Exposed=(Window,Worker)]
+interface WebTransport {
+  constructor(USVString url, optional WebTransportOptions options = {});
+
+  Promise&lt;WebTransportStats&gt; getStats();
   readonly attribute WebTransportState state;
   readonly attribute Promise&lt;WebTransportCloseInfo&gt; closed;
   void close(optional WebTransportCloseInfo closeInfo = {});
   attribute EventHandler onstatechange;
 };
+
+WebTransport includes UnidirectionalStreamsTransport;
+WebTransport includes BidirectionalStreamsTransport;
+WebTransport includes DatagramTransport;
 </pre>
+
+## Constructor ##  {#webtransport-constructor}
+
+When the {{WebTransport/constructor()}} constructor is invoked, the user
+agent MUST run the following steps:
+1. Let |parsedURL| be the [=URL record=] resulting from [=URL parser|parsing=]
+   of {{WebTransport/constructor(url, options)/url}}.
+1. If |parsedURL| is a failure, [=throw=] a {{SyntaxError}} exception.
+1. If |parsedURL| [=scheme=] is not `quic-transport` or `https`, [=throw=]
+   a {{SyntaxError}} exception.
+1. If |parsedURL| [=fragment=] is not null, [=throw=] a {{SyntaxError}} exception.
+1. Let |transport| be a newly constructed {{WebTransport}} object.
+1. Let |transport| have a
+   <dfn attribute for="WebTransport">\[[OutgoingStreams]]</dfn> internal slot
+   representing a sequence of {{OutgoingStream}} objects, initialized to empty.
+1. Let |transport| have a
+   <dfn attribute for="WebTransport">\[[ReceivedStreams]]</dfn> internal slot
+   representing a {{ReadableStream}} of {{IncomingStream}} objects, initialized
+   to empty.
+1. Let |transport| have a
+  <dfn attribute for="WebTransport">\[[ReceivedBidirectionalStreams]]</dfn>
+  internal slot representing a {{ReadableStream}} of {{IncomingStream}}
+  objects, initialized to empty.
+1. Let |transport| have a
+   <dfn attribute for="WebTransport">\[[WebTransportState]]</dfn> internal
+   slot, initialized to `"connecting"`.
+1. Let |transport| have a
+   <dfn attribute for="WebTransport">\[[SentDatagrams]]</dfn> internal slot
+   representing a {{WritableStream}} of {{Uint8Array}}s, initialized to empty.
+1. Let |transport| have a
+   <dfn attribute for="WebTransport">\[[ReceivedDatagrams]]</dfn> internal
+   slot representing a {{ReadableStream}} of {{Uint8Array}}s, initialized to
+   empty.
+1. If the scheme of |parsedURL| is `quic-transport`, [=in parallel=],
+   [=initialize WebTransport over QUIC=].
+1. If the scheme of |parsedURL| is `https`, [=in parallel=],
+   [=initialize WebTransport over HTTP=].
+1. Run these steps in parallel:
+1. Return |transport|.
+
+<div algorithm="initialize WebTransport over QUIC">
+To <dfn>initialize WebTransport over QUIC</dfn> for a given |transport| and
+|parsedURL|, do the following:
+
+1. Let |clientOrigin| be |transport|'s [=relevant settings object=]'s
+   [=origin=],
+   [serialized](https://html.spec.whatwg.org/multipage/origin.html#ascii-serialisation-of-an-origin).
+1. Establish a QUIC connection to the address identified by |parsedURL|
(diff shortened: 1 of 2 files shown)


vasilvv:
@aboba @yutakahirano @ricea: please take a look.

index.bs
```diff
@@ -346,20 +346,99 @@ interface mixin DatagramTransport {
...
 
-# `WebTransport` Mixin #  {#web-transport}
+# `WebTransport` Interface #  {#web-transport}
 
-The `WebTransport` includes the methods common to all transports, such as
-state, state changes, and the ability to close the transport.
+The `WebTransport` provides a unified interface to all client-server transports
+that are supported by the WebTransport API.  It implements all of the transport
+mixins ({{UnidirectionalStreamsTransport}}, {{BidirectionalStreamsTransport}},
+{{DatagramTransport}}), as well as stats and transport state information.
```
yutakahirano, line 354: Do we still need these mixins given that we only have WebTransport now?

index.bs
```diff
@@ -391,6 +470,76 @@ interface mixin WebTransport {
...
+     1. Let |p| be a new promise.
+     1. Return |p| and continue the following steps in background.
+         1. Gather the stats from the underlying QUIC connection.
+         1. Once stats have been gathered, resolve |p| with the
+            {{WebTransportStats}} object, representing the gathered stats.
+
+## Configuration ##  {#web-transport-configuration}
+
+<pre class="idl">
+dictionary WebTransportOptions : WebTransportOptions {
```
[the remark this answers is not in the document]
vasilvv: Whoops.  Fixed.

index.bs
vasilvv: We still need some way to leave a plug for a peer-to-peer version of WebTransport, but I suspect we don't need all three of them.  Let's fix this in a follow-up, though.

index.bs
```diff
@@ -346,20 +346,99 @@ interface mixin DatagramTransport {
...
+  constructor(USVString url, optional WebTransportOptions options = {});
+
+  Promise&lt;WebTransportStats&gt; getStats();
   readonly attribute WebTransportState state;
   readonly attribute Promise&lt;WebTransportCloseInfo&gt; closed;
   void close(optional WebTransportCloseInfo closeInfo = {});
   attribute EventHandler onstatechange;
 };
+
+WebTransport includes UnidirectionalStreamsTransport;
```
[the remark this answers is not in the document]
vasilvv: We still need some way to leave a plug for a peer-to-peer version of WebTransport, but I suspect we don't need all three of them.  Let's fix this in a follow-up, though.

index.bs
```diff
@@ -346,20 +346,99 @@ interface mixin DatagramTransport {
    but large enough to avoid dropping packets when for the stream is not read
    for short periods of time (due to the reader being paused).
 
-# `WebTransport` Mixin #  {#web-transport}
+# `WebTransport` Interface #  {#web-transport}
 
-The `WebTransport` includes the methods common to all transports, such as
-state, state changes, and the ability to close the transport.
+The `WebTransport` provides a unified interface to all client-server transports
```
ricea, line 351: `The WebTransport` reads strangely to me. Maybe write `The WebTransport constructor` or just `WebTransport`?

index.bs
```diff
@@ -346,20 +346,99 @@ interface mixin DatagramTransport {
...
+WebTransport includes BidirectionalStreamsTransport;
+WebTransport includes DatagramTransport;
 </pre>
 
+## Constructor ##  {#webtransport-constructor}
+
+When the {{WebTransport/constructor()}} constructor is invoked, the user
+agent MUST run the following steps:
+1. Let |parsedURL| be the [=URL record=] resulting from [=URL parser|parsing=]
+   of {{WebTransport/constructor(url, options)/url}}.
```
ricea, line 378: I think "of" is not needed here.

index.bs
```diff
@@ -346,20 +346,99 @@ interface mixin DatagramTransport {
...
+1. If |parsedURL| is a failure, [=throw=] a {{SyntaxError}} exception.
+1. If |parsedURL| [=scheme=] is not `quic-transport` or `https`, [=throw=]
+   a {{SyntaxError}} exception.
+1. If |parsedURL| [=fragment=] is not null, [=throw=] a {{SyntaxError}} exception.
+1. Let |transport| be a newly constructed {{WebTransport}} object.
+1. Let |transport| have a
```
ricea, line 384: s/a/an/

index.bs
```diff
@@ -346,20 +346,99 @@ interface mixin DatagramTransport {
...
+   <dfn attribute for="WebTransport">\[[OutgoingStreams]]</dfn> internal slot
+   representing a sequence of {{OutgoingStream}} objects, initialized to empty.
+1. Let |transport| have a
+   <dfn attribute for="WebTransport">\[[ReceivedStreams]]</dfn> internal slot
+   representing a {{ReadableStream}} of {{IncomingStream}} objects, initialized
+   to empty.
+1. Let |transport| have a
+  <dfn attribute for="WebTransport">\[[ReceivedBidirectionalStreams]]</dfn>
+  internal slot representing a {{ReadableStream}} of {{IncomingStream}}
```
ricea, line 393: Should this be `{{BidirectionStream}}` objects?

index.bs
```diff
@@ -346,20 +346,99 @@ interface mixin DatagramTransport {
...
+   representing a {{WritableStream}} of {{Uint8Array}}s, initialized to empty.
+1. Let |transport| have a
+   <dfn attribute for="WebTransport">\[[ReceivedDatagrams]]</dfn> internal
+   slot representing a {{ReadableStream}} of {{Uint8Array}}s, initialized to
+   empty.
+1. If the scheme of |parsedURL| is `quic-transport`, [=in parallel=],
+   [=initialize WebTransport over QUIC=].
+1. If the scheme of |parsedURL| is `https`, [=in parallel=],
+   [=initialize WebTransport over HTTP=].
+1. Run these steps in parallel:
```
ricea, line 409: I think this line should be removed.

index.bs
```diff
@@ -346,20 +346,99 @@ interface mixin DatagramTransport {
...
+1. Return |transport|.
+
+<div algorithm="initialize WebTransport over QUIC">
+To <dfn>initialize WebTransport over QUIC</dfn> for a given |transport| and
+|parsedURL|, do the following:
+
+1. Let |clientOrigin| be |transport|'s [=relevant settings object=]'s
+   [=origin=],
+   [serialized](https://html.spec.whatwg.org/multipage/origin.html#ascii-serialisation-of-an-origin).
```
ricea, line 418: Note for future updates: find a way to auto-linkify this.

index.bs
```diff
@@ -346,20 +346,99 @@ interface mixin DatagramTransport {
...
+   internal slot to `"failed"` and abort these steps.
+1. Set |transport|'s {{[[WebTransportState]]}} internal slot to `"connected"`.
+
+</div>
+
+<div algorithm="initialize WebTransport over HTTP">
+To <dfn>initialize WebTransport over HTTP</dfn> for a given |transport| and
+|parsedURL|, do the following:
+
+1. Establish an HTTP/3 WebTransport session to the address identified by
```
ricea, line 434: We should throw an exception if the `serverCertificateFingerprints` option was specified.

index.bs
```diff
@@ -391,6 +470,76 @@ interface mixin WebTransport {
...
 
+: <dfn for="WebTransport" method>getStats()</dfn>
+:: Gathers stats for this {{WebTransport}}'s QUIC
+   connection and reports the result asynchronously.</p>
+
+   When close is called, the user agent MUST run the following steps:
+     1. Let |transport| be the WebTransport on which `getStats` is invoked.
+     1. Let |p| be a new promise.
+     1. Return |p| and continue the following steps in background.
+         1. Gather the stats from the underlying QUIC connection.
```
ricea, line 481: Is this safe from a security and privacy standpoint with an HTTP connection?

index.bs
```diff
@@ -391,6 +470,76 @@ interface mixin WebTransport {
...
+<pre class="idl">
+dictionary WebTransportOptions {
+  sequence&lt;RTCDtlsFingerprint&gt; serverCertificateFingerprints;
+};
+</pre>
+
+<dfn dictionary>WebTransportOptions</dfn> is a dictionary of parameters
+that determine how WebTransport connection is established and used.
+
+: <dfn for="WebTransportOptions" dict-member>serverCertificateFingerprints</dfn>
```
ricea, line 496: I think it would be less ambiguous to use a non-normative description here and put the normative behaviour in the algorithm steps for the constructor.

index.bs
```diff
@@ -391,6 +470,76 @@ interface mixin WebTransport {
...
+   Certificate message defined in [[!RFC5280]].
+1. Compute the SHA-256 hash of |cert|.  Format it as `fingerprint` BNF
+   rule described in Section 5 of [[!RFC8122]].
+
+</div>
+
+<div algorithm="verify a certificate fingerprint">
+To <dfn>verify a certificate fingerprint</dfn>, do the following:
+1. Let |fprs| be the input array of fingerprints.
+1. Let |ref_fpr| be the [=compute a certificate fingerprint|computed
```
ricea, line 522: `fprs` and `ref_fpr` seem to be unnecessarily abbreviated. How about `fingerprints` and `referenceFingerprint`?

index.bs
```diff
@@ -484,169 +633,30 @@ dictionary QuicTransportStats {
 
 The dictionary SHALL have the following attributes:
 
-: <dfn for="QuicTransportStats" dict-member>timestamp</dfn>
+: <dfn for="WebTransportStats" dict-member>timestamp</dfn>
 :: The `timestamp` for when the stats are gathered, relative to the
```
ricea, line 637: Should use DOMTimeStamp: https://heycam.github.io/webidl/#DOMTimeStamp

index.bs
```diff
@@ -1031,6 +979,6 @@ Alvestrand, Stefan H&aring;kansson, Bernard Aboba and Dominique
 Haza&euml;l-Massieux, for their support. Contributions to this specification
 were provided by Robin Raymond.
 
-The {{QuicTransport}} and `QuicStream` objects were initially described in the
+The `QuicTransport` and `QuicStream` objects were initially described in the
```
ricea, line 982: How about something like

> The {{WebTransport}} interface is based on the `QuicTransport` interface initially described ...

Since we're not using `QuicStream` it's probably unnecessary to mention it.

yutakahirano (approved):

index.bs
vasilvv: Is there any specific reason for it to use low-res version? https://w3c.github.io/webrtc-pc/#dom-rtcstats uses `DOMHighResTimeStamp` too.

index.bs
vasilvv: Done.

index.bs
vasilvv: Done.

index.bs
vasilvv: Some of them are safe, others are not.  I made it an error for now, and added a note saying we should revisit this and split them into "available only in dedicated" and "always available".

index.bs
vasilvv: Done.

index.bs
vasilvv: Agreed, though I'd have to write more text about how connection establishment happens in general, so perhaps that would be better left for a follow-up CL.

index.bs
ricea: Sorry, yes, `DOMHighResTimeStamp` is the right choice.

index.bs
ricea: Thanks.

index.bs
```diff
@@ -979,6 +986,6 @@ Alvestrand, Stefan H&aring;kansson, Bernard Aboba and Dominique
 Haza&euml;l-Massieux, for their support. Contributions to this specification
 were provided by Robin Raymond.
 
-The `QuicTransport` and `QuicStream` objects were initially described in the
-[W3C ORTC CG](https://www.w3.org/community/ortc/), and have been adapted for
-use in this specification.
+The {{WebTransport}} interface is based on the `QuicTransport` interface
+initially described in the [W3C ORTC CG](https://www.w3.org/community/ortc/),
+and have been adapted for use in this specification.
```
ricea, line 991: s/have/has/.

ricea (approved):
lgtm

PR #149, opened 2020-10-21
Adding use-cases document

wilaw:
Per an action item at the W3C WebTransport TPAC [meeting #2](https://www.w3.org/wiki/WebTransport/TPAC_2020#WebTransport_WG_Virtual_Meeting_2:_Next_Steps), committing the collected use-cases as a new document.

### DIFF
--- a/use-cases.md
+++ b/use-cases.md
@@ -1,3 +1,5 @@
+## Use-Cases for WebTransport
+
 Ordinality of listing does not imply priority. 
 
 1. **Machine learning**
@@ -23,7 +25,8 @@ Ordinality of listing does not imply priority.
     1. Online document sharing
     1. Remote assistance temporarily "taking over" control of a system
 1. **Time Synchronized Multimedia Web communications**
-    1. Combining geo-separate singing and/or instruments together online with precise time synchronization
+    1. Combining geo-separate singing and/or instruments together online with precise time synchronization (deprecated as valid-case during Oct 19 TPAC meeting)
+    1. Add HUD overlay on top of received video from sensors or game data.
 1. **IOT sensor and analytics data transfer**
     1. Efficient and intermittent transmission of data. For example  - sending a 1 bit flag, GPS position updates, mouse clicks on site etc. 
     1. Sensor data upload  - including filters, aggregation, triggers.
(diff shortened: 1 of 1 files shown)


use-cases.md
```diff
@@ -0,0 +1,50 @@
...
+    1. Simpler session establishment
+    1. Censorship circumvention - preventing fingerprinting and identification during session establishment.
+1. **Remote desktop**
+    1. Transmission of screen capture/sharing and control instructions.
+    1. Collaborative work on a shared screen.
+    1. Including scaling to very large audiences.
+    1. Online document sharing
+    1. Remote assistance temporarily "taking over" control of a system
+1. **Time Synchronized Multimedia Web communications**
+    1. Combining geo-separate singing and/or instruments together online with precise time synchronization (deprecated as valid-case during Oct 19 TPAC meeting)
```
jan-ivar, line 28:
```suggestion
    1. Add heads up display information on top of received video from sensors or game data.
```

use-cases.md
```diff
@@ -0,0 +1,50 @@
+## Use-Cases for WebTransport
+
+Ordinality of listing does not imply priority. 
+
+1. **Machine learning**
+    1. Speech translation/emotion analysis  - sending audio/video data from client to server for analysis and receiving translated data/text/audio in return.
+    1. Security camera analysis - data and/or video sent to cloud service for analysis. Service may return data instructions.
```
jan-ivar, line 7: This use case may need clarification. Is this a p2p situation, a non-web-based "security camera" talking to a server, or a local camera connected to a browser client? Unless it's the latter it may be out of scope.

martinthomson:
I have a few comments:

1. There are lots of diverse use cases here and it is not clear how each of them connects to the requirements.

2. You need to capture the bidirectionality requirement implied by some of these.

3. There might need to be some scoping tweaks required.  Some things might need to be ruled out of scope.    As a concrete example, the long-poll replacement might include some protocol changes if the goal is to accommodate intermediation.  That might also manifest in API-visible artifacts in the form of different identifiers and so forth.  Keeping that out of scope - at least until the group understands the needs better - is probably wise.

use-cases.md
jan-ivar: ok if web-based device.

PR #153, opened 2020-10-29
Add missing WebIDL for ready

aboba:
PR https://github.com/w3c/webtransport/pull/150 did not include the WebIDL for ready.  The problem was not caught because the PR evaluation process doesn't include a bikeshed compile automatic action (Issue https://github.com/w3c/webtransport/issues/108 )

Partial fix for Issue https://github.com/w3c/webtransport/issues/137

PR #150, opened 2020-10-27
Introduce WebTransport.ready

yutakahirano:
Fixes #92.

PR #160, opened 2020-10-30
Put WebTransport.ready to the IDL definition

yutakahirano:
I forgot to put the attribute at #150.


yutakahirano:
I noticed there already is a fix: #153

PR #161, opened 2020-10-30
Add a complete example

aboba:
This PR adds a complete example.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1001,6 +1001,9 @@ the server, and sending and receiving datagrams.
 <pre class="example" highlight="js">
 // Adds an entry to the event log on the page, optionally applying a specified
 // CSS class.
+
+var currentTransport, streamNumber, currentTransportDatagramWriter;
+
 function addToEventLog(text, severity = 'info') {
   let log = document.getElementById('event-log');
   let mostRecentEntry = log.lastElementChild;
@@ -1020,7 +1023,7 @@ function addToEventLog(text, severity = 'info') {
 
 // "Connect" button handler.
 async function connect() {
-  let url = document.getElementById('url').value;
+  const url = document.getElementById('url').value;
   try {
     var transport = new WebTransport(url);
     addToEventLog('Initiating connection...');
@@ -1045,15 +1048,9 @@ async function connect() {
         addToEventLog('Connection closed abruptly.', 'error');
       });
 
-  globalThis.currentTransport = transport;
-  globalThis.streamNumber = 1;
-  try {
-    globalThis.currentTransportDatagramWriter = transport.datagramWritable.getWriter();
-    addToEventLog('Datagram writer ready.');
-  } catch (e) {
-    addToEventLog('Datagram writer failed. ' + e, 'error');
-    return;
-  }
+  currentTransport = transport;
+  currentTransportDatagramWriter = transport.datagramWritable.getWriter();
+  streamNumber = 1;
   readDatagrams(transport);
   acceptUnidirectionalStreams(transport);
   document.forms.sending.elements.send.disabled = false;
@@ -1066,11 +1063,11 @@ async function sendData() {
   let encoder = new TextEncoder('utf-8');
   let rawData = sending.data.value;
   let data = encoder.encode(rawData);
-  let transport = globalThis.currentTransport;
+  let transport = currentTransport;
   try {
     switch (form.sendtype.value) {
       case 'datagram':
-        await globalThis.currentTransportDatagramWriter.write(data);
+        await currentTransportDatagramWriter.write(data);
         addToEventLog('Sent datagram: ' + rawData);
         break;
       case 'unidi': {
@@ -1083,7 +1080,7 @@ async function sendData() {
       }
       case 'bidi': {
         let stream = await transport.createBidirectionalStream();
-        let number = globalThis.streamNumber++;
+        let number = streamNumber++;
         readFromIncomingStream(stream, number);
 
         let writer = stream.writable.getWriter();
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -990,6 +990,174 @@ while (true) {
...
+        addToEventLog('Connection closed normally.');
+      })
+      .catch(() => {
+        addToEventLog('Connection closed abruptly.', 'error');
+      });
+
+  globalThis.currentTransport = transport;
+  globalThis.streamNumber = 1;
+  try {
+    globalThis.currentTransportDatagramWriter = transport.datagramWritable.getWriter();
```
ricea, line 1051: I don't think this can fail, so the `try ... catch` block is redundant.

index.bs
```diff
@@ -990,6 +990,174 @@ while (true) {
...
+*This section is non-normative.*
+
+This example illustrates use of the closed and ready promises, opening
+of uni-directional and bi-directional streams by either the client or
+the server, and sending and receiving datagrams.
+
+<pre class="example" highlight="js">
+// Adds an entry to the event log on the page, optionally applying a specified
+// CSS class.
+function addToEventLog(text, severity = 'info') {
```
ricea, line 1004: Since this utility function isn't related to WebTransport, it might be an unnecessary distraction. Perhaps leave it as an exercise for the reader instead?

index.bs
```diff
@@ -990,6 +990,174 @@ while (true) {
...
+  if (mostRecentEntry != null &&
+      mostRecentEntry.getBoundingClientRect().top <
+          log.getBoundingClientRect().bottom) {
+    entry.scrollIntoView();
+  }
+}
+
+// "Connect" button handler.
+async function connect() {
+  let url = document.getElementById('url').value;
```
ricea, line 1023: Since `url` doesn't change, it can be const.

index.bs
```diff
@@ -990,6 +990,174 @@ while (true) {
...
+
+  transport.closed
+      .then(() => {
+        addToEventLog('Connection closed normally.');
+      })
+      .catch(() => {
+        addToEventLog('Connection closed abruptly.', 'error');
+      });
+
+  globalThis.currentTransport = transport;
```
ricea, line 1048: I suggest declaring `currentTransport` with a top-level `let` rather than using `globalThis`. It doesn't make much practical difference, but it's good to promote best practices.

index.bs
```diff
@@ -990,6 +990,174 @@ while (true) {
...
+        break;
+      }
+      case 'bidi': {
+        let stream = await transport.createBidirectionalStream();
+        let number = globalThis.streamNumber++;
+        readFromIncomingStream(stream, number);
+
+        let writer = stream.writable.getWriter();
+        await writer.write(data);
+        writer.close();
```
ricea, line 1091: I think since bidirectional streams don't support a half-closed state, this will prevent reading data sent by the server. Probably better to just not close here, and rely on the server to close the stream.

index.bs
```diff
@@ -990,6 +990,174 @@ while (true) {
...
+            'Opened bidirectional stream #' + number +
+            ' with data: ' + rawData);
+        break;
+      }
+    }
+  } catch (e) {
+    addToEventLog('Error while sending data: ' + e, 'error');
+  }
+}
+// Reads datagrams from |transport| into the event log until EOF is reached.
```
ricea, line 1102: Blank line here please.

index.bs
```diff
@@ -990,6 +990,174 @@ while (true) {
...
+async function readDatagrams(transport) {
+  let reader = transport.datagramReadable.getReader();
+  let decoder = new TextDecoder('utf-8');
+  try {
+    while (true) {
+      let result = await reader.read();
```
ricea, line 1108: Optional: I prefer to use destructuring assignment like:
```javascript
const { value, done } = await reader.read();
```
but it might be harder to understand?

ricea (commented):
Sorry for the late review.

index.bs
aboba: Heh. I put that in there when looking for the correct attribute.

index.bs
aboba: Fixed.

index.bs
aboba: Fixed.

index.bs
aboba: Fixed.

index.bs
aboba: Opened Issue https://github.com/w3c/webtransport/issues/163 to clarify.

index.bs
aboba: Opened Issue https://github.com/w3c/webtransport/issues/164 to clarify.

index.bs
aboba: Looks like bi-directional streams are intended to support a half-closed state (though a bug in the current implementation doesn't allow this). The example runs anyway, because the server opens a separate bi-directional stream on which it returns the count.

index.bs
aboba: Fixed. Other examples also use destructuring assignment.

index.bs
aboba: Moved it to the end, so it will be less distracting.

index.bs
ricea: Ah, I see. Making datagramWritable undefined would be one way to express that there is no datagram support, which would indeed cause this code to throw.

index.bs
```diff
@@ -990,6 +990,184 @@ while (true) {
...
+      case 'bidi': {
+        let stream = await transport.createBidirectionalStream();
+        let number = streamNumber++;
+        readFromIncomingStream(stream, number);
+
+        let writer = stream.writable.getWriter();
+        await writer.write(data);
+        writer.close();
+        addToEventLog(
+            'Opened bidirectional stream #' + number +
```
martinthomson, line 1079: This is likely problematic: the stream number here is not the stream ID from QUIC (or h2).  That has the potential to create some really nasty confusion.

PR #166, opened 2020-11-11
Update CONTRIBUTING.md to w3c.

jan-ivar:
Please check:

1. @ylafon that I got the right patent policy links
2. @vasilvv I left the instructions for adding additional contributors the same though they differ from [other w3c specs](https://github.com/w3c/webrtc-pc/blob/master/CONTRIBUTING.md). Did they rely on WICG specific tooling? Do we want to keep them or converge?
3. Do we want to mention/give credit to WICG?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,8 +1,8 @@
-# Web Platform Incubator Community Group
-
-This repository is being used for work in the W3C Web Platform Incubator Community Group, governed by the [W3C Community License
-Agreement (CLA)](http://www.w3.org/community/about/agreements/cla/). To make substantive contributions,
-you must join the CG.
+Contributions to this repository are intended to become part of Recommendation-track documents governed by the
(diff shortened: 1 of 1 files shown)


jan-ivar:
@ylafon is there a way to automate bikeshed when PRs are pushed?

ylafon (approved):

ylafon:
There is even a  way to automatically publish on https://www.w3.org/TR/ (we need to publish a FPWD first, though). Or do you mean only generate the html for the edcopy?

jan-ivar:
@ylafon I just meant updating the html copy. Is there something like githhub Actions for that?

ylafon:
@jan-ivar yes, I can take care of that, do you want the generated html to be on another branch or the same one?

jan-ivar:
@ylafon Great thanks! I don't know that I have a strong preference. Where does the cow path lead?

PR #170, opened 2020-11-25
automatic spec build

ylafon:
his should build the spec for every push to the main branch, result will be published in the gh-pages branch.


jan-ivar (approved):
Only one way to find out if it works!

Issue #152, opened 2020-10-29
problem in testing maximum datagram sending speed


xujiasci:
I want to test the maximum datagram sending speed when using quic in web application, so I made an quic echo server and using the following script as client . here is the sending part.
```    
try {
        DatagramWriter = transport.sendDatagrams().getWriter();
        //await DatagramWriter.ready;
        startMill = (new Date()).getTime();
        var ii = 0;
        while (state != "stop" && ii++ < total) {
            let numdrawData = send_seq + rawData;
            numdrawData = numdrawData.substring(0, size);
            let data = encoder.encode(numdrawData);
            record[send_seq] = [(new Date()).getTime(), 0];
            await DatagramWriter.ready;
            DatagramWriter.write(data);
            send_seq++;
        }
    }
```
But it seems that a lot of packets event hasn`t being sent out. I wonder if it is casued by the sending buffer limit, Quic will drop some application data as the they are lost in the transmitting. How should I correct this ?

xujiasci:
I suppose that when DatagramWriter.ready is resolved, the transport should actually have the ability to send my data using datagram frame, not drop them silently. And I will receive echo datagram from server, only not if the packet contain this frame has beening droped by the net.

xujiasci:
When DatagramWriter.ready is rejected, I will get the reason it is caused by connection things or too much data.

ricea:
> I suppose that when DatagramWriter.ready is resolved, the transport should actually have the ability to send my data using datagram frame, not drop them silently. And I will receive echo datagram from server, only not if the packet contain this frame has beening droped by the net.

I think your expectation is reasonable, but that's not how it currently works. Depending on the resolution of #93, it may work that way in future.

For now, you can wait for `transport.ready` before trying to send any datagrams.

The Chrome implementation currently drops more datagrams than is necessary. I expect to change it to have a certain amount of backpressure on the datagrams in future.

jan-ivar:
This sounds like a Chrome issue. Sending datagrams is now [well-defined in the spec](https://w3c.github.io/webtransport/#writedatagrams) and does offer (sender-side) back pressure to producers of data to ensure a balance between throughput and timeliness with https://github.com/w3c/webtransport/pull/267.

xujiasci:
@yutakahirano Hello, I'm a little confused about all your discussing, can you give me an example about the right way to use webtransport datagram, if I don't want any packet droped by browser. Looking forward to your reply.

xujiasci:
[...]

PR #179, opened 2021-01-05
Control of datagramWritable Queue

aboba:
Fixes https://github.com/w3c/webtransport/issues/152

***
<a href="https://pr-preview.s3.amazonaws.com/aboba/webtransport/pull/179.html" title="Last updated on Feb 3, 2021, 1:39 AM UTC (fea30fb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/179/b8d1a94...aboba:fea30fb.html" title="Last updated on Feb 3, 2021, 1:39 AM UTC (fea30fb)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -468,7 +468,7 @@ To <dfn>initialize WebTransport over HTTP</dfn> for a given |transport| and
 
 ## Methods ##  {#webtransport-methods}
 
-: <dfn for "WebTransport" method>setDatagramWritableQueueExpirationDuration(long duration)</dfn>
+: <dfn for "WebTransport" method>setDatagramWritableQueueExpirationDuration()</dfn>
 :: Sets the expiration of the DatagramWritable queue, in milliseconds.
 
 : <dfn for="WebTransport" method>close()</dfn>
--- a/index.bs
+++ b/index.bs
@@ -469,7 +469,7 @@ To <dfn>initialize WebTransport over HTTP</dfn> for a given |transport| and
 ## Methods ##  {#webtransport-methods}
 
 : <dfn for="WebTransport" method>setDatagramWritableQueueExpirationDuration()</dfn>
-:: Sets the expiration of the {{DatagramWritable}} queue, in milliseconds.
+:: Sets the expiration of the {{datagramWritable}} queue, in milliseconds.
 
 : <dfn for="WebTransport" method>close()</dfn>
 :: Closes the WebTransport object. For QUIC, this triggers an <dfn>Immediate
@@ -527,7 +527,7 @@ that determine how WebTransport connection is established and used.
    [=fetch=] operations.
    
 : <dfn for="WebTransportOptions" dict-member>datagramWritableHighWaterMark</dfn>
-:: This controls the high water mark for the datagramWritable stream, as defined
+:: This controls the high water mark for the {{datagramWritable}} stream, as defined
    in [[WHATWG-STREAMS]] Section 2.5.
 
 <div algorithm="compute a certificate fingerprint">
--- a/index.bs
+++ b/index.bs
@@ -468,8 +468,8 @@ To <dfn>initialize WebTransport over HTTP</dfn> for a given |transport| and
 
 ## Methods ##  {#webtransport-methods}
 
-: <dfn for "WebTransport" method>setDatagramWritableQueueExpirationDuration()</dfn>
-:: Sets the expiration of the DatagramWritable queue, in milliseconds.
+: <dfn for="WebTransport" method>setDatagramWritableQueueExpirationDuration()</dfn>
+:: Sets the expiration of the {{DatagramWritable}} queue, in milliseconds.
 
 : <dfn for="WebTransport" method>close()</dfn>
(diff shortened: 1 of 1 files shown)


yutakahirano:
@ricea 

Thank you for working on this!

From our implementation experience we cannot use [the high water mark concept in the streams spec](https://streams.spec.whatwg.org/#high-water-mark) because if we use that we would have a queue in the WritableStream and  the underlying sink wouldn't know when the chunk is enqueued, and wouldn't be able to calculate the expiration correctly.

aboba:
@yutakahirano The reference to the high water mark concept has been removed.  Is it accurate now?

jan-ivar:
> From our implementation experience we cannot use [the high water mark concept in the streams spec](https://streams.spec.whatwg.org/#high-water-mark) because if we use that we would have a queue in the WritableStream and the underlying sink wouldn't know when the chunk is enqueued, and wouldn't be able to calculate the expiration correctly.

Can we put queues in both the WritableStream and the underlying source?

An underlying source can then fill its underlying queue by [resolving sinkWritePromises](https://streams.spec.whatwg.org/#writable-stream-default-controller-process-write) prematurely: _"Upon fulfillment of sinkWritePromise, ... Perform ! [DequeueValue](https://streams.spec.whatwg.org/#dequeue-value)(controller)."_ I.e. resolve promises immediately until the underlying buffer fills up, then start returning pending promises, to begin filling up the WritableStream's upper buffer instead, which causes upstream backpressure thanks to `highWaterMark`. Then over time resolve those pending promises one by one as the underlying buffer is emptied to keep it filled.

The size of the underlying queue would be the minimal needed for throughput to not suffer. The size of upper WritableStream queue I don't think matters much as long as it's > 0.

index.bs
```diff
@@ -467,6 +468,11 @@ To <dfn>initialize WebTransport over HTTP</dfn> for a given |transport| and
 
 ## Methods ##  {#webtransport-methods}
 
+: <dfn for="WebTransport" method>setDatagramWritableQueueExpirationDuration()</dfn>
+:: Sets the expiration of the {{datagramWritable}} queue, in milliseconds.
+   {{writer}}.{{ready}} is not fulfilled when there are more outgoing datagrams
+   neither sent nor discarded due to a timeout.
```
jan-ivar, line 474: The `ready` API is defined in algorithms in the streams spec, so we need to influence inputs to those algorithms. https://github.com/w3c/webtransport/issues/185

jan-ivar (changes requested):

jan-ivar:
With back pressure, I don't think we'll need `setDatagramWritableQueueExpirationDuration`, since from https://github.com/w3c/webtransport/issues/187#issuecomment-773010072 it's the receiver we'll need to drop packets on, if anything. So I would vote against merging this.

yutakahirano:
@xujiasci, you opened #152. Do you have thoughts on https://github.com/w3c/webtransport/pull/179#issuecomment-773284028?

aboba:
Let's not confuse the behavior of the  QUIC datagram sender implementation with what is going on in Javascript. As I understand it, `setDatagramWritableQueueExpirationDuration()` is about the behavior of the QUIC datagram sender.  The goal was to implement the behavior described in the [QUIC datagram specification Section 5.3](https://tools.ietf.org/html/draft-ietf-quic-datagram-01#section-5.3): 

   `DATAGRAM frames employ the QUIC connection's congestion controller.
   As a result, a connection may be unable to send a DATAGRAM frame
   generated by the application until the congestion controller allows
   it [I-D.ietf-quic-recovery].  The sender implementation MUST
   either delay sending the frame until the controller allows it or drop
   the frame without sending it (at which point it MAY notify the
   application).

   Implementations can optionally support allowing the application to
   specify a sending expiration time, beyond which a congestion-
   controlled DATAGRAM frame ought to be dropped without transmission.`

jan-ivar:
> The sender implementation MUST either delay sending the frame until the controller allows it

I think we should do that. I think this is a problem only with poorly implemented producers that don't listen to back pressure.

I think it's on the receiving end that we need to drop packets, not here.

aboba:
Delaying frames until the controller allows it could cause the frames to lose timeliness, so that they are no longer worth sending. For example, consider an audio or video payload that becomes useless if delayed too long.

index.bs
```diff
@@ -363,6 +363,7 @@ interface WebTransport {
   readonly attribute WebTransportState state;
   readonly attribute Promise&lt;undefined&gt; ready;
   readonly attribute Promise&lt;WebTransportCloseInfo&gt; closed;
+  undefined setDatagramWritableQueueExpirationDuration(long duration);
```
jan-ivar, line 366: @yutakahirano suggested moving this to `datagrams`...

Maybe `datagrams.setSendQueueExpiration(long duration)`?

jan-ivar:
Meeting:
- Even with backpressure, stale data may end up in the send queue, due to congestion control/ACKs
- Flushing the send queue of stale data is still useful to say "don't send data more stale than x seconds"

index.bs
aboba: That makes sense. Should we close this PR and add the method somewhere else?

PR #182, opened 2021-01-22
Improve examples.

jan-ivar:
***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webtransport/pull/182.html" title="Last updated on Jan 23, 2021, 1:28 AM UTC (b76e31a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/182/b8d1a94...jan-ivar:b76e31a.html" title="Last updated on Jan 23, 2021, 1:28 AM UTC (b76e31a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -999,7 +999,7 @@ and then consuming each {{ReceiveStream}} by iterating over its chunks.
 async function receiveData(url, processTheData) {
   const transport = new WebTransport(url);
   for await (const stream of transport.incomingUnidirectionalStreams) {
-    // consume streams individually, reporting errors in case they're per-stream
+    // consume streams individually, reporting per-stream errors
     ((async () => {
       try {
         for await (const chunk of stream.readable.getReader()) {
@@ -1020,7 +1020,7 @@ Decoding can also be done through pipes to new WritableStreams, for example usin
 async function receiveText(url, createWritableStreamForTextData) {
   const transport = new WebTransport(url);
   for await (const stream of transport.incomingUnidirectionalStreams) {
-    // consume streams individually, reporting errors in case they're per-stream
+    // consume streams individually, reporting per-stream errors
     (async () => {
       try {
         await stream.readable
--- a/index.bs
+++ b/index.bs
@@ -1151,15 +1151,15 @@ async function readFromIncomingStream(stream, number) {
 
 function addToEventLog(text, severity = 'info') {
   const log = document.getElementById('event-log');
-  const last = log.lastElementChild;
+  const previous = log.lastElementChild;
   const entry = document.createElement('li');
   entry.innerText = text;
   entry.className = &#96;log-${severity}&#96;;
   log.appendChild(entry);
 
-  // If the last entry in the log is visible, scroll the log to the new element.
-  if (last &&
-      last.getBoundingClientRect().top < log.getBoundingClientRect().bottom) {
+  // If the previous entry in the log was visible, scroll to the new element.
+  if (previous &&
+      previous.getBoundingClientRect().top < log.getBoundingClientRect().bottom) {
     entry.scrollIntoView();
   }
 }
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -945,43 +949,80 @@ Receiving datagrams can be achieved by calling
...
 <pre class="example" highlight="js">
-const transport = getTransport();
-const reader = transport.datagramReadable.getReader();
-while (true) {
-  const {value: datagram, done} = await reader.read();
-  if (done) {
-    break;
+async function(url) {
+  const transport = new WebTransport(url);
+  for await (datagram of transport.datagramReadable) {
```
ricea, line 954: I suggest writing `const datagram` to avoid using a global variable. Also below.

index.bs
```diff
@@ -909,13 +909,14 @@ Sending a buffer of datagrams can be achieved by using the
...
 
 <pre class="example" highlight="js">
-const transport = getTransport();
-const writer = transport.datagramWritable.getWriter();
-const datagrams = getDatagramsToSend();
-datagrams.forEach((datagram) => {
-  await writer.ready;
-  writer.write(datagram);
-});
+async function(url, datagrams) {
```
yutakahirano, line 912: It's better to name this function, or run this function. Ditto below.

index.bs
jan-ivar: Good catch, I keep doing that.

jan-ivar:
Hmm, I added some error propagation for the in parallel steps, but I'm not happy with it, since it defers errors until the end. Let me try again.

jan-ivar:
Are errors likely to occur per stream, or are they all fatal to the transport?

jan-ivar:
Ended up reporting per stream, since the iteration on `incomingUnidirectionalStreams` itself should fail if the connection fails.

jan-ivar:
For those interested, I wrote a quick [polyfill](https://jsfiddle.net/jib1/52byh1kq/) to verify this syntax:
```js
for await (const stream of transport.incomingUnidirectionalStreams) {
  await stream.readable.pipeThrough(new TextDecoderStream("utf-8")).pipeTo(new WritableConsole());
}
```
I dunno if it's production ready or anything, but it appears to work right in Canary.

jan-ivar:
> ```js
> for await (const stream of transport.incomingUnidirectionalStreams) {
>  await stream.readable.pipeThrough(new TextDecoderStream("utf-8")).pipeTo(new WritableConsole());
>}
> ```

For those curious, the reason for not reading in parallel here is to avoid interleaving the text output, since it's going to a combined log.

Now that I think about it, that might be good to show in the text decoder example.

index.bs
```diff
@@ -998,167 +1039,123 @@ the server, and sending and receiving datagrams.
...
-  entry.className = 'log-' + severity;
+  entry.className = &#96;log-${severity}&#96;;
   log.appendChild(entry);
 
-  // If the most recent entry in the log was visible, scroll the log to the
-  // newly added element.
-  if (mostRecentEntry != null &&
-      mostRecentEntry.getBoundingClientRect().top <
-          log.getBoundingClientRect().bottom) {
+  // If the last entry in the log is visible, scroll the log to the new element.
```
wilaw, line 1160: If the most recent entry in the log is not visible, ...

jan-ivar: I renamed it to `last` for brevity in line 1150 above, otherwise the logic is unchanged, except I see present tense snuck in there, which doesn't help. Let me find a better name...

index.bs
```diff
@@ -998,167 +1045,121 @@ the server, and sending and receiving datagrams.
...
-            'Opened bidirectional stream #' + number +
-            ' with data: ' + rawData);
+        await writer.close();
+        addToEventLog(&#96;Sent bidirectional stream #${n} with data: ${rawData}&#96;);
         break;
       }
     }
   } catch (e) {
-    addToEventLog('Error while sending data: ' + e, 'error');
+    addToEventLog(&#96;Error while sending data: ${e}&#96;, 'error');
```
martinthomson, line 1107: We can't use `<![CDATA[` blocks so that this quoting is not necessary?

yutakahirano (approved):

Issue #181, opened 2021-01-20
Make transport.datagram a duplex stream?


jan-ivar:
The streams spec defines a [duplex stream](https://streams.spec.whatwg.org/#other-specs-duplex) as a [`{readable, writable}`](https://streams.spec.whatwg.org/#example-both) pair. In light of that, would the following change be more ergonomic?
```js
-   const writer = transport.datagramWritable.getWriter();
+   const writer = transport.datagrams.writable.getWriter();

-   transport.datagramReadable.pipeTo(sink);
+   transport.datagrams.readable.pipeTo(sink);
```

This would allow [clever things](https://jsfiddle.net/jib1/uxeyrvk8/) like piping through a server processing step ([polyfill](https://jan-ivar.github.io/polyfills/datagrams.js)):
```js
new InputBox()
  .pipeThrough(new TextEncoderStream("utf-8"))
  .pipeThrough(transport.datagrams)
  .pipeThrough(new TextDecoderStream("utf-8"))
  .pipeTo(new WritableConsole());
```

yutakahirano:
Sounds reasonable.

jan-ivar:
I guess it should also be `datagrams` plural.

aboba:
Using duplex streams has implications. From the [WHATWG Streams specification Section 9.4.1](https://streams.spec.whatwg.org/#other-specs-duplex): 

"The trickiest thing to consider when specifying duplex streams is how to handle operations like canceling the readable side, or closing or aborting the writable side. It might make sense to leave duplex streams "half open", with such operations one one side not impacting the other side. Or it might be best to carry over their effects to the other side, e.g. by specifying that your readable side’s cancelAlgorithm will close the writable side."

jan-ivar:
Once we figure out https://github.com/w3c/webtransport/issues/197, I'd think _"half open"_ would make sense, since we can't assume all datagrams are request-response. This is also the behavior unless we specify something.

PR #192, opened 2021-02-13
Add transport.datagrams duplex stream + algorithm outline.

jan-ivar:
Fixes https://github.com/w3c/webtransport/issues/181. @yutakahirano I left _writeDatagrams_ and _readDatagrams_ to hook into. PTAL.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webtransport/pull/192.html" title="Last updated on Feb 17, 2021, 2:55 PM UTC (d58bfdc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/192/edaead8...jan-ivar:d58bfdc.html" title="Last updated on Feb 17, 2021, 2:55 PM UTC (d58bfdc)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -384,11 +384,11 @@ interface DuplexStream {
 
 : <dfn for="DuplexStream" attribute>readable</dfn>
 :: The getter steps are:
-1. Return [=this=].`[[readable]]`.
+   1. Return [=this=].`[[readable]]`.
 
 : <dfn for="DuplexStream" attribute>writable</dfn>
 :: The getter steps are:
-1. Return [=this=].`[[writable]]`.
+   1. Return [=this=].`[[writable]]`.
 
 # `WebTransport` Interface #  {#web-transport}
 
--- a/index.bs
+++ b/index.bs
@@ -384,11 +384,11 @@ interface DuplexStream {
 
 : <dfn for="DuplexStream" attribute>readable</dfn>
 :: The getter steps are:
-   1. Return [=this=].`[[readable]]`.
+     1. Return [=this=].`[[readable]]`.
 
 : <dfn for="DuplexStream" attribute>writable</dfn>
 :: The getter steps are:
-   1. Return [=this=].`[[writable]]`.
+     1. Return [=this=].`[[writable]]`.
 
 # `WebTransport` Interface #  {#web-transport}
 
--- a/index.bs
+++ b/index.bs
@@ -317,7 +317,7 @@ Datagrams are encrypted and congestion controlled.
 <pre class="idl">
 interface mixin DatagramTransport {
     readonly attribute unsigned short maxDatagramSize;
-    readonly attribute DuplexStream datagrams;
+    readonly attribute DatagramDuplexStream datagrams;
 };
 </pre>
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -400,13 +443,21 @@ agent MUST run the following steps:
...
    slot, initialized to a promise.
+1. Let |transport| have an
+   <dfn attribute for="WebTransport">\[[OutgoingDatagrams]]</dfn> internal slot
+   initialized to the result of [=WritableStream/creating=] a {{WritableStream}},
+   its [=WritableStream/create/writeAlgorithm=] set to [=writeDatagrams=] given
+   |transport| as a parameter.
+1. Let |transport| have an
+   <dfn attribute for="WebTransport">\[[IncomingDatagrams]]</dfn> internal
+   slot initialized to the result of <a dfn for="ReadableStream">creating</a> a
+   {{ReadableStream}}. `transport.[[IncomingDatagrams]]` is provided datagrams
```
yutakahirano, line 454: How about ...to the result of creating a ReadableStream, its pullAlgorithm set to readDatagrams given _transport_ as parameter?

jan-ivar: Because it's a push-based reader,  not pull-based.

yutakahirano: Ah, sorry I misunderstood.

index.bs
```diff
@@ -317,36 +317,79 @@ Datagrams are encrypted and congestion controlled.
...
+
+The <dfn>writeDatagrams</dfn> algorithm is given a |transport| as parameter and
+|data| as input. It is defined by running the following steps:
+1. TODO
+1. Enqueue data for datagram transmission.
+1. Return [=a promise resolved with=] undefined.
+
+# `DuplexStream` Interface #  {#duplex-stream}
+
+A <dfn interface>DuplexStream</dfn> is a generic duplex stream.
```
yutakahirano, line 363: Isn't this too generic? Do you want other specs to use this interface?

We may want to add datagram specific attributes/methods to this interface - for example, `WebTransport.setDatagramWritableQueueExpirationDuration()` proposed at #179 can be `WebTransport.datagrams.setWritableQueueExpirationDuration()`. What do you think about renaming this to a more datagram specific name, such as `Datagrams` or `WebTransportDatagrams`?


cc: @ricea

jan-ivar: Yes it's generic. I was actually surprised not to find it in the streams spec which already defines the term [duplex stream](https://streams.spec.whatwg.org/#duplex-stream). Agree we should move or specialize it, perhaps including a more generic one, if one becomes available? cc @domenic

> ... can be `WebTransport.datagrams.setWritableQueueExpirationDuration()`

Though I happen to disagree with https://github.com/w3c/webtransport/pull/179, I agree in theory. E.g. we might want to move our existing
```diff
- transport.maxDatagramSize
+ transport.datagrams.maxSize
```

domenic: Indeed, the reason one isn't defined is because you shouldn't have an object with _just_ readable/writable properties. You could put readable/writable properties directly on DatagramTransport, for example, and that would make DatagramTransport a conceptual duplex stream.

vasilvv: For what it's worth, I think `DatagramFlow` would be a good name given the terminology used elsewhere right now.

jan-ivar: @vasilvv where is that used? I was thinking `DatagramDuplexStream` would be the most straightforward and in keeping with the surrounding names.

yutakahirano (approved):
LGTM

Issue #137, opened 2020-10-01
Mapping to QUIC


aboba:
Originally submitted by Luke Curley on the web-transport-dev mailing list. 

I went through and enumerated the relevant BidirectionalStream methods and how I think they should map to QUIC:

abortReading() : send STOP_SENDING with code
abortWriting() : send RESET_STREAM with code 

writable.abort()             : send RESET_STREAM  
writable.close()             : send STREAM_FINAL  
writable.getWriter().write() : send STREAM
writable.getWriter().close() : send STREAM_FINAL
writable.getWriter().abort() : send RESET_STREAM

readable.cancel()             : send STOP_SENDING  
readable.getReader().read()   : receive STREAM or STREAM_FINAL 
readable.getReader().cancel() : send STOP_SENDING

Unless I'm mistaken, there's quite a few ways to close a stream... It's another discussion but it might be worth consolidating abortReading() and abortWriting() into WHATWG streams' cancel() and abort() respectively.

Here's what I have tested and observed in the Chrome implementation:

writable.getWriter().write() : send STREAM
writable.getWriter().close() : send STREAM_FINAL, RESET_STREAM with code 268?, and STOP_SENDING with code 6?
readable.getReader().read()  : receive STREAM or STREAM_FINAL

aboba:
Going through the WebTransport spec, here is the behavior that appears to be specified in various sections relating to closing or aborting. 

QuicTransport interfaces

BiDirectionalStream or ReceiveStream.abortReading() : send STOP_SENDING with errorCode. (see: https://wicg.github.io/web-transport/#dom-incomingstream-abortreading).  Note typo that refers to "abortwriting" instead of "abortreading" in this section. 
BiDirectionalStream or SendStream.abortWriting() : send RESET_STREAM with errorCode  (see: https://wicg.github.io/web-transport/#dom-outgoingstream-abortwriting)

QuicTransport.close()             : send CONNECTION_CLOSE with errorCode, reason (see: https://wicg.github.io/web-transport/#dom-webtransport-close) 

There is also WHATWG streams behavior from the cancel() and abort() methods which isn't mentioned in the document.  It would seem reasonable to clarify that they are mapped as you indicated: 

writable.abort()             : send RESET_STREAM  
writable.getWriter().abort() : send RESET_STREAM

readable.cancel()             : send STOP_SENDING  
readable.getReader().cancel() : send STOP_SENDING

[...]

PR #176, opened 2020-12-11
Mapping to QUIC

aboba:
Fixes https://github.com/w3c/webtransport/issues/137

***
<a href="https://pr-preview.s3.amazonaws.com/aboba/webtransport/pull/176.html" title="Last updated on Feb 3, 2021, 1:39 AM UTC (ea1a397)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/176/b8d1a94...aboba:ea1a397.html" title="Last updated on Feb 3, 2021, 1:39 AM UTC (ea1a397)">Diff</a>

Issue #94, opened 2020-01-10
Two ways to abort read side of the stream


vasilvv:
Apparently WritableStream has an abort() method, which takes a `reason` argument.  It's not entirely clear to me what type is `reason` expected to have, and what are the exact semantics, so it's not entirely clear to me how it should interact with `abortWriting()`.  I suspect we want to have only one of those at the end.

ricea:
What I did for WebSocketStream is treat `abort(reason)` specially if the reason was an object with a `code` property. So if you do `abort({code: 4000, reason: 'stuff'})` it closes with `(4000, 'stuff')`. If you do `abort({code: 4001})` it closes with `(4001, '')`, otherwise `abort({something: 'else'})` would do a normal (no code specified) close.

This is a bit different for QuicTransport because it has separate "close" and "abort" concepts, but I think a similar approach would work.

aboba:
@ricea This seems reasonable. Can be label this as "ready for PR?"

Related to Issue https://github.com/w3c/webtransport/issues/137

aboba:
What about `abortReading()` and `readable.cancel()`? Is it possible to do `readable.cancel({code: 4000, reason: 'stuff'})` ?  If so, then we may not need `abortReading()`.

lgrahl:
My understanding is that the `reason` of `cancel` on the readable stream and `abort` on the writable stream have very similar semantics, so IMO that suggestion seems appropriate.

@ricea I cannot find any documentation for `WebSocketStream` handling `reason` that way? The [WebSocketStream API Design](https://docs.google.com/document/d/1La1ehXw76HP6n1uUeks-WJGFgAnpX2tCjKts7QFJ57Y/edit#heading=h.qyzhypmbt4hn) document does not mention arguments for `abort` or `cancel` on the streams.

ricea:
> @ricea I cannot find any documentation for WebSocketStream handling reason that way? The WebSocketStream API Design document does not mention arguments for abort or cancel on the streams.

Oops, sorry about that. I intended to document it, but haven't done so yet.

yutakahirano:
We're removing `abortWriting`/`reset`.

PR #195, opened 2021-02-16
Mapping to QUIC

aboba:
Rebase of PR https://github.com/w3c/webtransport/pull/176

Fixes https://github.com/w3c/webtransport/issues/137
Related to Issue https://github.com/w3c/webtransport/issues/94

***
<a href="https://pr-preview.s3.amazonaws.com/aboba/webtransport/pull/195.html" title="Last updated on Feb 16, 2021, 2:40 PM UTC (baf631c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/195/d1e913f...aboba:baf631c.html" title="Last updated on Feb 16, 2021, 2:40 PM UTC (baf631c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -840,7 +840,9 @@ ReceiveStream includes IncomingStream;
 
 # Protocol Mappings # {#protocol-mapping}
 
-This section specifies the QUIC protocol actions expected from methods defined
+*This section is non-normative.*
+
+This section describes the QUIC protocol behavior of methods defined
 in this specification, utilizing [[WEB-TRANSPORT-HTTP3]].
 
   <table class="data">
--- a/index.bs
+++ b/index.bs
@@ -841,7 +841,7 @@ ReceiveStream includes IncomingStream;
 # Protocol Mappings # {#protocol-mapping}
 
 This section specifies the QUIC protocol actions expected from methods defined
-in this specification, utilizing [[HTTP3-TRANSPORT]].
+in this specification, utilizing [[WEB-TRANSPORT-HTTP3]].
 
   <table class="data">
     <colgroup class="header"></colgroup>
--- a/index.bs
+++ b/index.bs
@@ -838,6 +838,64 @@ interface ReceiveStream {
 ReceiveStream includes IncomingStream;
 </pre>
 
+# Protocol Mappings # {#protocol-mapping}
+
+This section specifies the QUIC protocol actions expected from methods defined
+in this specification, utilizing [[HTTP3-TRANSPORT]].
+
+  <table class="data">
+    <colgroup class="header"></colgroup>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -838,6 +838,64 @@ interface ReceiveStream {
 ReceiveStream includes IncomingStream;
 </pre>
 
+# Protocol Mappings # {#protocol-mapping}
+
+This section specifies the QUIC protocol actions expected from methods defined
+in this specification, utilizing [[WEB-TRANSPORT-HTTP3]].
```
jan-ivar, line 844: I think this section needs to be non-normative.

I see some value in it as an overview of typical behavior, but the normative steps are more involved and need to be provided in algorithms for bi and uni streams, similar to what is being done for datagrams in https://github.com/w3c/webtransport/pull/192#issue-572955808 which has hooks for `writeDatagrams` and `readDatagrams` that I believe @yutakahirano is working on in patches.

Once such patches are merged, this section would be competing normatively otherwise.

index.bs
```diff
@@ -838,6 +838,64 @@ interface ReceiveStream {
...
+        <th>QUIC Protocol Action</th>
+      </tr>
+    </thead>
+    <tbody>
+      <tr>
+        <td>{{IncomingStream/abortReading}}</td>
+        <td>send STOP_SENDING with code</td>
+      </tr>
+      <tr>
+        <td>{{OutgoingStream/abortWriting}}</td>
```
jan-ivar, line 861: Also, this top-level section links to `IncomingStream` and `OutgoingStream`, but doesn't mention datagrams, so it seems a bit unclear how broadly it applies.

I'm not sure how this structure would accommodate a solution to that.

jan-ivar (changes requested):

index.bs
```diff
@@ -838,6 +838,64 @@ interface ReceiveStream {
...
+      <tr>
+        <td>{{OutgoingStream/writable}}.abort()</td>
+        <td>send RESET_STREAM</td>
+      </tr>
+      <tr>
+        <td>{{OutgoingStream/writable}}.close()</td>
+        <td>send STREAM_FINAL</td>
+      </tr>
+      <tr>
+        <td>{{OutgoingStream/writable}}.getWriter().write()</td>
```
jan-ivar, line 873: For example: This may be ok to say as an example, but as a normative statement wouldn't cover e.g. using `pipeTo` (or even in some instances `pipeThrough` even) for instance, since a [writer](https://streams.spec.whatwg.org/#writer) is not involved then.

index.bs
jan-ivar: Bernard to open an issue on what happens if you abort or cancel datagrams

PR #211, opened 2021-02-28
 Add transport.bi/unidirectionalStreams duplex streams.

### LINKED ISSUE #40: Using WHATWG streams for stream creation
The current explainer now uses WHATWG streams for datagrams and for the reading/writing of data for a given uindirectional streams and bidirectional streams.  It also uses a stream of streams for receiving streams (so backpressure could be applied if one did not receive a large number of incoming streams fast enough).  

But there is one place we don't use streams: creating streams (either unidirectional or bidirectional).  

Some options:

A.  Don't use streams of streams for creating streams and just apply backpressure to the streams. So while you could have tons of streams created, writes on them all get backpressure.  Presumably an app would know not to have too many pending streams (that are receiving back pressure).  This would leave the API a big asymmetric between send and receive (making it harder to pipe a stream of messages into a stream-per-message approach), but it works well for both unidirectional and bidirectional streams.

B. Use a stream of streams for creation of unidirectional send streams where you pass in a ReadableStream that gets is used to write to the send stream.  This has the oddity that you're using a read stream to write, but one can trivially make a ReadableStream out of a byte array, so this makes it easy to do a "stream-per-message" approach.  I don't see how it would work for bidirectional streams, but a "stream-per-message" approach will likely only want unidirectional streams anyway.

C.  Don't use streams of streams for creating streams, but apply backpressure on creation using the fact that creating streams is async.  Just don't resolve the promise until stream creation is unblocked.  This seems like a natural thing to do.  Like A, it's not ideal for stream-per-message approaches, but I believe one could implement a B on top of C with some JS.

Right now, I'm inclined toward doing both C and maybe B.  C seems to work well for most use cases but B seems to offer more symmetry and also is better for stream-per-message approaches.  And even though I think you could do it with JS on top, there might be performance and convenience reasons to make it part of the API.

jan-ivar:
Fixes #40.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webtransport/pull/211.html" title="Last updated on Feb 28, 2021, 8:43 PM UTC (eb72eb1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/211/e4c4d50...jan-ivar:eb72eb1.html" title="Last updated on Feb 28, 2021, 8:43 PM UTC (eb72eb1)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -154,10 +154,34 @@ unreliability.  All stream data is encrypted and congestion-controlled.
 <pre class="idl">
 interface mixin UnidirectionalStreamsTransport {
   Promise&lt;SendStream&gt; createUnidirectionalStream(optional SendStreamParameters parameters = {});
-  /* a ReadableStream of ReceiveStream objects */
-  readonly attribute ReadableStream incomingUnidirectionalStreams;
+  readonly attribute UnidirectionalStreamDuplexStream unidirectionalStreams;
 };
 </pre>
+
+## Attributes ##  {#unidirectional-streams-transport-attributes}
+
+: <dfn for="UnidirectionalStreamsTransport" attribute>unidirectionalStreams</dfn>
+:: A single duplex stream for sending and receiving unidirectional streams
+   over this connection.
+
+   The {{UnidirectionalStreamDuplexStream/readable}} member is a
+   {{ReadableStream}} of {{ReceiveStream}} objects. Each unidirectional stream
+   received from the remote host will be surfaced as a single {{ReceiveStream}}
+   object.
+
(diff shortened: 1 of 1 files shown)


aboba:
@jan-ivar Build validation error (due to 404 on draft-ietf-webtrans-http3) should be fixed now.

yutakahirano:
I feel that having two ways to create a unidirectional (outgoing) stream is a bit confusing.

I also want to understand the role of `SendStreamParameters` here. Perhaps it should be a stream of pair<WritableStream, SendStreamParameters>.

jan-ivar:
Closing for lack of interest.

PR #244, opened 2021-04-02
[Editorial] Tidy up "internal slots"

### LINKED ISSUE #222: Editorial: Define internal slots outside of algorithms
Pointed out by @annevk in https://github.com/w3c/webtransport/pull/217.

> It seems bad to define internal slots inside an algorithm... They should be defined under the object they belong to.

I agree with that. I'm imagining something like https://streams.spec.whatwg.org/#rs-internal-slots.

### LINKED ISSUE #223: Editorial: "internal slot"
The term "internal slot" is used very much in the current spec, but in various ways.

1. An internal slot is a slot on which a value is stored. This is the most formal use I think.
   - "Let _stream_ have a __\[[Writable]]__ internal slot initizlied to a new `WritableStream`."
   -  "On getting it MUST return the value of the `IncomingStream`'s __\[[Readable]]__ slot."
   - "Return the value of the __\[[ReceivedStreams]]__ internal slot."
2. Sometimes it means the value stored in it.
   - "Remove the stream from the _transport_’s __\[[OutgoingStreams]]__ internal slot."
3. Sometimes "internal slot" is missing.
   - "If _transport_’s __\[[WebTransportState]]__ is `"closed"` or `"failed"`, then abort these steps.

I prefer the third way - given we enclose every internal slot name with "\[[]]", and we are interested in values rather than slots (except for definition - please see also #222), we can remove "internal slot" occurrences except for internal slot definitions, and the spec will be less repetitive and more readable I think.

Just for reference, the streams spec defines a lot of internal slots (e.g., https://streams.spec.whatwg.org/#rbs-controller-internal-slots) but they are called without "internal slot" (e.g., "Let _stream_ be this.__\[[stream]]__." in https://streams.spec.whatwg.org/#rbs-controller-private-pull).

yutakahirano:
- Stop defining internal slots in the constructor and have tables
   for internal slots instead.
 - Remove "internal slots" words as they are redundant.
 - Replace \{{[[slot]]}} with \[=[[slot]]=].

Fixes #222 
Fixes #223

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/pull/244.html" title="Last updated on Apr 13, 2021, 9:10 PM UTC (5887633)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/244/87bcbf1...5887633.html" title="Last updated on Apr 13, 2021, 9:10 PM UTC (5887633)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -461,7 +461,7 @@ A {{WebTransport}} object has the following internal slots.
  <tbody>
   <tr>
    <td><dfn>\[[OutgoingStreams]]</dfn>
-   <td class="non-normative">A sequence of {{OutgoingStream}} objects.
+   <td class="non-normative">A sequence of {{SendStream}} objects.
   </tr>
   <tr>
    <td><dfn>\[[ReceivedStreams]]</dfn>
--- a/index.bs
+++ b/index.bs
@@ -465,7 +465,7 @@ A {{WebTransport}} object has the following internal slots.
   </tr>
   <tr>
    <td><dfn>\[[ReceivedStreams]]</dfn>
-   <td class="non-normative">A {{ReadableStream}} consisting of {{IncomingStream}} objects.
+   <td class="non-normative">A {{ReadableStream}} consisting of {{ReceiveStream}} objects.
   </tr>
   <tr>
    <td><dfn>\[[ReceivedBidirectionalStreams]]</dfn>
--- a/index.bs
+++ b/index.bs
@@ -527,36 +527,36 @@ agent MUST run the following steps:
 1. If |dedicated| is false and |serverCertificateFingerprints| is non-null, then [=throw=] a
    {{TypeError}}.
 1. Let |transport| be a newly constructed {{WebTransport}} object, with:
-: [=[[OutgoingStreams]]=]
-:: empty
-: [=[[ReceivedStreams]]=]
-:: a new {{ReadableStream}}
-: [=[[ReceivedBidirectionalStreams]]=]
-:: a new {{ReadableStream}}
-: [=[[State]]=]
-:: `"connecting"`
-: [=[[Ready]]=]
-:: a new promise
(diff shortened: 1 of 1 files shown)


yutakahirano:
@jan-ivar @vasilvv can you take a look?

index.bs
```diff
@@ -447,6 +447,69 @@ WebTransport includes BidirectionalStreamsTransport;
...
+  </tr>
+ </thead>
+ <tbody>
+  <tr>
+   <td><dfn>\[[OutgoingStreams]]</dfn>
+   <td class="non-normative">A sequence of {{OutgoingStream}} objects.
+  </tr>
+  <tr>
+   <td><dfn>\[[ReceivedStreams]]</dfn>
+   <td class="non-normative">A {{ReadableStream}} consisting of {{IncomingStream}} objects.
```
jan-ivar, line 468:
```suggestion
   <td class="non-normative">A {{ReadableStream}} consisting of {{ReceiveStream}} objects.
```

index.bs
```diff
@@ -447,6 +447,69 @@ WebTransport includes BidirectionalStreamsTransport;
...
+ <thead>
+  <tr>
+   <th>Internal Slot
+   <th>Description (<em>non-normative</em>)
+  </tr>
+ </thead>
+ <tbody>
+  <tr>
+   <td><dfn>\[[OutgoingStreams]]</dfn>
+   <td class="non-normative">A sequence of {{OutgoingStream}} objects.
```
jan-ivar, line 464:
```suggestion
   <td class="non-normative">A sequence of {{SendStream}} objects.
```

jan-ivar (commented):
I missed two instances in the merge

jan-ivar (approved):
👌 just two nits.

yutakahirano:
Thank you!

Issue #242, opened 2021-03-31
Rename abortWriting() and remove abortReading()?


jan-ivar:
I find it hard to distinguish [abortWriting](https://w3c.github.io/webtransport/#dom-outgoingstream-abortwriting) from [abort](https://streams.spec.whatwg.org/#ws-abort) or intuit [what it does differently](https://github.com/w3c/webtransport/issues/213#issuecomment-789411510) or how they overlap (if they do):

```js
const writable = await wt.createUnidirectionalStream();
// ...
writable.abort(1);
writable.abortWriting({errorCode: 1});
```
We should clarify the difference in the spec, and explain any overlap. The [mappings table](https://w3c.github.io/webtransport/#protocol-mapping) doesn't say whether `1` is used.

It's also a naming problem, since "writing" doesn't add much context. `abortSending()`, `stopSending()`, `abruptAbort()`?

Ditto [abortReading](https://w3c.github.io/webtransport/#dom-incomingstream-abortreading) and [cancel](https://streams.spec.whatwg.org/#rs-cancel):
```js
const {readable} = await wt.createBidirectionalStream();
// ...
readable.cancel(0);
readable.abortReading({errorCode: 0});
```
Here I don't actually know what the difference is. @ricea? `abortReceiving()`, `stopReceiving()`? Remove it?

ricea:
Yeah, I actually prefer `reset()` personally. Or maybe `drop()` since it may drop data.

jan-ivar:
@ricea You mean instead of `abortWriting()`? `reset()` might be confusing, since `abort()` also [resets](https://w3c.github.io/webtransport/#protocol-mapping), and neither is a subset of the other. I like `drop()`.

Can we remove *abortReading* then?

ricea:
> @ricea You mean instead of `abortWriting()`? `reset()` might be confusing, since `abort()` also [resets](https://w3c.github.io/webtransport/#protocol-mapping), and neither is a subset of the other. I like `drop()`.

Yes, instead of `abortWriting()`.

> Can we remove _abortReading_ then?

I don't know of any interesting semantics that _abortReading_ has that can't be done with `cancel()`, but maybe someone else here does?

aboba:
@ricea I like `reset()` as well.  In fact, as I recall that was the original name of the method.

martinthomson:
`reset(code : u64, direction : { "read", "write", "both" })` ?

PR #248, opened 2021-04-13
Rename abortWriting()/writingAborted → reset()/remoteCanceled & abortReading()/readingAborted → cancel()/remoteReset

jan-ivar:
Fixes https://github.com/w3c/webtransport/issues/242.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webtransport/pull/248.html" title="Last updated on May 13, 2021, 11:49 PM UTC (dc1790c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/248/a19a918...jan-ivar:dc1790c.html" title="Last updated on May 13, 2021, 11:49 PM UTC (dc1790c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1081,6 +1081,7 @@ queue a task to run the following:
          from.
       1. Remove |stream| from the |transport|'s [=[[OutgoingStreams]]=].
       1. Let |reason| be the value from the STOP_SENDING frame from the remote side.
+      1. Perform ! [$WritableStreamStartErroring$](|stream|, |reason|).
       1. [=Resolve=] {{SendStream}}.{{SendStream/remoteCanceled}} with a new
          {{StreamAbortInfo}} with the {{StreamAbortInfo/errorCode}} set to
          |reason|.
@@ -1145,7 +1146,7 @@ queue a task to run the following:
       1. Let |transport| be the {{WebTransport}}, which the |stream| was created
          from.
       1. Let |reason| be the value from the message from the remote side.
-      1. Error |stream| with |reason| (TODO).
+      1. Perform ! [$ReadableStreamError$](|stream|, |reason|).
 
 The {{ReceiveStream}}'s <dfn>cancelAlgorithm</dfn> is:
      1. Let |stream| be the {{ReceiveStream}} object.
--- a/index.bs
+++ b/index.bs
@@ -1038,46 +1038,67 @@ that can be written to, to transmit data to the remote host.
 <pre class="idl">
 [ Exposed=(Window,Worker) ]
 interface SendStream : WritableStream /* of Uint8Array */ {
-  readonly attribute Promise&lt;StreamAbortInfo&gt; writingAborted;
-  undefined abortWriting(optional StreamAbortInfo abortInfo = {});
+  readonly attribute Promise&lt;StreamAbortInfo&gt; dropped;
+  undefined drop(optional StreamAbortInfo abortInfo = {});
 };
 </pre>
 
-## Overview ##  {#outgoing-stream-overview}
+## Overview ##  {#send-stream-overview}
 
 The {{SendStream}} will be initialized by running the {{WritableStream}}
 initialization steps.
 
-## Attributes ##  {#outgoing-stream-attributes}
+## Attributes ##  {#send-stream-attributes}
 
-: <dfn attribute for="SendStream">writingAborted</dfn>
-:: The `writingAborted` attribute represents a promise that is [=fulfilled=]
+: <dfn attribute for="SendStream">dropped</dfn>
+:: The `dropped` attribute represents a promise that is [=fulfilled=]
     when the a message from the remote side aborting the stream is received.
-    For QUIC, that message is a STOP_SENDING frame. When the stream receives
-    this mesage, the user agent MUST run the following:
-      1. Let |stream| be the {{SendStream}} object.
-      1. Let |transport| be the {{WebTransport}}, which the |stream| was created
-         from.
-      1. Remove the stream from the |transport|'s [=[[OutgoingStreams]]=].
-      1. [=Resolve=] the promise with the resulting {{StreamAbortInfo}} with
-         the {{StreamAbortInfo/errorCode}} set to the value from the aborting
-         message from the remote side.
+    For QUIC, that message is a STOP_SENDING frame.
 
-## Methods ##  {#outgoing-stream-methods}
+## Methods ##  {#send-stream-methods}
 
-: <dfn method for="SendStream">abortWriting()</dfn>
+: <dfn method for="SendStream">drop()</dfn>
 :: A hard shutdown of the {{SendStream}}. It may be called regardless of
    whether the {{SendStream}} was created by the local or remote peer. When
-   the `abortWriting` method is called, the user agent MUST run the following
+   the `drop` method is called, the user agent MUST run the following
    steps:
      1. Let |stream| be the {{SendStream}} object which is about to abort
         writing.
      1. Let |transport| be the {{WebTransport}}, which the |stream| was created
         from.
-     1. Remove the stream from the transport's [=[[OutgoingStreams]]=].
+     1. Remove |stream| from the transport's [=[[OutgoingStreams]]=].
      1. Let |abortInfo| be the first argument.
-     1. Start the closing procedure by sending a RST_STREAM frame with its
-        error code set to the value of |abortInfo.errorCode|.
+     1. Let |reason| be |abortInfo|.errorCode.
+     1. If it hasn't started already, start the closing procedure by sending a
+        RST_STREAM frame with its error code set to the value of |reason|.
+     1. Let |p| be the result of [=WritableStream/abort|aborting=] |stream|, with |reason|.
+     1. [=Upon fulfillment=] or [=upon rejection|rejection=] of |p|, do nothing.
+
+## Procedures ##  {#send-stream-procedures}
+
+When a {{SendStream}} receives a STOP_SENDING frame from the remote side,
+the user agent MUST queue a task to run the following:
+      1. Let |stream| be the {{SendStream}} object.
+      1. Let |transport| be the {{WebTransport}}, which the |stream| was created
+         from.
+      1. Remove |stream| from the |transport|'s [=[[OutgoingStreams]]=]
+         internal slot.
+      1. Let |reason| be the value from the STOP_SENDING frame from the remote side.
+      1. Let |p| be the result of [=WritableStream/abort|aborting=] |stream|, with |reason|.
+      1. [=Upon fulfillment=] or [=upon rejection|rejection=] of |p|, do nothing.
+      1. [=Resolve=] {{SendStream}}.{{SendStream/dropped}} with a new
+         {{StreamAbortInfo}} with the {{StreamAbortInfo/errorCode}} set to
+         |reason|.
+
+The {{SendStream}}'s <dfn>abortAlgorithm</dfn> (TODO) is:
+     1. Let |stream| be the {{ReceiveStream}} object.
+     1. Let |transport| be the {{WebTransport}}, which the |stream| was created
+        from.
+     1. Let |reason| be the first argument, if it is numeric, or 0.
+     1. If |stream| was aborted by a STOP_SENDING frame, abort these steps.
+     1. If it hasn't started already, start the closing procedure by sending a
+        STOP_SENDING frame to the remote side, with its error code set to the
+        value of |reason|.
 
 ## `StreamAbortInfo` Dictionary ##  {#stream-abort-info}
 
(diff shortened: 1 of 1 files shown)


aboba (commented):
`abortReading()` is being removed because it is considered a duplicate of `cancel(reason)`, correct? 
lin 904: With `readingAborted` removed, how does the application know that a `RST_STREAM` has been received with an errorCode?

line 796/797: I don't find "drop/dropped" to be more informative than "abortWriting/writingAborted". 
line 810:  HTTP/2 fallback would not involve a `STOP_SENDING` frame.  So being specific about Http3Transport would be better. 
Line 828: HTTP/2 fallback would not involve a `RST_STREAM` frame. So being specific about Http3Transport would be better. 
line 974, 990, 994, 1002: can you add the code argument?

jan-ivar:
@domenic do you know if the `reason` when calling `cancel(reason)` on a stream is surfaced anywhere else than in the promise it immediately returns?

domenic:
It is sent to the underlying source, which can interpret it as desired.

index.bs
```diff
@@ -853,6 +846,32 @@ The dictionary SHALL have the following fields:
...
 
+## Procedures ##  {#outgoing-stream-procedures}
+
+When an {{OutgoingStream}} receives a STOP_SENDING frame from the remote side,
+the user agent MUST queue a task to run the following:
+      1. Let |stream| be the {{OutgoingStream}} object.
+      1. Let |transport| be the {{WebTransport}}, which the |stream| was created
+         from.
+      1. Remove |stream| from the |transport|'s {{[[OutgoingStreams]]}}
+         internal slot.
```
yutakahirano, line 857: Let's remove "internal slot" as discussed at #223.

index.bs
```diff
@@ -853,6 +846,32 @@ The dictionary SHALL have the following fields:
...
+      1. Let |reason| be the value from the STOP_SENDING frame from the remote side.
+      1. Let |p| be the result of [=WritableStream/abort|aborting=] |stream|, with |reason|.
+      1. [=Upon fulfillment=] or [=upon rejection|rejection=] of |p|, do nothing.
```
yutakahirano, line 860: Why do we need this step?

index.bs
```diff
@@ -805,36 +805,29 @@ initialization steps.
...
+     1. Remove |stream| from the transport's {{[[OutgoingStreams]]}} internal
         slot.
      1. Let |abortInfo| be the first argument.
-     1. Start the closing procedure by sending a RST_STREAM frame with its
-        error code set to the value of |abortInfo.errorCode|.
+     1. Let |reason| be |abortInfo|.errorCode.
+     1. If it hasn't started already, start the closing procedure by sending a
+        RST_STREAM frame with its error code set to the value of |reason|.
+     1. Let |p| be the result of [=WritableStream/abort|aborting=] |stream|, with |reason|.
+     1. [=Upon fulfillment=] or [=upon rejection|rejection=] of |p|, do nothing.
```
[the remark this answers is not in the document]
ricea: Maybe we should do `Set p.[[PromiseIsHandled]] to true.` like the Stream Standard does in a few places?

index.bs
```diff
@@ -853,6 +847,32 @@ The dictionary SHALL have the following fields:
 : <dfn for="StreamAbortInfo" dict-member>errorCode</dfn>
 :: The error code. The default value of 0 means "CLOSING."
 
+## Procedures ##  {#outgoing-stream-procedures}
+
+When an {{OutgoingStream}} receives a message from the remote side aborting the
+stream (For QUIC, message is a STOP_SENDING frame), the user agent MUST queue a
```
martinthomson, line 853:
```suggestion
stream (for QUIC, message is a STOP_SENDING frame), the user agent MUST queue a
```

index.bs
```diff
@@ -873,37 +891,28 @@ interface mixin IncomingStream : ReadableStream /* of Uint8Array */ {
...
+## Procedures ##  {#incoming-stream-procedures}
 
-: <dfn attribute for="IncomingStream">readingAborted</dfn>
-:: The `readingAborted` attribute represents a promise that is [=fulfilled=]
-    when the a message from the remote side aborting the stream is received.
-    For QUIC, that message is a RST_STREAM frame. When the stream receives
-    this mesage, the user agent MUST run the following:
-      1. Let |stream| be the {{IncomingStream}} object for which the abort
+When an {{IncomingStream}} receives a message from the remote side aborting the
+stream (For QUIC, that message is an RST_STREAM frame), the user agent MUST
```
jan-ivar, line 897:
```suggestion
stream (for QUIC, that message is an RST_STREAM frame), the user agent MUST
```

index.bs
```diff
@@ -853,6 +847,32 @@ The dictionary SHALL have the following fields:
 : <dfn for="StreamAbortInfo" dict-member>errorCode</dfn>
 :: The error code. The default value of 0 means "CLOSING."
 
+## Procedures ##  {#outgoing-stream-procedures}
+
+When an {{OutgoingStream}} receives a message from the remote side aborting the
+stream (for QUIC, message is a STOP_SENDING frame), the user agent MUST queue a
```
jan-ivar, line 853:
```suggestion
stream (for QUIC, that message is a STOP_SENDING frame), the user agent MUST queue a
```

yutakahirano:
Reg: `remoteReset`, I think we can add properties to the exception with which `closed` is rejected, instead of introducing `remoteReset`. What do you think about the idea?

jan-ivar:
Rebased

jan-ivar:
> Reg: remoteReset, I think we can add properties to the exception with which closed is rejected, instead of introducing remoteReset. What do you think about the idea?

Good idea! I've opened https://github.com/w3c/webtransport/issues/252 on it.

In the meantime I've updated this PR to s/readingAborted/remoteReset/ as well (we can reuse the same name on readable and writable now, since we got rid of mixins).

This seems like a reasonable incremental step, so I'm hoping we can merge this PR and treat #252 as a separate idea. PTAL

index.bs
```diff
@@ -931,46 +931,67 @@ that can be written to, to transmit data to the remote host.
...
+         from.
+      1. Remove |stream| from the |transport|'s [=[[OutgoingStreams]]=]
+         internal slot.
+      1. Let |reason| be the value from the STOP_SENDING frame from the remote side.
+      1. Let |p| be the result of [=WritableStream/abort|aborting=] |stream|, with |reason|.
+      1. Set |p|.[=[[PromiseIsHandled]]=] to true.
+      1. [=Resolve=] {{SendStream}}.{{SendStream/remoteReset}} with a new
+         {{StreamAbortInfo}} with the {{StreamAbortInfo/errorCode}} set to
+         |reason|.
+
```
yutakahirano, line 985: Shouldn't we error the stream?

index.bs
```diff
@@ -931,46 +931,67 @@ that can be written to, to transmit data to the remote host.
...
-        error code set to the value of |abortInfo.errorCode|.
+     1. Let |reason| be |abortInfo|.errorCode.
+     1. If it hasn't started already, start the closing procedure by sending a
+        RST_STREAM frame with its error code set to the value of |reason|.
+     1. Let |p| be the result of [=WritableStream/abort|aborting=] |stream|, with |reason|.
+     1. Set |p|.[=[[PromiseIsHandled]]=] to true.
+
+## Procedures ##  {#send-stream-procedures}
+
+When a {{SendStream}} receives a STOP_SENDING frame from the remote side,
```
yutakahirano, line 972: Shouldn't this be RST_STREAM?

index.bs
jan-ivar: @yutakahirano Sorry, the rebase was quite painful and "internal slot" must have snuck back in. I like the shorter format. I'll remember to squash next time to avoid bugs like this.

index.bs
```diff
@@ -931,46 +931,67 @@ that can be written to, to transmit data to the remote host.
...
         writing.
      1. Let |transport| be the {{WebTransport}}, which the |stream| was created
         from.
-     1. Remove the stream from the transport's [=[[OutgoingStreams]]=].
+     1. Remove |stream| from the transport's [=[[OutgoingStreams]]=].
      1. Let |abortInfo| be the first argument.
-     1. Start the closing procedure by sending a RST_STREAM frame with its
-        error code set to the value of |abortInfo.errorCode|.
+     1. Let |reason| be |abortInfo|.errorCode.
+     1. If it hasn't started already, start the closing procedure by sending a
```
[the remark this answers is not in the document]
jan-ivar: It is, but my understanding is that [aborting](https://streams.spec.whatwg.org/#writable-stream-abort) in the streams spec is not guaranteed to be immediate (it's not possible to abort a stream that's already been closed for instance), which is the reason for having `reset()`.

So we're starting the closing procedure opportunistically from both places to have it happen sooner rather than later.

index.bs
jan-ivar: According to [writingAborted](https://w3c.github.io/webtransport/#dom-sendstream-writingaborted) it's STOP_SENDING. Is that not correct?

index.bs
jan-ivar: Isn't line 980 (two steps above) doing that?
```
Let |p| be the result of [=WritableStream/abort|aborting=] |stream|, with |reason|.
```

index.bs
```diff
@@ -931,46 +931,63 @@ that can be written to, to transmit data to the remote host.
...
+   the `reset` method is called, the user agent MUST run the following
    steps:
-     1. Let |stream| be the {{SendStream}} object which is about to abort
-        writing.
+     1. Let |transport| be the {{WebTransport}}, which [=this=] stream was created
+        from.
+     1. Remove [=this=] from the transport's [=[[OutgoingStreams]]=].
+     1. Let |reason| be |abortInfo|.errorCode.
+     1. If it hasn't started already, start the closing procedure by sending a
+        RST_STREAM frame with its error code set to the value of |reason|.
```
yutakahirano, line 963: nit; s/RST_STREAM/RESET_STREAM/

index.bs
```diff
@@ -931,46 +931,63 @@ that can be written to, to transmit data to the remote host.
...
+      1. Remove |stream| from the |transport|'s [=[[OutgoingStreams]]=].
+      1. Let |reason| be the value from the STOP_SENDING frame from the remote side.
+      1. Let |p| be the result of [=WritableStream/abort|aborting=] |stream|, with |reason|.
+      1. Set |p|.[=[[PromiseIsHandled]]=] to true.
+      1. [=Resolve=] {{SendStream}}.{{SendStream/remoteReset}} with a new
+         {{StreamAbortInfo}} with the {{StreamAbortInfo/errorCode}} set to
+         |reason|.
+
+The {{SendStream}}'s <dfn>abortAlgorithm</dfn> (TODO) is:
+     1. Let |stream| be the {{ReceiveStream}} object.
```
yutakahirano, line 983: nit; SendStream

yutakahirano:
Hmm, I'm confused (again).

Sending a `STOP_SENDING` to a ReceiveStream makes sense, sending a `STOP_SENDING` to a SendStream doesn't.
Isn't what you want to do removing abortWriting, not abortReading? The mapping table would be:


|interface|method|action|
| ---- | ----      |  ----  |
|ReceiveStream|cancel|Send `STOP_SENDING` with code |
|ReceiveStream|reset  | Send `RESET_STREAM` with code |
|SendStream|abort|Send `RESET_STREAM` with code |

OR, as commented [above](https://github.com/w3c/webtransport/pull/248#discussion_r627373628), we can remove `reset` by adding a field to the reason argument.

|interface|method|action|
| ---- | ----      |  ----  |
|ReceiveStream|cancel|Send `STOP_SENDING` or `RESET_STREAM` with code|
|SendStream|abort|Send `RESET_STREAM` with code |

index.bs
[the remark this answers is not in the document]
jan-ivar: > ```
> 3. Return ! WritableStreamAbort(this, abortInfo).
> ```

@yutakahirano My understanding is the whole point of `reset()` is that it works immediately where `abort()` wouldn't (e.g. when `"closed"`), so we can't rely on [WritableStreamAbort](https://streams.spec.whatwg.org/#writable-stream-abort), because among other things it says: _"2. If state is "closed" or "errored", return a promise resolved with undefined."_

Regarding checking if the stream is locked, this rename-or-remove PR tries to preserve how [abortWriting](https://w3c.github.io/webtransport/#dom-sendstream-abortwriting) is spec'ed to work today, while making the implied impact on streams a bit more explicit. So let's open a new issue on that.

> Alternatively, we could unify abort and reset by adding one more field to the "reason" argument.

We'd still need `reset()` for the same reason that `abort()` won't let us reset closed streams (which still may be flushing data the way the streams spec is designed).

> ```
> 1. Send a STOP_SENDING frame with |reason|.
> ```

Whoops, that was meant to be RESET_STREAM. My bad. I think my mistake confused you. Let me fix it and we start over.

Also, `remoteReset` on the sender is confusing. It should be `remoteCancelled` (since the remote end is a ReceiveStream with a cancel method).

jan-ivar:
> Hmm, I'm confused (again).

I think so (my fault). Sending a `STOP_SENDING` to a SendStream makes sense. It's the other one that was nonsense.

index.bs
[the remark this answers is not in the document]
yutakahirano: > My understanding is the whole point of reset() is that it works immediately where abort() wouldn't (e.g. when "closed"), so we can't rely on WritableStreamAbort, because among other things it says: "2. If state is "closed" or "errored", return a promise resolved with undefined."

WritableStream's state becomes "closed" or "errored" when the closing or erroring operation starts running, not when it is queued. You can run the following code


```
ws = new WritableStream({
  write(chunk) {
    console.log(`write(${chunk})`);
    return new Promise(resolve => setTimeout(resolve, 5000));
  },
  close(reason) {
    console.log(`close(${reason})`);
    return new Promise(resolve => setTimeout(resolve, 5000));
  },
  abort(reason) {
    console.log(`abort(${reason})`);
    return new Promise(resolve => setTimeout(resolve, 5000));
  }
}); writer = ws.getWriter();

writer.write(1); writer.write(2); writer.write(3); writer.close(); writer.abort();
```

and you will see that abort operation runs before the second chunk arrvies.

Your concern is reasonable for the [inflight write request](https://streams.spec.whatwg.org/#writablestream-inflightwriterequest) - In the above example, the abort operation waits for the first write request to be completed. https://github.com/whatwg/streams/issues/1015 may be a solution for the problem.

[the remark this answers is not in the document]
jan-ivar: @yutakahirano makes sense.

Meeting:
* Merge PR for forward progress to not block on future nice things, but with the intent that we would eventually be able to remove these methods in favor of making `abort` in the streams spec work for us.

jan-ivar:
Meeting:
* Merge PR for forward progress to not block on future nice things, but with the intent that we would eventually be able to remove these methods in favor of making `abort` in the streams spec work for us.

jan-ivar:
@domenic is [this](https://github.com/w3c/webtransport/pull/248/commits/e7e5d75122ab4dbc2de04280354dabf4f9ff6bc3) the right way to error a stream?

index.bs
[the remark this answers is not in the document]
yutakahirano: Thank you!

I filed #260. Let's discuss this there.

domenic:
> @domenic is this the right way to error a stream?

Not quite. In general you should use the algorithms from https://streams.spec.whatwg.org/#other-specs , so in this case https://streams.spec.whatwg.org/#readablestream-error and ... oops, I didn't create such a wrapper for WritableStream, let me do so right now. (It will be a wrapper around WritableStreamDefaultControllerError though, not StartErroring.

jan-ivar:
@yutakahirano does it look good now?

jan-ivar:
> ... I've reinstated `readingAborted` and calling it `remoteReset` for now, until we can figure out the right mapping in #257. 

With our [decision](https://github.com/w3c/webtransport/issues/257#issuecomment-838596732) to error the stream on reception of STOP_SENDING and RESET_STREAM, the remoteCanceled and remoteReset promises seem redundant. Can we remove them?

yutakahirano:
Still LGTM.

Let's land this change and continue the discussion (at #260 for example).

jan-ivar:
> Still LGTM.
> 
> Let's land this change and continue the discussion (at #260 for example).

I'll take this as a ✅ and merge this.

PR #253, opened 2021-04-30
Define WebTransport session and stream

### LINKED ISSUE #250: Is it a good idea to define WebTransport "session"?
Hi,

Currently we have "connection" in our spec but we don't have "session".  We're going to specify datagrams and streams in more details, and I think "session" is useful concept for various reasons:

 1. We'll be able to group together many protocol related actions and notifications.
 2. "session" is a key concept in the protocol, and having the corresponding concept here will help developers who read both specs.
 3. Without the concept we always need to think about "session ID" which is purely tiresome for both readers and writers.

What do you think?

yutakahirano:
- Define WebTransport session and stream.
- Define capabilities and map them with the protocol spec.
- Fix existing use of the capabilities.
   - _receiveDatagrams_
   - _sendDatagrams_
   - WebTransport constructor
   - WebTransport.prototype.close

Fixes #250.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/pull/253.html" title="Last updated on May 10, 2021, 4:00 AM UTC (0568af5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/253/1c96d59...0568af5.html" title="Last updated on May 10, 2021, 4:00 AM UTC (0568af5)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -190,14 +190,9 @@ There may be multiple [=WebTransport session=]s on one [=connection=], when pool
 To <dfn for=session>establish</dfn> a [=WebTransport session=], follow [[!WEB-TRANSPORT-HTTP3]]
 [section 3.3](https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3/#section-3.3).
 
-To <dfn for=session>terminate</dfn> a [=WebTransport session=] |session| with |code|, run these
-steps:
-
-1. Terminate |session|, as described in [[!WEB-TRANSPORT-HTTP3]]
-   [section 5](https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3/#section-5),
-   with using |code| as each RESET_STREAM frame's "Application Protocol Error Code"
-   defined in [[!QUIC]]
-   [section 19.4](https://tools.ietf.org/html/draft-ietf-quic-transport#section-19.4).
+To <dfn for=session>terminate</dfn> a [=WebTransport session=] |session| with an integer |code|,
+follow [[!WEB-TRANSPORT-HTTP3]]
+[section 5](https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3/#section-5).
 
 <dfn>WebTransport stream</dfn> is a concept for HTTP/3 stream on a [=WebTransport session=].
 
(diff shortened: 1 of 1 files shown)


yutakahirano:
@vasilvv can you take a look?

yutakahirano:
@vasilvv can you take a look?

index.bs
```diff
@@ -139,6 +139,120 @@ The terms [=fulfilled=], [=rejected=], [=resolved=], [=pending=] and
...
+
+To <dfn for=session>establish</dfn> a [=WebTransport session=], follow [[!WEB-TRANSPORT-HTTP3]]
+[section 3.3](https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3/#section-3.3).
+
+To <dfn for=session>terminate</dfn> a [=WebTransport session=] |session| with |code|, run these
+steps:
+
+1. Terminate |session|, as described in [[!WEB-TRANSPORT-HTTP3]]
+   [section 5](https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3/#section-5),
+   with using |code| as each RESET_STREAM frame's "Application Protocol Error Code"
```
vasilvv, line 198: Could you remove the specific mention of RESET_STREAM?  We may change the way how this works soon ietf-wg-webtrans/draft-ietf-webtrans-http3#41.

index.bs
```diff
@@ -139,6 +139,120 @@ The terms [=fulfilled=], [=rejected=], [=resolved=], [=pending=] and
...
+   [section 3.5](https://datatracker.ietf.org/doc/html/draft-ietf-quic-transport#section-3.5)
+   <td>Yes
+   <td>No
+   <td>Yes
+  </tr>
+  <tr>
+   <td><dfn>reset</dfn> a [=WebTransport stream=]
+   <td>[[!QUIC]]
+   [section 19.4](https://datatracker.ietf.org/doc/html/draft-ietf-quic-transport#section-19.4)
+   <td>Yes
```
vasilvv, line 248: RESET_STREAM only works in the outgoing direction, so you can't use it on an incoming stream.  On the other hand, didn't we intend to merge those two into a two-way reset?

index.bs
yutakahirano: Done. Can you mention the "code" parameter in https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3/#section-5?

index.bs
yutakahirano: Fixed. I'm happy to update the description whenever https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3/ is updated, but that can be done separately.

index.bs
vasilvv: I'll add it when I rewrite it to use trailers.

vasilvv (approved):

PR #254, opened 2021-04-30
Move datagram related internal slots

yutakahirano:
...from WebTransport to DatagramDuplexStream.

This is an editorial change.


yutakahirano:
@jan-ivar can you take a look?

yutakahirano:
@jan-ivar can you take a look?

PR #255, opened 2021-05-05
Create SECURITY.md

Sfaraahfarah:


wilaw:
Security.md is meant for github repos that are hosting code and that have the potential for security vulnerabilities.  Not needed for Webtransport repo since our output is a text document. For reference to the applicability see https://docs.github.com/en/code-security/security-advisories/adding-a-security-policy-to-your-repository

Issue #221, opened 2021-03-08
Discarding stale incoming datagrams


yutakahirano:
This is a dual of #152.

When web developers do NOT read incoming datagrams from a WebTransport, the incoming datagrams are stored in the ReadableStream (i.e., `wt.datagrams.readable`). This is confusing so I'm proposing to discard stale incoming datagrams at the UnderlyingSource layer. Like #152, we can expose the expiration duration and high water mark to web developers.

ricea:
sgtm

yutakahirano:
This has been done.

Issue #210, opened 2021-02-21
UnidirectionalStream could be simpler


jan-ivar:
Conceptually, bidirectional streams ([1](https://w3c.github.io/webtransport/#dom-bidirectionalstreamstransport-createbidirectionalstream), [2](https://w3c.github.io/webtransport/#dom-bidirectionalstreamstransport-incomingbidirectionalstreams)) are inherent [duplex streams](https://streams.spec.whatwg.org/#other-specs-duplex), but unidirectional streams ([1](https://w3c.github.io/webtransport/#dom-unidirectionalstreamstransport-createunidirectionalstream), [2](https://w3c.github.io/webtransport/#dom-unidirectionalstreamstransport-incomingunidirectionalstreams)) are not. They're conceptually just [writable streams](https://streams.spec.whatwg.org/#ws-class) on sending, and just [readable streams](https://streams.spec.whatwg.org/#rs-class) on reception.

So it feels like I should be able to write:
```js
const wt = new WebTransport(url);
await readableStreamOfTextData
  .pipeThrough(new TextEncoderStream("utf-8"))
  .pipeTo(await wt.createUnidirectionalStream());
```
...instead of:
```js
const wt = new WebTransport(url);
await readableStreamOfTextData
  .pipeThrough(new TextEncoderStream("utf-8"))
  .pipeTo((await wt.createUnidirectionalStream()).writable); // clunky
```

Similarly on reception, it feels more natural to guess I would need to write:
```js
for await (const stream of wt.incomingUnidirectionalStreams) {
    await stream
     .pipeThrough(new TextDecoderStream("utf-8"))
     .pipeTo(createWritableStreamForTextData());
}
```
...instead of:
```js
for await (const stream of wt.incomingUnidirectionalStreams) {
    await stream.readable // why?
     .pipeThrough(new TextDecoderStream("utf-8"))
     .pipeTo(createWritableStreamForTextData());
}
```

A counter-argument might be symmetry with bidirectional streams — that I may want to drop in one for the other, or not care which one I used. But this seems like false symmetry. I think I'd rather have a TypeError if I did that, to catch either why no data is being received, or being dropped because I'm not handling it.

[...]

PR #267, opened 2021-05-18
Define datagram backpressure control attributes

### LINKED ISSUE #251: Add datagrams.setSendQueueExpiration(long duration)?

yutakahirano:
Define the following attributes on DatagramDuplexStream.

 - incomingMaxAge
 - outgoingMaxAge
 - incomingHighWaterMark
 - outgoingHighWaterMark

Fixes #251, #221, #152

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/pull/267.html" title="Last updated on Jun 16, 2021, 4:59 AM UTC (4c11ddc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/267/dec79ad...4c11ddc.html" title="Last updated on Jun 16, 2021, 4:59 AM UTC (4c11ddc)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -368,7 +368,7 @@ A {{DatagramDuplexStream}} object has the following internal slots.
      1. Return [=this=]'s [=[[IncomingDatagramsExpirationDuration]]=].
 :: The setter steps are:
      1. Let |value| be the given value.
-     1. If |value| > 0:
+     1. If |value| is null or |value| > 0:
        1. Set [=this=]'s [=[[IncomingDatagramsExpirationDuration]]=] to |value|.
 
 : <dfn for="DatagramDuplexStream" attribute>outgoingMaxAge</dfn>
@@ -376,7 +376,7 @@ A {{DatagramDuplexStream}} object has the following internal slots.
      1. Return [=this=]'s [=[[OutgoingDatagramsExpirationDuration]]=].
 :: The setter steps are:
      1. Let |value| be the given value.
-     1. If |value| > 0:
+     1. If |value| is null or |value| > 0:
        1. Set [=this=]'s [=[[OutgoingDatagramsExpirationDuration]]=] to |value|.
 
 : <dfn for="DatagramDuplexStream" attribute>incomingHighWaterMark</dfn>
--- a/index.bs
+++ b/index.bs
@@ -367,25 +367,33 @@ A {{DatagramDuplexStream}} object has the following internal slots.
 :: The getter steps are:
      1. Return [=this=]'s [=[[IncomingDatagramsExpirationDuration]]=].
 :: The setter steps are:
-     1. Set [=this=]'s [=[[IncomingDatagramsExpirationDuration]]=] to the given value.
+     1. Let |value| be the given value.
+     1. If |value| > 0:
+       1. Set [=this=]'s [=[[IncomingDatagramsExpirationDuration]]=] to |value|.
 
 : <dfn for="DatagramDuplexStream" attribute>outgoingMaxAge</dfn>
 :: The getter steps are:
      1. Return [=this=]'s [=[[OutgoingDatagramsExpirationDuration]]=].
 :: The setter steps are:
-     1. Set [=this=]'s [=[[OutgoingDatagramsExpirationDuration]]=] to the given value.
+     1. Let |value| be the given value.
+     1. If |value| > 0:
+       1. Set [=this=]'s [=[[OutgoingDatagramsExpirationDuration]]=] to |value|.
 
 : <dfn for="DatagramDuplexStream" attribute>incomingHighWaterMark</dfn>
 :: The getter steps are:
      1. Return [=this=]'s [=[[IncomingDatagramsHighWaterMark]]=].
 :: The setter steps are:
-     1. Set [=this=]'s [=[[IncomingDatagramsHighWaterMark]]=] to the given value.
+     1. Let |value| be the given value.
+     1. If |value| >= 0:
+       1. Set [=this=]'s [=[[IncomingDatagramsHighWaterMark]]=] to |value|.
 
 : <dfn for="DatagramDuplexStream" attribute>outgoingHighWaterMark</dfn>
 :: The getter steps are:
      1. Return [=this=]'s [=[[OutgoingDatagramsHighWaterMark]]=].
 :: The setter steps are:
-     1. Set [=this=]'s [=[[OutgoingDatagramsHighWaterMark]]=] to the given value.
+     1. Let |value| be the given value.
+     1. If |value| >= 0:
+       1. Set [=this=]'s [=[[OutgoingDatagramsHighWaterMark]]=] to |value|.
 
 ## Procedures ## {#datagram-duplex-stream-procedures}
 
--- a/index.bs
+++ b/index.bs
@@ -258,6 +258,11 @@ A <dfn interface>DatagramDuplexStream</dfn> is a generic duplex stream.
 interface DatagramDuplexStream {
   readonly attribute ReadableStream readable;
   readonly attribute WritableStream writable;
+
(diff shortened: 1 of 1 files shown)


yutakahirano:
I made set(Incoming|Outgoing)HighWaterMark methods rather than members in WebTransportOptions because it'd be useful to set them multiple times, especially when the network condition changes.

yutakahirano:
@jan-ivar @ricea can you take a look? Names may change based on the discussion at #251.

index.bs
```diff
@@ -346,6 +346,11 @@ A <dfn interface>DatagramDuplexStream</dfn> is a generic duplex stream.
 interface DatagramDuplexStream {
   readonly attribute ReadableStream readable;
   readonly attribute WritableStream writable;
+
+  undefined setIncomingExpirationDuration(double ms);
+  undefined setOutgoingExpirationDuration(double ms);
```
jan-ivar, line 351: These look like simple setters, and should therefore be read-write attributes instead of methods, I claim. E.g. using latest naming from https://github.com/w3c/webtransport/issues/251#issuecomment-843800902:
```js
wt.datagrams.incomingMaxAge = 300; // ms
wt.datagrams.outgoingMaxAge = 800; // ms
```
I'm also wondering if we should take a cue from recent changes to bidirectional streams, and subclass readable and writable stream here, so we can push the attributes down:
```js
wt.datagrams.readable.maxAge = 300; // ms
wt.datagrams.writable.maxAge = 800; // ms
```

index.bs
```diff
@@ -346,6 +346,11 @@ A <dfn interface>DatagramDuplexStream</dfn> is a generic duplex stream.
...
+  undefined setIncomingHighWaterMark(long highWaterMark);
+  undefined setOutgoingHighWaterMark(long highWaterMark);
```
jan-ivar, line 353: When it comes to highwaterMark, there's potential confusion with the [same-named concept in the streams spec](https://streams.spec.whatwg.org/#cqs-class):
```js
const stream = new WritableStream(
  { ... },
  new CountQueuingStrategy({ highWaterMark: 5 })
);
```
But I'm not 100% sure it's the same thing. That API I think is generally understood to be the buffer in the ReadableStream/WritableStream itself, distinct from its underlying source, whereas in our case, we have a setup at least on the writable side where we pre-fill the underlying sink's write-queue (for throughput and threading purposes), and what we really mean to control here is the depth of that write queue. But maybe this is a detail and that from the viewpoint of a producer (or consumer in the readable case) this is close enough to the the same thing?

jan-ivar (changes requested):

index.bs
yutakahirano: One concern is it can return null (when not set/the browser default behavior). Are you fine with the following IDL?

```
attribute double? incomingMaxAge;
attribute double? outgoingMaxAge;
```

I share [the sentiment](https://github.com/w3c/webtransport/issues/210#issuecomment-789367240) about subclassing with @ricea and I'd like to avoid that without a clear benefit. I'm not sure this is enough.

index.bs
yutakahirano: They represent the same concept, the length of the queue in a stream. We have them in the underlying source and sink because removing stale chunks is not supported in general streams. Given that I think it's good to keep the same name. Web developers cannot control the buffer lengths in the streams, but instead they can control these values.

yutakahirano:
I replaced the methods with attributes. PTAL again.

index.bs
```diff
@@ -446,6 +451,30 @@ A {{DatagramDuplexStream}} object has the following internal slots.
 :: The getter steps are:
      1. Return [=this=].`[[Writable]]`.
 
+: <dfn for="DatagramDuplexStream" attribute>incomingMaxAge</dfn>
+:: The getter steps are:
+     1. Return [=this=]'s [=[[IncomingDatagramsExpirationDuration]]=].
+:: The setter steps are:
+     1. Set [=this=]'s [=[[IncomingDatagramsExpirationDuration]]=] to the given value.
```
ricea, line 458: Should we check that the value is > 0 and not-NaN here?

index.bs
```diff
@@ -446,6 +451,30 @@ A {{DatagramDuplexStream}} object has the following internal slots.
...
+:: The getter steps are:
+     1. Return [=this=]'s [=[[OutgoingDatagramsExpirationDuration]]=].
+:: The setter steps are:
+     1. Set [=this=]'s [=[[OutgoingDatagramsExpirationDuration]]=] to the given value.
+
+: <dfn for="DatagramDuplexStream" attribute>incomingHighWaterMark</dfn>
+:: The getter steps are:
+     1. Return [=this=]'s [=[[IncomingDatagramsHighWaterMark]]=].
+:: The setter steps are:
+     1. Set [=this=]'s [=[[IncomingDatagramsHighWaterMark]]=] to the given value.
```
ricea, line 470: Should we check that the value is non-negative?

index.bs
jan-ivar: Meeting:
- Might be better to expose the implementation-defined default than return `null`?

yutakahirano: Please note that the implementation-default value can be chosen for each receiveDatagram call when  [[IncomingDatagramsExpirationDuration]]. So exposing a fixed value reduces the flexisibility. Ditto for the outgoing direction.

index.bs

index.bs
yutakahirano: Done. According to https://heycam.github.io/webidl/#idl-double and https://heycam.github.io/webidl/#idl-unrestricted-double we don't need to care about inf and NaN.

index.bs
```diff
@@ -446,6 +451,38 @@ A {{DatagramDuplexStream}} object has the following internal slots.
...
+     1. Let |value| be the given value.
+     1. If |value| > 0:
+       1. Set [=this=]'s [=[[OutgoingDatagramsExpirationDuration]]=] to |value|.
+
+: <dfn for="DatagramDuplexStream" attribute>incomingHighWaterMark</dfn>
+:: The getter steps are:
+     1. Return [=this=]'s [=[[IncomingDatagramsHighWaterMark]]=].
+:: The setter steps are:
+     1. Let |value| be the given value.
+     1. If |value| >= 0:
```
ricea, line 475: Maybe use `≥`? Or maybe not since I don't think we're using it elsewhere.

ricea (approved):
lgtm

index.bs
yutakahirano: Thanks, done. I found it's used in WebIDL.

yutakahirano:
@jan-ivar PTAL again.

I'm aware of https://github.com/w3c/webtransport/pull/267#discussion_r647519130 but I'd like to land this change soon. We can file another issue and discuss it there. If you don't like the possible breaking change, I can remove the getter part of incomingMaxAge and outgoingMaxAge.

yutakahirano:
Removing the "Discuss at next meeting" label as it is for the previous one.

index.bs
ricea: Maybe we want to explicitly allow Infinity as a MaxAge?

index.bs
```diff
@@ -446,6 +451,38 @@ A {{DatagramDuplexStream}} object has the following internal slots.
 :: The getter steps are:
      1. Return [=this=].`[[Writable]]`.
 
+: <dfn for="DatagramDuplexStream" attribute>incomingMaxAge</dfn>
+:: The getter steps are:
+     1. Return [=this=]'s [=[[IncomingDatagramsExpirationDuration]]=].
+:: The setter steps are:
+     1. Let |value| be the given value.
+     1. If |value| > 0:
```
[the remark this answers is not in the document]
yutakahirano: Thanks, done.

yutakahirano:
I'm going to land this change on Jun 16 (JST) if unless further comments are made.

Issue #268, opened 2021-05-19
Specify session termination


yutakahirano:
- Define the cleanup steps.
    - Abort/cancel streams.
    - Deal with the connection.
    - Resolve/reject the `closed` promise.
    - Transition the state.
 - A session can be terminated by client (i.e., WebTransport.close).
    - Terminate the session (refer to the protocol spec)
    - Run the cleanup steps.
  - A session can be terminated by the server or due to a network error
    - Run the cleanup steps.

yutakahirano:
> Deal with the connection.

This may not be needed - this can be done in the network layer, and covered by the protocol spec(s).

PR #269, opened 2021-05-19
Define "cleanup" procedure for WebTransport

yutakahirano:
Define the cleanup steps so that we can use it for
 - client-initiated termination,
 - server-initiated termination and
 - termination caused by a network error.

Add "Procedure" section in WebTransport, and move some procedure into it.

This is for #268.


yutakahirano:
@ricea @jan-ivar can you PTAL? I'm going to specify the server-initiated case and the network error case in subsequent PRs.

[[SendStreams]] and [[ReceiveStreams]] are defined in #266. Please assume they are sets of SendStreams and ReceiveStreams owned by the WebTransport, respectively.

index.bs
```diff
@@ -736,10 +712,13 @@ To <dfn>initialize WebTransport over HTTP</dfn>, given a {{WebTransport}} object
      1. Let |transport| be [=this=].
      1. If |transport|'s [=[[State]]=] is `"closed"` or `"failed"`,
         then abort these steps.
-     1. Set |transport|'s [=[[State]]=] to `"closed"`.
      1. Let |session| be |transport|'s [=[[Session]]=].
      1. [=session/Terminate=] |session| with |closeInfo|.{{WebTransportCloseInfo/errorCode}}.
 
+       Note: This also [=resets=] or [=sends STOP_SENDING=] [=WebTransport streams=] contained in
+       |transport|'s [=[[SendStreams]]=] and [=[[ReceiveStreams]]=].
```
jan-ivar, line 719: Does it? A note implies this is done elsewhere, yet I see nothing about this in [section 5](https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3/#section-5). Unless there's a different section we can reference, this note should probably morph into normative steps that take care of this here (in parallel) I think.

index.bs
```diff
@@ -736,10 +712,13 @@ To <dfn>initialize WebTransport over HTTP</dfn>, given a {{WebTransport}} object
      1. Let |transport| be [=this=].
      1. If |transport|'s [=[[State]]=] is `"closed"` or `"failed"`,
         then abort these steps.
-     1. Set |transport|'s [=[[State]]=] to `"closed"`.
      1. Let |session| be |transport|'s [=[[Session]]=].
      1. [=session/Terminate=] |session| with |closeInfo|.{{WebTransportCloseInfo/errorCode}}.
```
jan-ivar, line 716: Btw, does this sentence need to start with, "In parallel, "? I doubt this should block the main thread.

```suggestion
     1. In parallel, [=session/terminate=] |session| with |closeInfo|.{{WebTransportCloseInfo/errorCode}}.
```

index.bs
```diff
@@ -795,6 +800,56 @@ To <dfn>initialize WebTransport over HTTP</dfn>, given a {{WebTransport}} object
...
+   1. If {{RTCDtlsFingerprint/algorithm}} of |fingerprint| is equal to "sha-256",
+      and {{RTCDtlsFingerprint/value}} of |fingerprint| is equal to
+      |referenceFingerprint|, the certificate is valid.  Return true.
+1. Return false.
+
+</div>
+
+<div algorithm="cleanup a WebTransport">
+To <dfn for="WebTransport">cleanup</dfn> a {{WebTransport}} |transport| with |reason|,
+|error| and a boolean |abruptly|, run these steps:
```
jan-ivar, line 827: Nit: To me, this sounds like the "close" algorithm more than a "cleanup" algorithm (it sets the `"closed"` state and settles the `closed` promise). I'd rename it to that or — since it's only called from one place — maybe inline it? "Abruptly" makes more sense then too (an abrupt close makes more sense than an abrupt cleanup).

jan-ivar (changes requested):
Site-note: since github's diff viewer isn't great at catching this, I recommend moving code in separate editorial PRs to simplify review.

index.bs
yutakahirano: I don't have a strong opinion here. Accepting the suggestion.

index.bs
yutakahirano: "the endpoint MUST stop sending new datagrams and reset all of the streams associated with the session." It was not clear to me whether stream reset was included in the session termination action, so I added this note. @vasilvv do you have any comments?

index.bs
yutakahirano: As in the PR description I'm planning to use it in the server-initiated termination case and the network error case.

I'd like to avoid calling it "close" because it's used by the server initiated termination and the network error case. Calling it "close" will give an wrong impression that it's for WebTransport.close.

yutakahirano:
> Site-note: since github's diff viewer isn't great at catching this, I recommend moving code in separate editorial PRs to simplify review.

Reverted the moving part.

yutakahirano:
PTAL again.

yutakahirano:
@jan-ivar @ricea can you take a look again?

ricea:
lgtm once https://github.com/w3c/webtransport/pull/266 is landed.

yutakahirano:
Added the definition of [[SendStreams]] and [[ReceiveStreams]], in order to break the dependency to #266.

@ricea, can you take a look again? If you are fine I'll land this change.

ricea (approved):
lgtm

Issue #185, opened 2021-02-01
Call the streams algorithms


jan-ivar:
The streams spec has a [Using streams in other specifications](https://streams.spec.whatwg.org/#other-specs) section with algorithms we're meant to call. But we don't appear to be calling them. We should probably do that.

We seem light on algorithms to describe the fundamentals someone would need to get right to implement this web compatibly: threading model, timing, type checking, and error handling, which I'm unable to glean from current prose.

Coming from WebRTC, I'd assume networking requires a similar model to that: data I/O largely off main thread, doing a lot of "in parallel" and queued tasks to communicate with (main or worker) JS. https://github.com/w3c/webtransport/issues/112

This came to mind while I was reviewing unrelated streams patches over in https://github.com/w3c/webrtc-insertable-streams/pull/51#discussion_r567910473

Similar to that patch, once we start calling the streams algorithms directly, this should unfurl how to answer these questions.

jan-ivar:
Work seems to be happening in #266 for this. Closing.

yutakahirano:
Reopening as we have more things to do here.

PR #283, opened 2021-06-16
Specify SendStream algorithms

### LINKED ISSUE #259: Should writable.close wait for all the data to be sent to the peer?
According to [QUIC#section-3.1](https://datatracker.ietf.org/doc/html/draft-ietf-quic-transport#section-3.1), after the sender sends a STREAM frame with FIN, the stream enters the "Data Sent" state and it will only retransmits stream data as necessary.

Whether the promise returned by `writer.close()` should wait for [re]transmission of all the data needs to be specified. We have a few options I think.

1. The promise is fulfilled as soon as the STREAM+FIN frame is queued in an internal queue (or, we can say that the promise is fulfilled immediately).
2. The promise is fulfilled as soon as the stream enters the "Data Sent" state (i.e., the STREAM+FIN frame is sent to the network once).
3. The promise is fulfilled as soon as the stream enters the "Data Recvd" state.
4. We can think of a two-way closing handshake, similar to WebSocket. The promise is fulfilled as soon as the handshake finishes.

yutakahirano:
- Specify write, close and abort algorithms for SendStream.
 - Specify the behavior for STOP_SENDING signal.

This is for #185.
Fixes #259.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/pull/283.html" title="Last updated on Jun 23, 2021, 3:51 AM UTC (4a95cd3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/283/bf4d9cd...4a95cd3.html" title="Last updated on Jun 23, 2021, 3:51 AM UTC (4a95cd3)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -248,6 +248,37 @@ or <dfn for=stream>bidirectional</dfn>.
  </tbody>
 </table>
 
+[=WebTransport stream=] has the following signals:
+
+<table class="data" dfn-for="stream-event">
+ <thead>
+  <tr>
+   <th>event
+   <th>definition
+   <th>incoming
+   <th>outgoing
+   <th>bidirectional
+  </tr>
+ </thead>
+ <tbody>
+  <tr>
+   <td><dfn>STOP_SENDING</dfn>
+   <td>[[!QUIC]]
+   [section 3.5](https://datatracker.ietf.org/doc/html/draft-ietf-quic-transport#section-3.5)
+   <td>No
+   <td>Yes
+   <td>Yes
+  </tr>
+  <tr>
+   <td><dfn>reset</dfn>
+   <td>[[!QUIC]]
+   [section 19.4](https://datatracker.ietf.org/doc/html/draft-ietf-quic-transport#section-19.4)
+   <td>Yes
+   <td>No
+   <td>Yes
+  </tr>
+ </tbody>
+</table>
 
 # `DatagramDuplexStream` Interface #  {#duplex-stream}
 
@@ -1068,6 +1099,11 @@ A {{SendStream}} has the following internal slots.
    <td><dfn>\[[InternalStream]]</dfn>
    <td class="non-normative">An [=outgoing=] or [=bidirectional=] [=WebTransport stream=].
   </tr>
+  <tr>
+   <td><dfn>\[[Transport]]</dfn>
+   <td class="non-normative">A {{WebTransport}} which owns this {{SendStream}}.
+  </tr>
+ <tbody>
 </table>
 
 ## Procedures ##  {#send-stream-procedures}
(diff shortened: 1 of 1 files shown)


yutakahirano:
@vasilvv @jan-ivar @ricea can you take a look?

index.bs
```diff
@@ -1062,6 +1093,12 @@ A {{SendStream}} has the following internal slots.
    <td><dfn>\[[InternalStream]]</dfn>
    <td class="non-normative">An [=outgoing=] or [=bidirectional=] [=WebTransport stream=].
   </tr>
+  <tr>
+   <td><dfn>\[[Transport]]</dfn>
+   <td class="non-normative">A {{WebTransport}} which owns this {{SendStream}}.
+  </tr>
+  <tbody>
```
ricea, line 1100: `<tbody>` placement seems odd.

index.bs
```diff
@@ -1075,12 +1112,68 @@ To <dfn export for="SendStream" lt="create|creating">create</dfn> a
 1. Let |stream| be a [=new=] {{SendStream}}, with:
     : [=[[InternalStream]]=]
     :: |internalStream|
-1. [=WritableStream/Set up=] |stream| with TBD.
-1. [=set/Add=] |stream| to |transport|'s [=[[SendStreams]]=].
+    : [=[[Transport]]=]
+    :: |transport|
+1. Let |writeAlgorithm| be an action that [=writes=] |chunk| with |stream|, given |chunk|.
+1. Let |closeAlgorithm| be an action that [=closes=] |stream|.
+1. Let |abortAlgorithm| be an action that [=aborts=] |stream|.
```
ricea, line 1119: Maybe you need to specify `reason` here?

index.bs
```diff
@@ -1075,12 +1112,68 @@ To <dfn export for="SendStream" lt="create|creating">create</dfn> a
...
+1. [=set/Append=] |stream| to |transport|'s [=[[SendStreams]]=].
 1. Return |stream|.
 
 </div>
 
+<div algorithm>
+To <dfn for="SendStream">write</dfn> |chunk| with a {{SendStream}} |stream|, run these steps:
+
+1. Let |transport| be |stream|'s [=[[Transport]]=].
+1. If |chunk| is not a {{Uint8Array}}, return [=a promise rejected with=] a {{TypeError}}.
```
ricea, line 1132: In Blink we allow any `BufferSource` type here. Is there a reason to be more restrictive?

index.bs
```diff
@@ -1075,12 +1112,68 @@ To <dfn export for="SendStream" lt="create|creating">create</dfn> a
...
+   complete.
+1. [=Queue a network task=] with |transport| to [=resolve=] |promise| with undefined.
+
+Note: The user-agent MAY have a buffer to improve the transfer performance. Such a buffer
+SHOULD have a fixed size, to carry the backpressure information to the user of {{SendStream}}. This
+also means the [=fulfillment=] of the promise returned from this algorithm (or,
+{{WritableStreamDefaultWriter.write}}) does **NOT** necessarily mean that the chunk is acked by
+the server [[!QUIC]]. It may just mean that the chunk is just appended to the buffer. To
+make sure that the chunk arrives at the server, close the {{SendStream}} and wait for the
+result promise to be fulfilled, or design an application-level protocol.
```
ricea, line 1146: Maybe say "use" rather than "design", since someone may have already designed one for you 🙂

ricea (commented):
We need to use https://github.com/whatwg/streams/pull/1132 once it is landed.

index.bs
yutakahirano: I don't have a strong reason. I filed #284. Let's discuss there.

index.bs

index.bs

index.bs
yutakahirano: Fixed.

index.bs
```diff
@@ -248,6 +248,37 @@ or <dfn for=stream>bidirectional</dfn>.
...
+[=WebTransport stream=] has the following events:
+
+<table class="data" dfn-for="stream-event">
+ <thead>
+  <tr>
+   <th>event
+   <th>definition
+   <th>incoming
+   <th>outgoing
+   <th>bidirectional
```
vasilvv, line 260: I'm confused by the terms used here; a bidirectional stream can be either incoming or outgoing.  Did you mean to use send/receive stream?

index.bs
```diff
@@ -1075,12 +1111,68 @@ To <dfn export for="SendStream" lt="create|creating">create</dfn> a
...
+1. Let |transport| be |stream|'s [=[[Transport]]=].
+1. If |chunk| is not a {{Uint8Array}}, return [=a promise rejected with=] a {{TypeError}}.
+1. Let |promise| be a new promise.
+1. Let |bytes| be a copy of the [=byte sequence=] which |chunk| represents.
+1. Return |promise| and run the remaining steps [=in parallel=].
+1. [=stream/Send=] |bytes| on |stream|'s [=[[InternalStream]]=] and wait for the operation to
+   complete.
+1. [=Queue a network task=] with |transport| to [=resolve=] |promise| with undefined.
+
+Note: The user-agent MAY have a buffer to improve the transfer performance. Such a buffer
```
vasilvv, line 1139: The user agent will always have the buffer, since that's what it uses to retransmit data in case of packet loss.

index.bs
```diff
@@ -1075,12 +1111,68 @@ To <dfn export for="SendStream" lt="create|creating">create</dfn> a
...
+SHOULD have a fixed size, to carry the backpressure information to the user of {{SendStream}}. This
+also means the [=fulfillment=] of the promise returned from this algorithm (or,
+{{WritableStreamDefaultWriter.write}}) does **NOT** necessarily mean that the chunk is acked by
+the server [[!QUIC]]. It may just mean that the chunk is just appended to the buffer. To
+make sure that the chunk arrives at the server, close the {{SendStream}} and wait for the
+result promise to be fulfilled, or use an application-level protocol.
```
vasilvv, line 1145: I don't think we want to encourage relying on the `close()` promise, because even if the chunk arrives at the server, that does not necessarily mean that the server application has processed the data in question.

index.bs
yutakahirano: In theory the user agent can block the sending operation until it sends _bytes_ and gets ACKs for it. In that case the fulfillment of the result promise means the data arrived at the server. This note emphasizes that it is NOT guaranteed.

index.bs
yutakahirano: Here is my mental model.

 - HTTP/3 stream is specified in HTTP/3.
 - WebTransport uses several HTTP/3 streams. They are specified in [WEB-TRANSPORT-HTTP3].
   - The CONNECT stream (one for a session)
   - Incoming streams (from the client PoV, ditto below)
   - Outgoing streams
   - Bidirectional streams
  - SendStreams and ReceiveStreams are JS objects. This specification specifies the behavior of them by using capabilities and events for [=WebTransport streams=].

This section consists of adapters of concepts defined in [WEB-TRANSPORT-HTTP3].

Does this make sense?

index.bs
yutakahirano: Fixed.

index.bs
vasilvv: I find it weird using "Incoming streams" to specifically mean "Incoming unidirectional streams", when we have both `incomingBidirectionalStreams` and `incomingUnidirectionalStreams` in the API.

index.bs
vasilvv: Ah, I missed that that's a possible implementation of write().  Fair point.

index.bs
yutakahirano: I can replace ["incoming"](https://w3c.github.io/webtransport/#stream-incoming) with "incoming unidirectional" and ["outgoing"](https://w3c.github.io/webtransport/#stream-outgoing) with "outgoing unidirectional", in a separate change.

index.bs
```diff
@@ -248,6 +248,37 @@ or <dfn for=stream>bidirectional</dfn>.
  </tbody>
 </table>
 
+[=WebTransport stream=] has the following events:
```
jan-ivar, line 251: "events" is the wrong word here, as it has existing meaning in JS specs. How about "signal" (which you use below)?

```suggestion
[=WebTransport stream=] has the following signals:
```

index.bs
```diff
@@ -1075,12 +1111,67 @@ To <dfn export for="SendStream" lt="create|creating">create</dfn> a
...
+1. [=WritableStream/Set up=] |stream| with [=WritableStream/set up/writeAlgorithm=] set to
+   |writeAlgorithm|, [=WritableStream/set up/closeAlgorithm=] set to |closeAlgorithm|,
+   [=WritableStream/set up/abortAlgorithm=] set to |abortAlgorithm|.
+1. [=set/Append=] |stream| to |transport|'s [=[[SendStreams]]=].
 1. Return |stream|.
 
 </div>
 
+<div algorithm>
+To <dfn for="SendStream">write</dfn> |chunk| with a {{SendStream}} |stream|, run these steps:
```
jan-ivar, line 1128: Seems formal. Maybe "to"?

```suggestion
To <dfn for="SendStream">write</dfn> |chunk| to a {{SendStream}} |stream|, run these steps:
```

jan-ivar (approved):

vasilvv (approved):

index.bs
yutakahirano: https://github.com/w3c/webtransport/pull/289

PR #295, opened 2021-06-25
Have SendStream's write/close abortable

### LINKED ISSUE #259: Should writable.close wait for all the data to be sent to the peer?

### LINKED ISSUE #260: Do we need writable.reset() for "unreliable streams"?
At #242 and #248 we talked about whether we need writable.reset() (or abortWriting) to abort the stream. The conclusion was we don't need it and just using writable.abort() (or writer.abort()) is fine, assuming whatwg/streams/issues/1015 will be resolved.

But @ricea pointed out that it would be useful to make the stream unreliable.

```
const writer = writable.getWriter();
writer.write(data);
await writer.close();
// To suppress retransmission of `data`.
writable.reset();
```

This use case is not covered by writable.abort() because it is no-op when the stream is closed.

See also: #259

yutakahirano:
Have SendStream's write and close abortable with
WritableStreamDefaultController.signal.

Remove SendStream.reset as it's now covered by abort().

Fixes #260.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/pull/295.html" title="Last updated on Jun 25, 2021, 3:24 PM UTC (95c2a3b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/295/4571dcf...95c2a3b.html" title="Last updated on Jun 25, 2021, 3:24 PM UTC (95c2a3b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1114,7 +1114,6 @@ that can be written to, to transmit data to the server.
 <pre class="idl">
 [Exposed=(Window,Worker), SecureContext]
 interface SendStream : WritableStream /* of Uint8Array */ {
-  undefined reset(optional StreamAbortInfo abortInfo = {});
 };
 </pre>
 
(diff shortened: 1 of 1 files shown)


ricea (approved):
lgtm

PR #297, opened 2021-06-28
Set up ReceiveStream with proper algorithms

### LINKED ISSUE #185: Call the streams algorithms

yutakahirano:
Define pullAlgorithm and cancelAlgorithm, and set up ReceiveStream
with them.

This relies on https://github.com/whatwg/streams/pull/1130.

Fixes #185.

***

### :boom: Error: 500 Internal Server Error :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Jul 29, 2021, 2:42 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [CSS Spec Preprocessor](https://api.csswg.org/bikeshed/) - CSS Spec Preprocessor is the web service used to build Bikeshed specs.

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2Fw3c%2Fwebtransport%2Fe23c9c325fb45ba758701e943a496e25da6e8867%2Findex.bs&md-warning=not%20ready)

```
<html><body><p><strong>WARNING: </strong>mysqli_connect(): (HY000/2002): Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (111)</p></body></html>
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/webtransport%23297.)._
</details>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1354,7 +1354,7 @@ To <dfn for="ReceiveStream">pull bytes</dfn> from a {{ReceiveStream}} |stream|, 
   1. If |hasReceivedFIN| is true:
     1. [=set/Remove=] |stream| from |transport|'s [=[[ReceiveStreams]]=].
     1. [=ReadableStream/Close=] |stream|.
-  1. [=Resolve=] promise with undefined.
+  1. [=Resolve=] |promise| with undefined.
 
 </div>
 
--- a/index.bs
+++ b/index.bs
@@ -1354,7 +1354,7 @@ To <dfn for="ReceiveStream">pull bytes</dfn> from a {{ReceiveStream}} |stream|, 
 1. [=Queue a network task=] with |transport| to run these steps:
 
    Note: If the buffer described above is available in the [=event loop=] where this procedure is
-   running, this block MAY run immediately.
+   running, the following steps MAY run immediately.
 
   1. If |read| > 0, then perform [=ReadableByteStreamControllerRespond=] with |controller| and
      |read|.
--- a/index.bs
+++ b/index.bs
@@ -1364,7 +1364,7 @@ To <dfn for="ReceiveStream">pull bytes</dfn> from a {{ReceiveStream}} |stream|, 
    running, the following steps may run immediately.
 
   1. Set |view| to a [=new=] {{Uint8Array}} with |arrayBuffer|, |offset| and |read|.
-  1. If |read| > 0, then [=ReadableStream/enqueue=] into |stream| with |view|.
+  1. If |read| > 0, then [=ReadableStream/enqueue=] |view| into |stream|.
   1. If |hasReceivedFIN| is true:
     1. [=set/Remove=] |stream| from |transport|'s [=[[ReceiveStreams]]=].
     1. [=ReadableStream/Close=] |stream|.
--- a/index.bs
+++ b/index.bs
@@ -1333,7 +1333,7 @@ To <dfn for="ReceiveStream">pull bytes</dfn> from a {{ReceiveStream}} |stream|, 
 1. Let |internalStream| be |stream|'s [=ReceiveStream/[[InternalStream]]=].
 1. Let |promise| be a new promise.
 1. Let |buffer|, |offset|, and |maxBytes| be null.
-1. If |stream|'s [=ReadableStream/current BYOB request view=] for |stream| is not null, then:
+1. If |stream|'s [=ReadableStream/current BYOB request view=] for |stream| is not null:
   1. Set |offset| to |stream|'s [=ReadableStream/current BYOB request view=].\[[ByteOffset]].
   1. Set |maxBytes| to |stream|'s [=ReadableStream/current BYOB request view=]'s
      [=BufferSource/byte length=].
@@ -1364,7 +1364,7 @@ To <dfn for="ReceiveStream">pull bytes</dfn> from a {{ReceiveStream}} |stream|, 
    Note: If the buffer described above is available in the [=event loop=] where this procedure is
    running, the following steps may run immediately.
 
-  1. If |read| > 0, then:
+  1. If |read| > 0:
     1. Set |view| to a [=new=] {{Uint8Array}} with |buffer|, |offset| and |read|.
     1. [=ReadableStream/Enqueue=] |view| into |stream|.
   1. If |hasReceivedFIN| is true:
--- a/index.bs
+++ b/index.bs
@@ -1361,10 +1361,10 @@ To <dfn for="ReceiveStream">pull bytes</dfn> from a {{ReceiveStream}} |stream|, 
 1. [=Queue a network task=] with |transport| to run these steps:
 
    Note: If the buffer described above is available in the [=event loop=] where this procedure is
-   running, the following steps MAY run immediately.
+   running, the following steps may run immediately.
 
   1. Set |view| to a [=new=] {{Uint8Array}} with |arrayBuffer|, |offset| and |read|.
-  1. If |read| > 0, then [=ReadableStream/enqueue bytes=] into |stream| with |view|.
+  1. If |read| > 0, then [=ReadableStream/enqueue=] into |stream| with |view|.
   1. If |hasReceivedFIN| is true:
     1. [=set/Remove=] |stream| from |transport|'s [=[[ReceiveStreams]]=].
     1. [=ReadableStream/Close=] |stream|.
--- a/index.bs
+++ b/index.bs
@@ -1315,10 +1315,11 @@ To <dfn export for="ReceiveStream" lt="create|creating">create</dfn> a
     :: |transport|
 1. Let |pullAlgorithm| be an action that [=pulls bytes=] from |stream|.
 1. Let |cancelAlgorithm| be an action that [=ReceiveStream/cancels=] |stream|.
-1. [=ReadableByteStream/Set up=] |stream| with
-   [=ReadableByteStream/set up/pullAlgorithm=] set to |pullAlgorithm|,
-   [=ReadableByteStream/set up/cancelAlgorithm=] set to |cancelAlgorithm|, and
-   [=ReadableByteStream/set up/highWaterMark=] set to 0.
+1. [=ReadableStream/Set up with byte reading support=] |stream| with
+   [=ReadableStream/set up with byte reading support/pullAlgorithm=] set to |pullAlgorithm| and
+   [=ReadableStream/set up with byte reading support/cancelAlgorithm=] set to |cancelAlgorithm|.
+1. Set |stream|'s [=ReadableByteStream/[[controller]]=]'s
+   [=ReadableStreamController/autoAllocateChunkSize=] to an [=implementation-defined=] integer.
 1. [=set/Add=] |stream| to |transport|'s [=[[ReceiveStreams]]=].
 1. Return |stream|.
 
@@ -1335,6 +1336,9 @@ To <dfn for="ReceiveStream">pull bytes</dfn> from a {{ReceiveStream}} |stream|, 
 1. Let |promise| be a new promise.
 1. Let |request| be the result of running [=ReadableByteStreamControllerGetBYOBRequest=] with
    |controller|.
+
+  Note: |request| is not null, because of [=ReadableStreamController/autoAllocateChunkSize=] set in
+  the [=ReceiveStream/create=] procedure.
 1. Let |view| be |request|'s [=ReadableStreamBYOBRequest/view=].
 1. Assert: |view| is not null.
 1. Let |bytes| be the byte sequence |view| represents.
--- a/index.bs
+++ b/index.bs
@@ -1343,7 +1343,7 @@ To <dfn for="ReceiveStream">pull bytes</dfn> from a {{ReceiveStream}} |stream|, 
   1. Set |offset| to 0.
   1. Set |maxBytes| to an [=implementation-defined=] size.
   1. Set |buffer| be a [=new=] {{ArrayBuffer}} with |maxBytes| size. If allocating the
-     {{Uint8Array}} fails, return a promise [=rejected with=] a {{RangeError}}.
+     {{ArrayBuffer}} fails, return a promise [=rejected with=] a {{RangeError}}.
 1. Return |promise| and run the remaining steps [=in parallel=].
 1. [=ArrayBuffer/Write=] the bytes that area [=stream/receive|read=] from |internalStream| into
    |buffer| with offset |offset|, up to |maxBytes| bytes. Wait until either at least one byte is
--- a/index.bs
+++ b/index.bs
@@ -1329,8 +1329,6 @@ To <dfn export for="ReceiveStream" lt="create|creating">create</dfn> a
 
 To <dfn for="ReceiveStream">pull bytes</dfn> from a {{ReceiveStream}} |stream|, run these steps.
 
-1. Let |controller| be |stream|'s [=ReadableStream/[[controller]]=].
-1. Assert: |controller| is a {{ReadableByteStreamController}}.
 1. Let |transport| be |stream|'s [=ReceiveStream/[[Transport]]=].
 1. Let |internalStream| be |stream|'s [=ReceiveStream/[[InternalStream]]=].
 1. Let |promise| be a new promise.
@@ -1340,13 +1338,11 @@ To <dfn for="ReceiveStream">pull bytes</dfn> from a {{ReceiveStream}} |stream|, 
    {{RangeError}}.
 1. Let |bytes| be the byte sequence |view| represents.
 1. Assert: |bytes|'s length equals to |view|.\[[ByteLength]].
-1. Let |arrayBuffer| be |view|.\[[ViewedArrayBuffer]].
-1. Let |offset| be |view|.\[[ByteOffset]].
 1. Return |promise| and run the remaining steps [=in parallel=].
-1. [=stream/receive|Read bytes=] into |bytes|, from |internalStream|. Wait until either at least one
-   byte is read or FIN is received. Let |read| be the number of read bytes, and let |hasReceivedFIN|
-   be whether FIN was accompanied.
-
+1. [=ArrayBufferView/Write=] the bytes that area [=stream/receive|read=] from |internalStream| into
+   |view|, up to |bytes|'s length bytes. Wait until either at least one byte is read or FIN is
+   received. Let |read| be the number of read bytes, and let |hasReceivedFIN| be whether FIN was
+   accompanied.
 
    Note: The user-agent MAY have a buffer to improve the transfer performance. Such a buffer
    SHOULD have a fixed size, to carry the backpressure information to the server.
@@ -1363,8 +1359,13 @@ To <dfn for="ReceiveStream">pull bytes</dfn> from a {{ReceiveStream}} |stream|, 
    Note: If the buffer described above is available in the [=event loop=] where this procedure is
    running, the following steps may run immediately.
 
-  1. Set |view| to a [=new=] {{Uint8Array}} with |arrayBuffer|, |offset| and |read|.
-  1. If |read| > 0, then [=ReadableStream/enqueue=] |view| into |stream|.
+  1. If [=ReadableStream/current BYOB request view=] for |stream| is null:
+    1. Let |arrayBuffer| be |view|.\[[ViewedArrayBuffer]].
+    1. Let |offset| be |view|.\[[ByteOffset]].
+    1. Set |view| to a [=new=] {{Uint8Array}} with |arrayBuffer|, |offset| and |read|.
+    1. If |read| > 0, then [=ReadableStream/enqueue=] |view| into |stream|.
+  1. Otherwise:
+    1. If |read| > 0, then [=ReadableStream/enqueue=] |view| into |stream| with |read|.
   1. If |hasReceivedFIN| is true:
     1. [=set/Remove=] |stream| from |transport|'s [=[[ReceiveStreams]]=].
     1. [=ReadableStream/Close=] |stream|.
--- a/index.bs
+++ b/index.bs
@@ -1318,8 +1318,6 @@ To <dfn export for="ReceiveStream" lt="create|creating">create</dfn> a
 1. [=ReadableStream/Set up with byte reading support=] |stream| with
    [=ReadableStream/set up with byte reading support/pullAlgorithm=] set to |pullAlgorithm| and
    [=ReadableStream/set up with byte reading support/cancelAlgorithm=] set to |cancelAlgorithm|.
-1. Set |stream|'s [=ReadableByteStream/[[controller]]=]'s
-   [=ReadableStreamController/autoAllocateChunkSize=] to an [=implementation-defined=] integer.
 1. [=set/Add=] |stream| to |transport|'s [=[[ReceiveStreams]]=].
 1. Return |stream|.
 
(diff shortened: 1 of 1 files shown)


yutakahirano:
@ricea, can you take a look?
@domenic for the interaction with whatwg/streams#1130: Am I using it correctly?

index.bs
```diff
@@ -1294,44 +1294,101 @@ A {{ReceiveStream}} has the following internal slots.
...
-1. [=ReadableStream/Set up=] |stream| with TBD.
+    : [=ReceiveStream/[[Transport]]=]
+    :: |transport|
+1. Let |pullAlgorithm| be an action that [=pulls bytes=] from |stream|.
+1. Let |cancelAlgorithm| be an action that [=ReceiveStream/cancels=] |stream|.
+1. [=ReadableStream/Set up with byte reading support=] |stream| with
+   [=ReadableStream/set up with byte reading support/pullAlgorithm=] set to |pullAlgorithm| and
+   [=ReadableStream/set up with byte reading support/cancelAlgorithm=] set to |cancelAlgorithm|.
+1. Set |stream|'s [=ReadableByteStream/[[controller]]=]'s
+   [=ReadableStreamController/autoAllocateChunkSize=] to an [=implementation-defined=] integer.
```
domenic, line 1322: Our current plan is for spec-created streams to not use autoAllocateChunkSize, so this line should not be needed.

index.bs
```diff
@@ -1294,44 +1294,101 @@ A {{ReceiveStream}} has the following internal slots.
...
-     1. Let |reason| be the first argument, if it is numeric, or 0.
-     1. If |stream| has received a RESET_STREAM frame, abort these steps.
-     1. [=In parallel=], [=send STOP_SENDING=] over |stream|'s
-        [=WebTransport stream=] with |reason| as the Application Protocol Error
-         Code.
+<div algorithm>
+
+To <dfn for="ReceiveStream">pull bytes</dfn> from a {{ReceiveStream}} |stream|, run these steps.
+
+1. Let |controller| be |stream|'s [=ReadableStream/[[controller]]=].
```
domenic, line 1332: This algorithm uses too many internals, which should not be necessary. Instead you should use https://whatpr.org/streams/1130.html#readablestream-current-requested-number-of-bytes and https://whatpr.org/streams/1130.html#readablestream-enqueue-bytes as appropriate.

If those don't work for you, then that would be great feedback for the streams spec.

domenic (changes requested):

ricea (commented):
I think we should wait for whatwg/streams#1130.

index.bs
yutakahirano: Thank you! 

That will lead to additional bytes copy. What do you think? We could say that implementers can always perform an optimization, but...

domenic: Right, the "implementers can always perform an optimization" is what I am currently proposing in the streams spec. There it says

> Implementations are encouraged to, when possible, do this directly from whatever underlying I/O system is producing _bytes_. In this sense the existence of an intermediate _bytes_ variable might end up as spec fiction, if the underlying I/O system can write directly into the backing data block.

This seems worthwhile so that specs can be written more simply. Note that autoAllocateChunkSize doesn't work for all cases; see some of the discussion in https://github.com/whatwg/streams/issues/1121#issuecomment-814473948 .

I'm very open to feedback though if you think this is too clever and we should be more explicit in the client specs like WebTransport, with less magic in the Streams spec.

yutakahirano: Thanks, I can live with that.

Another possible approach is, to have "enqueue bytes" take an ArrayBufferView _view_ instead of a byte sequence, and use _byobRequest_ when _view_ is contained in _byobRequest_'s \[[view]].

It will look like:

4. If _byobRequest_ is non-null, _byobRequest_.[[view]].[[ViewedArrayBuffer]] is _view_.[[ViewedArrayBuffer]], 
       _byobRequest_.[[view]].[[ByteOffset]] is _view_.[[ByteOffset]] and 
       _byobRequest_.[[view]].[[ByteLength]] ≤ _view_.[[ByteLength]]:
   1. Perform ! ReadableByteStreamControllerRespond(stream.[[controller]], _view_.[[ByteLength]]). 
5. Otherwise:
   1. Perform ! ReadableByteStreamControllerEnqueue(stream.[[controller]], _view_).

And also expose the byobRequest's \[[view]] as "current requested view" instead of "current requested number of bytes".

I prefer this approach because it matches with my mental model better: when the byobRequest is provided, I can borrow the ArrayBufferView, update the contents and return it or its subarray. Otherwise I can create a new ArrayBufferView, update its contents and enqueue it.

index.bs
```diff
@@ -1294,44 +1294,101 @@ A {{ReceiveStream}} has the following internal slots.
...
+
+  Note: This operation may return before filling up all of |bytes|.
+
+1. [=Queue a network task=] with |transport| to run these steps:
+  1. If |read| > 0, then perform [=ReadableByteStreamControllerRespond=] with |controller| and
+     |read|.
+  1. If |hasReceivedFIN| is true:
+    1. [=set/Remove=] |stream| from |transport|'s [=[[ReceiveStreams]]=].
+    1. [=ReadableStream/Close=] |stream|.
+  1. [=Resolve=] |promise| with undefined.
```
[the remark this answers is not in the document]
yutakahirano: Added a step and a note. We need to reject the promise with a proper error object, so I'd like to leave it as is.

index.bs
```diff
@@ -1294,44 +1294,101 @@ A {{ReceiveStream}} has the following internal slots.
...
+</div>
+
+<div algorithm>
+
+To <dfn for="ReceiveStream">cancel</dfn> a {{ReceiveStream}} |stream|, run these steps.
+
+1. Let |transport| be |stream|'s [=ReceiveStream/[[Transport]]=].
+1. Let |internalStream| be |stream|'s [=ReceiveStream/[[InternalStream]]=].
+1. Let |promise| be a new promise.
+1. Return |promise| and run the remaining steps [=in parallel=].
```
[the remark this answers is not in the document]
yutakahirano: Added a note - Is it clear?

index.bs
```diff
@@ -1294,44 +1294,101 @@ A {{ReceiveStream}} has the following internal slots.
...
+1. Let |request| be the result of running [=ReadableByteStreamControllerGetBYOBRequest=] with
+   |controller|.
+
+  Note: |request| is not null, because of [=ReadableStreamController/autoAllocateChunkSize=] set in
+  the [=ReceiveStream/create=] procedure.
+1. Let |view| be |request|'s [=ReadableStreamBYOBRequest/view=].
+1. Assert: |view| is not null.
+1. Let |bytes| be the byte sequence |view| represents.
+1. Return |promise| and run the remaining steps [=in parallel=].
+1. [=stream/receive|Read bytes=] into |bytes|, from |internalStream|. Let |read| be the number of
```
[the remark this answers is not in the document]
yutakahirano: Embedded it to this step.

index.bs
yutakahirano: done.

index.bs
```diff
@@ -1294,44 +1294,101 @@ A {{ReceiveStream}} has the following internal slots.
...
+
+To <dfn for="ReceiveStream">cancel</dfn> a {{ReceiveStream}} |stream|, run these steps.
+
+1. Let |transport| be |stream|'s [=ReceiveStream/[[Transport]]=].
+1. Let |internalStream| be |stream|'s [=ReceiveStream/[[InternalStream]]=].
+1. Let |promise| be a new promise.
+1. Return |promise| and run the remaining steps [=in parallel=].
+1. [=Send STOP_SENDING=] with |internalStream|.
+1. [=Queue a network task=] with |transport| to run these steps:
+  1. If |transport|'s [=[[State]]=] is `"closed"` or `"failed"`, abort these steps.
```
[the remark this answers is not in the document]

index.bs
```diff
@@ -1269,7 +1269,9 @@ The dictionary SHALL have the following fields:
 # Interface `ReceiveStream` #  {#receive-stream}
 
 A <dfn interface>ReceiveStream</dfn> is a {{ReadableStream}} of {{Uint8Array}}
-that can be read from, to consume data received from the server.
+that can be read from, to consume data received from the server. {{ReceiveStream}} is a
```
yutakahirano, line 1272: @domenic What do you think about this explanation? If this looks good, would you export the terms in https://github.com/whatwg/streams/pull/1130? Or is it better for me to create a separate PR?

index.bs
domenic: That makes sense to me. I will work on updating the spec in that direction.

index.bs
domenic: This explanation looks great. I did not realize those terms were not exported; I will do so in the PR.

domenic:
The Streams spec change has been merged and within 24 hours should show up in cross-linking databases. One thing that changed is we used "enqueue" instead of "enqueue bytes".

yutakahirano:
Thank you! I updated the PR. PTAL again.

index.bs
```diff
@@ -1294,44 +1296,112 @@ A {{ReceiveStream}} has the following internal slots.
...
+1. Let |view| be [=ReadableStream/current BYOB request view=] for |stream|.
+1. If |view| is null, set |view| to a [=new=] {{Uint8Array}} with an [=implementation-defined=]
+   size. If allocating the {{Uint8Array}} fails, return a promise [=rejected with=] a
+   {{RangeError}}.
+1. Let |bytes| be the byte sequence |view| represents.
+1. Assert: |bytes|'s length equals to |view|.\[[ByteLength]].
+1. Let |arrayBuffer| be |view|.\[[ViewedArrayBuffer]].
+1. Let |offset| be |view|.\[[ByteOffset]].
+1. Return |promise| and run the remaining steps [=in parallel=].
+1. [=stream/receive|Read bytes=] into |bytes|, from |internalStream|. Wait until either at least one
```
domenic, line 1346: I would suggest phrasing this in terms of https://heycam.github.io/webidl/#arraybufferview-write , i.e. something like

```html
[=ArrayBufferView/write=] the bytes that area [=stream/receive|read=] from |internalStream| into |view|.
```

You may also want to require somehow that the implementation not receive more bytes than |view|'s [byte length](https://heycam.github.io/webidl/#buffersource-byte-length) ? It's kind of implied, but not obvious.

index.bs
```diff
@@ -1294,44 +1296,112 @@ A {{ReceiveStream}} has the following internal slots.
...
+
+  Note: We don't reject |promise| here because we handle network errors elsewhere, and those logic
+  error |stream| and reject the result of this read operation.
+
+1. [=Queue a network task=] with |transport| to run these steps:
+
+   Note: If the buffer described above is available in the [=event loop=] where this procedure is
+   running, the following steps may run immediately.
+
+  1. Set |view| to a [=new=] {{Uint8Array}} with |arrayBuffer|, |offset| and |read|.
```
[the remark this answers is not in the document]
domenic: https://github.com/whatwg/streams/pull/1146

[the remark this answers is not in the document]
yutakahirano: Hmm, the current logic uses _chunk_.[[ByteLength]] which is aligned with my understanding.

I'm worried about this assertion in the new logic:

>    1. Assert: if |bytesWritten| was given, it's equal to |chunk|.\[[ByteLength]]. (Anything else indicates a logic error in the calling specification.)

This requires WebTransport to predict the allocation size in front (which is possible for some but not all platforms), or recreate the view.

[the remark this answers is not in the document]
domenic: > Hmm, the current logic uses chunk.[[ByteLength]] which is aligned with my understanding.

The idea is that your specification should not have to trim down the buffer, which incurs a copy. In other words the intent is to replace these two steps with

1. Enqueue |view| into |stream| with |read|.

> I'm worried about this assertion in the new logic:

I think recreating the view is necessary. The stream consumer should never receive a view which is padded with 0s. It's OK for the *buffer* to be padded with 0s, but not the view. The view's byte length should tell the consumer how many bytes were read.

Alternately, you could avoid allocating the view until |read| is known, and instead only allocate the underlying ArrayBuffer beforehand.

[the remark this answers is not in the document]
yutakahirano: > The idea is that your specification should not have to trim down the buffer, which incurs a copy.

Creating an ArrayBufferView from an ArrayBuffer doesn't lead to copying.

In the following code,

```
const ab = new ArrayBuffer(8);
let view1 = new Uint8Array(ab, 4, 3);
let view2 = new Uint8Array(ab, 4, 2);
```

_view1_ and _view2_ shares the same buffer (_ab_), and there is no copying when creating Uint8Arrays.
Am I missing something?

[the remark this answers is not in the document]
domenic: You are right; I misspoke when I mentioned the copy. However it is redundant as ReadableByteStreamControllerRespond will eventually construct a new view itself. (In [ReadableByteStreamControllerConvertPullIntoDescriptor](https://streams.spec.whatwg.org/#readable-byte-stream-controller-convert-pull-into-descriptor).) You just need to pass the |read| so we know what size view to construct.

index.bs
```diff
@@ -1294,44 +1296,112 @@ A {{ReceiveStream}} has the following internal slots.
...
-     1. If |stream| has received a RESET_STREAM frame, abort these steps.
-     1. [=In parallel=], [=send STOP_SENDING=] over |stream|'s
-        [=WebTransport stream=] with |reason| as the Application Protocol Error
-         Code.
+<div algorithm>
+
+To <dfn for="ReceiveStream">pull bytes</dfn> from a {{ReceiveStream}} |stream|, run these steps.
+
+1. Let |controller| be |stream|'s [=ReadableStream/[[controller]]=].
+1. Assert: |controller| is a {{ReadableByteStreamController}}.
```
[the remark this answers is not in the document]

index.bs

index.bs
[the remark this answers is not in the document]
yutakahirano: I think this is now addressed - what do you think about the newest change?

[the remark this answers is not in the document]
domenic: No, you are still manually constructing a view. My suggestion is the following:

Replace:

```
  1. If [=ReadableStream/current BYOB request view=] for |stream| is null:
    1. Let |arrayBuffer| be |view|.\[[ViewedArrayBuffer]].
    1. Let |offset| be |view|.\[[ByteOffset]].
    1. Set |view| to a [=new=] {{Uint8Array}} with |arrayBuffer|, |offset| and |read|.
    1. If |read| > 0, then [=ReadableStream/enqueue=] |view| into |stream|.
  1. Otherwise:
    1. If |read| > 0, then [=ReadableStream/enqueue=] |view| into |stream| with |read|.
```

with

```
  1. If |read| > 0, then [=ReadableStream/enqueue=] |view| into |stream| with |read|.
```

Then:

- If there was a BYOB request view back at the top of the algorithm, you are re-enqueuing that same view, but passing along |read|. The streams machinery will create a new view with the same backing buffer but with only |read| as its byte length.
- If there was no BYOB request view back at the top of the algorithm, you are enqueuing the new view you allocated at the top of the algorithm, but passing along |read|. The streams machinery will create a new view with the same backing buffer but with only |read| as its byte length, instead of "an implementation-defined size".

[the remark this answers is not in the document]
yutakahirano: That will be rejected with the assertion you added in whatwg/streams#1146. 

```
   1. Assert: if |bytesWritten| was given, it's equal to |chunk|.\[[ByteLength]]. (Anything else
      indicates a logic error in the calling specification.)
```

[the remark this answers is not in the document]
domenic: You are right; sorry, I was very confused. I keep making mistakes on this :(

I think what I meant to suggest was the following: replace

```
1. Let |view| be [=ReadableStream/current BYOB request view=] for |stream|.
1. If |view| is null, set |view| to a [=new=] {{Uint8Array}} with an [=implementation-defined=]
   size. If allocating the {{Uint8Array}} fails, return a promise [=rejected with=] a
   {{RangeError}}.
1. Let |bytes| be the byte sequence |view| represents.
1. Assert: |bytes|'s length equals to |view|.\[[ByteLength]].
1. Return |promise| and run the remaining steps [=in parallel=].
1. [=ArrayBufferView/Write=] the bytes that area [=stream/receive|read=] from |internalStream| into
   |view|, up to |bytes|'s length bytes. Wait until either at least one byte is read or FIN is
   received. Let |read| be the number of read bytes, and let |hasReceivedFIN| be whether FIN was
   accompanied.
```

with

```
1. Let |buffer|, |offset|, and |maxBytes| be null.
1. If |stream|'s [=ReadableStream/current BYOB request view=] for |stream| is not null, then:
  1. Set |offset| to |stream|'s [=ReadableStream/current BYOB request view=].\[[ByteOffset]].
  1. Set |maxBytes| to |stream|'s [=ReadableStream/current BYOB request view=]'s [=BufferSource/byte length=].
  1. Set |buffer| to |stream|'s [=ReadableStream/current BYOB request view=]'s [=BufferSource/underlying buffer=].
1. Otherwise,
  1. Set |offset| to 0.
  1. Set |maxBytes| to an [=implementation-defined=] size.
  1. Set |buffer| be a [=new=] {{ArrayBuffer}} with |maxBytes| size. If allocating the {{Uint8Array}} fails, return a promise [=rejected with=] a {{RangeError}}.
1. Return |promise| and run the remaining steps [=in parallel=].
1. [=ArrayBuffer/Write=] the bytes that area [=stream/receive|read=] from |internalStream| into
   |buffer|, up to |maxBytes| bytes. Wait until either at least one byte is read or FIN is
   received. Let |read| be the number of read bytes, and let |hasReceivedFIN| be whether FIN was
   accompanied.
```

and then replace

```
  1. If [=ReadableStream/current BYOB request view=] for |stream| is null:
    1. Let |arrayBuffer| be |view|.\[[ViewedArrayBuffer]].
    1. Let |offset| be |view|.\[[ByteOffset]].
    1. Set |view| to a [=new=] {{Uint8Array}} with |arrayBuffer|, |offset| and |read|.
    1. If |read| > 0, then [=ReadableStream/enqueue=] |view| into |stream|.
  1. Otherwise:
    1. If |read| > 0, then [=ReadableStream/enqueue=] |view| into |stream| with |read|.
```

with

```
  1. If |read| > 0, then:
    1. Set |view| to a [=new=] {{Uint8Array}} with |buffer|, |offset| and |read|.
    1. [=ReadableStream/Enqueue=] |view| into |stream|.
```

WDYT?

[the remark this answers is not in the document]
yutakahirano: Thank you, sounds reasonable.

I accepted the suggestion with adding _offset_ at the reading step, with some minor style fixes.

With this we don't need whatwg/streams#1146, I think.

domenic (approved):
Looks great. Thank you for the helpful back and forth, and your patience along the way!

yutakahirano:
@ricea, can you take a look again?

index.bs
```diff
@@ -1294,44 +1296,114 @@ A {{ReceiveStream}} has the following internal slots.
...
+  1. Set |maxBytes| to an [=implementation-defined=] size.
+  1. Set |buffer| be a [=new=] {{ArrayBuffer}} with |maxBytes| size. If allocating the
+     {{Uint8Array}} fails, return a promise [=rejected with=] a {{RangeError}}.
+1. Return |promise| and run the remaining steps [=in parallel=].
+1. [=ArrayBuffer/Write=] the bytes that area [=stream/receive|read=] from |internalStream| into
+   |buffer| with offset |offset|, up to |maxBytes| bytes. Wait until either at least one byte is
+   read or FIN is received. Let |read| be the number of read bytes, and let |hasReceivedFIN| be
+   whether FIN was accompanied.
+   Note: The user-agent MAY have a buffer to improve the transfer performance. Such a buffer
+   SHOULD have a fixed size, to carry the backpressure information to the server.
```
ricea, line 1353: Can we say "fixed upper limit" instead, to allow for implementations that share a shared memory segment for several purposes, etc.

index.bs
```diff
@@ -1294,44 +1296,114 @@ A {{ReceiveStream}} has the following internal slots.
...
+  1. Set |offset| to |stream|'s [=ReadableStream/current BYOB request view=].\[[ByteOffset]].
+  1. Set |maxBytes| to |stream|'s [=ReadableStream/current BYOB request view=]'s
+     [=BufferSource/byte length=].
+  1. Set |buffer| to |stream|'s [=ReadableStream/current BYOB request view=]'s
+     [=BufferSource/underlying buffer=].
+1. Otherwise:
+  1. Set |offset| to 0.
+  1. Set |maxBytes| to an [=implementation-defined=] size.
+  1. Set |buffer| be a [=new=] {{ArrayBuffer}} with |maxBytes| size. If allocating the
+     {{Uint8Array}} fails, return a promise [=rejected with=] a {{RangeError}}.
```
ricea, line 1346:
```suggestion
     {{ArrayBuffer}} fails, return a promise [=rejected with=] a {{RangeError}}.
```

index.bs
```diff
@@ -1294,44 +1296,114 @@ A {{ReceiveStream}} has the following internal slots.
...
+   read or FIN is received. Let |read| be the number of read bytes, and let |hasReceivedFIN| be
+   whether FIN was accompanied.
+   Note: The user-agent MAY have a buffer to improve the transfer performance. Such a buffer
+   SHOULD have a fixed size, to carry the backpressure information to the server.
+
+   Note: This operation may return before filling up all of |bytes|.
+
+1. If the previous step failed, abort the remaining steps.
+
+  Note: We don't reject |promise| here because we handle network errors elsewhere, and those logic
```
ricea, line 1359: I suggest "steps" rather than "logic".

index.bs
```diff
@@ -1294,44 +1296,114 @@ A {{ReceiveStream}} has the following internal slots.
...
+  error |stream| and reject the result of this read operation.
+
+1. [=Queue a network task=] with |transport| to run these steps:
+
+   Note: If the buffer described above is available in the [=event loop=] where this procedure is
+   running, the following steps may run immediately.
+
+  1. If |read| > 0:
+    1. Set |view| to a [=new=] {{Uint8Array}} with |buffer|, |offset| and |read|.
```
ricea, line 1368: Nit: WebIDL's `new` doesn't call the constructor. I think you need to do something like
```
1. Set |view| to ! Construct(constructor, « |buffer|, |offset|, |read| »).
```
as in https://heycam.github.io/webidl/#arraybufferview-create

index.bs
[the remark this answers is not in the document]
ricea: Can we also have a note for the `cancel` algorithm?

ricea (approved):
lgtm with nits.

index.bs
domenic: I would suggest instead just removing the link so it's `new {{Uint8Array}}`. This is IMO a missing primitive that Web IDL should supply, similar to how it supplies https://heycam.github.io/webidl/#arraybufferview-create .

It's better to avoid reaching into ES spec internals when we can.

index.bs

index.bs

index.bs
yutakahirano: Thank you, done.

index.bs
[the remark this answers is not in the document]
yutakahirano: Done.

ricea:
lgtm

yutakahirano:
Thank you!

Issue #306, opened 2021-07-13
Remove SendStream and ReceiveStream


yutakahirano:
Now they have no attributes or members except for inherited ones.

Let's remove them and use WritableStream and ReadableStream instead.

PR #307, opened 2021-07-13
Remove "SendStream" Web IDL interface

yutakahirano:
Have it a spec-only concept for streams created by the
"create a SendStream" procedure.

This is for #306.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/pull/307.html" title="Last updated on Jul 16, 2021, 8:07 AM UTC (d6b9c09)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/307/3be0e20...d6b9c09.html" title="Last updated on Jul 16, 2021, 8:07 AM UTC (d6b9c09)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1122,7 +1122,7 @@ To <dfn export for="SendStream" lt="create|creating">create</dfn> a
 {{SendStream}}, with an [=outgoing unidirectional=] or [=bidirectional=] [=WebTransport stream=]
 |internalStream| and a {{WebTransport}} |transport|, run these steps:
 
-1. Let |stream| be a [=new=] {{SendStream}}, with:
+1. Let |stream| be a [=new=] {{WritableStream}}, with:
     : [=[[InternalStream]]=]
     :: |internalStream|
     : [=[[PendingOperation]]=]
--- a/index.bs
+++ b/index.bs
@@ -1078,7 +1078,7 @@ The dictionary SHALL have the following attributes:
 :: The number of datagrams that were dropped, due to too many datagrams buffered
    between calls to {{DatagramTransport/datagrams}}' {{DatagramDuplexStream/readable}}.
 
-# Interface `SendStream` #  {#send-stream}
+# `SendStream` #  {#send-stream}
 
 In this spec, we call a {{WritableStream}} providing outgoing streaming features with an [=outgoing
 unidirectional=] or [=bidirectional=] [=WebTransport stream=] a <dfn>SendStream</dfn>.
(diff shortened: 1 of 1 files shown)


yutakahirano:
@ricea, can you take a look?

I'm leaning to pretending the internal slots to be attached to `SendStream` (or `WritableStream`). We could define another holder object and pass it to write/close/abort algorithms, but that would be too confusing, given we already have `stream` and `[[InternalStream]]` in them.

ricea:
I'm concerned that implementers may feel they actually have to create a way to attach extra data to a WritableStream. But maybe we could make it clearer in the note that this is just an abstraction. Maybe we could call it a "send stream" instead of a SendStream to reduce the expectation that it would literally appear in the implementation.

yutakahirano:
Added a note.

index.bs
```diff
@@ -1117,6 +1114,11 @@ A {{SendStream}} has the following internal slots.
  <tbody>
 </table>
 
+Note: These internal slots need to be attached to {{SendStream}} (which is a kind of
+{{WritableStream}}s) conceptually, not phisically. For example, implementers MAY have a weak hash
```
ricea, line 1118: Maybe the "s" in `{{WritableStream}}s` isn't needed.

Spelling: "physically".

I don't think you should use `MAY` within a note section.

In general, I would prefer to recommend an implementation strategy where the slots are owned by an object which is owned by the algorithms.

Something like
```
For example, an implementation could place them on an object which is
referenced from writeAlgorithm, closeAlgorithm and abortAlgorithm.
```

yutakahirano: Fixed nits: thank you!

> In general, I would prefer to recommend an implementation strategy where the slots are owned by an object which is owned by the algorithms.

I can do that: Just to confirm, are you fine with the recommendation with https://w3c.github.io/webtransport/#send-stream-STOP_SENDING?

ricea: I'm confused. That algorithm references a "remoteCanceled" promise which isn't defined anywhere.

yutakahirano: Ah sorry I forgot to remove the step. I'll do that separately.

What I'm talking about here is the reference to the `[[Transport]]` internal slot at the first step.

> 1. Let _transport_ be _stream_’s [[Transport]].

ricea: I think that's okay. It's not ambiguous.

yutakahirano: Fixed.

ricea (approved):
lgtm.

jan-ivar (approved):

PR #311, opened 2021-07-19
Define DatagramDuplexStream.maxDatagramSize

### LINKED ISSUE #301: WebTransport.maxDatagramSize bikeshedding
1. Should it be part of WebTransport or DatagramDuplexStream?
1. Should it be "maxDatagramSize" or "maxOutgoingDatagramSize"?
1. Is `unsigned short` enough?
1. Should it be a Promise? According to the [spec](https://datatracker.ietf.org/doc/html/draft-ietf-quic-datagram#section-3) the value is a transport parameter, so it is not available when the transport is contructed.

### LINKED ISSUE #302: Datagrams readiness
As described at https://datatracker.ietf.org/doc/html/draft-ietf-quic-datagram#section-3, we must not send a datagram until the transport parameter exchange is done.

We need to choose one from the following:

 1. There are two readiness, one for WebTransport over HTTP/3 and the other for datagrams.
 2. There are only one readiness. When both of the CONNECT response and the datagram transport parameters arrive at the client, the transport is ready.

Which do you prefer?

yutakahirano:
- Define the internal slot.
 - Use the internal slot in writeDatagram and sendDatagram.
 - Wire up the internal slot and the attribute.

Fixes #301 and #302.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/pull/311.html" title="Last updated on Aug 17, 2021, 11:02 AM UTC (309496c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/311/2fe564d...309496c.html" title="Last updated on Aug 17, 2021, 11:02 AM UTC (309496c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -298,6 +298,7 @@ interface DatagramDuplexStream {
   readonly attribute ReadableStream readable;
   readonly attribute WritableStream writable;
 
+  readonly attribute unsigned long maxDatagramSize;
   attribute double? incomingMaxAge;
   attribute double? outgoingMaxAge;
   attribute long incomingHighWaterMark;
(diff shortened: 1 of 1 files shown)


yutakahirano:
@vasilvv can you take a look?

jan-ivar:
Meeting:
- Move to DatagramDuplexStream?

jan-ivar (approved):

PR #312, opened 2021-07-19
Must not provide any credentials via HTTP headers

### LINKED ISSUE #188: Clarify that we don't support cookies

vasilvv:
Fixes #188.

***
<a href="https://pr-preview.s3.amazonaws.com/vasilvv/web-transport/pull/312.html" title="Last updated on Nov 10, 2021, 6:56 PM UTC (0f08c4a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/312/94d996e...vasilvv:0f08c4a.html" title="Last updated on Nov 10, 2021, 6:56 PM UTC (0f08c4a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -190,6 +190,8 @@ There may be multiple [=WebTransport sessions=] on one [=connection=], when pool
 
 To <dfn for=session>establish</dfn> a [=WebTransport session=], follow [[!WEB-TRANSPORT-HTTP3]]
 [section 3.3](https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3/#section-3.3).
+When establishing a session, the client MUST NOT provide any [=credentials=]
+via HTTP headers.
 
 To <dfn for=session>terminate</dfn> a [=WebTransport session=] |session| with an integer |code|,
 follow [[!WEB-TRANSPORT-HTTP3]]
--- a/index.bs
+++ b/index.bs
@@ -190,8 +190,7 @@ There may be multiple [=WebTransport sessions=] on one [=connection=], when pool
 
 To <dfn for=session>establish</dfn> a [=WebTransport session=], follow [[!WEB-TRANSPORT-HTTP3]]
 [section 3.3](https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3/#section-3.3).
-When establishing a session, the client MUST NOT provide any [=credentials=]
-via HTTP headers.
+When establishing a session, the client MUST NOT provide any [=credentials=].
 
 To <dfn for=session>terminate</dfn> a [=WebTransport session=] |session| with an integer |code|,
 follow [[!WEB-TRANSPORT-HTTP3]]


ricea:
We should probably also disallow client certificates (does QUIC even do those?)

yutakahirano:
> We should probably also disallow client certificates (does QUIC even do those?)

TLS client certificates are part of credentials.

> __Credentials__ are HTTP cookies, TLS client certificates, and authentication entries (for HTTP authentication)
> https://fetch.spec.whatwg.org/#credentials

Modifying "via HTTP headers" may be good.

aboba (approved):

index.bs
```diff
@@ -190,6 +190,8 @@ There may be multiple [=WebTransport sessions=] on one [=connection=], when pool
 
 To <dfn for=session>establish</dfn> a [=WebTransport session=], follow [[!WEB-TRANSPORT-HTTP3]]
 [section 3.3](https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3/#section-3.3).
+When establishing a session, the client MUST NOT provide any [=credentials=]
+via HTTP headers.
```
martinthomson, line 194: Adam mentioned client certificates.  I would remove "via HTTP headers" and then be more explicit and explain that this means "no cookies", "no authorization", and "no client auth".

vasilvv:
"No client auth" is actually somewhat complicated, since it happens at the different stage from sending a request.  The problem is, I'm not sure Fetch has a well-suited knob for turning off the client auth.

martinthomson:
To be clear, I think that this should say "no client auth" period.  That encompasses all of the things, including the difficult certificate-based client authentication @vasilvv is concerned about.  Prohibiting only a subset of the authentication-related things doesn't do that.  I would prefer to say "no client auth that might be populated by the user agent, which includes HTTP headers, cookies, and client certificates".

vasilvv:
The client certificate is a bit tricky here.  In general, we do want to actively pool HTTP/3 connections together with top-level HTTP traffic; however, if we require WebTransport to be non-credentialed, this would almost always not work, since most fetches (including the top-level ones) *are* credentialed.

Similarly with client certificates: if we already have an existing HTTP/3 connection to a server, do we actually want to avoid it just because it has a client cert associated with it

ricea:
> Similarly with client certificates: if we already have an existing HTTP/3 connection to a server, do we actually want to avoid it just because it has a client cert associated with it

Yes. Ambient authentication is dangerous.

martinthomson:
What Adam said. Confused deputy and all that.

Fact is, browsers already need to be able to control whether connections are credentialed. So this isn't so big of a deal.

index.bs
annevk: If this ends up referencing https://fetch.spec.whatwg.org/#credentials removing "via HTTP headers" is probably sufficient.

yutakahirano: +1

jan-ivar, line 194:
```suggestion
When establishing a session, the client MUST NOT provide any [=credentials=].
```

jan-ivar:
Made text changed as requested and merged.

PR #313, opened 2021-07-19
Make fingerprint matching case-insensitive.

### LINKED ISSUE #117: "verify a certificate fingerprint" algorithm (almost) always returns false
The "verify a certificate fingerprint" compares a fingerprint from the "compute a fingerprint" algorithm to fingerprints from the server_certificate_fingerprints list in the QuicTransportConfiguration.

The "compute a fingerprint" algorithm formats the fingerprint according to RFC 8122 section 5, which puts the fingerprint in uppercase hex. The server_certificate_fingerprints list is a sequence of RTCDtlsFingerprint, defined at https://www.w3.org/TR/webrtc/#dictionary-rtcdtlsfingerprint-members, which states that they are lowercase. Since one is uppercase and the other is lowercase, they can never be equal (unless by luck the fingerprint contains only hex digits 0 thru 9).

I think the ideal fix would be to modify the webrtc spec to make its fingerprints uppercase to match RFC 4572 and RFC 8122, but if that's not a practical change to make, then something needs to be changed in web-transport.

vasilvv:
Fixes #117.

***
<a href="https://pr-preview.s3.amazonaws.com/vasilvv/web-transport/pull/313.html" title="Last updated on Jul 19, 2021, 10:54 PM UTC (8f0962b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/313/789727d...vasilvv:8f0962b.html" title="Last updated on Jul 19, 2021, 10:54 PM UTC (8f0962b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1011,9 +1011,10 @@ To <dfn>verify a certificate fingerprint</dfn>, do the following:
 1. Let |referenceFingerprint| be the [=compute a certificate fingerprint|computed
    fingerprint=] of the input certificate.
 1. For every fingerprint |fingerprint| in |fingerprints|:
-   1. If {{RTCDtlsFingerprint/algorithm}} of |fingerprint| is equal to "sha-256",
-      and {{RTCDtlsFingerprint/value}} of |fingerprint| is equal to
-      |referenceFingerprint|, the certificate is valid.  Return true.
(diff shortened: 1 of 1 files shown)


yutakahirano (approved):
LGTM.

PR #288, opened 2021-06-22
Add WebTransportError interface.

jan-ivar:
Fixes https://github.com/w3c/webtransport/issues/252.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webtransport/pull/288.html" title="Last updated on Jun 22, 2021, 10:54 PM UTC (2edcff9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/288/1cb3edc...jan-ivar:2edcff9.html" title="Last updated on Jun 22, 2021, 10:54 PM UTC (2edcff9)">Diff</a>

Issue #252, opened 2021-04-28
Add custom error property for Application Protocol Error Codes?


jan-ivar:
From https://github.com/w3c/webtransport/pull/248#issuecomment-828363738:
> Reg: remoteReset, I think we can add properties to the exception with which closed is rejected, instead of introducing remoteReset. What do you think about the idea?

@yutakahirano Normally, I wouldn't be a fan of overloading errors with application-properties, but it might streamline (no pun intended) error handling, since we are indeed exposing an (application-specific) [server error code](https://quicwg.org/base-drafts/draft-ietf-quic-transport.html#name-application-protocol-error-) to JS here, and our current approach of trying to avoid defining a custom error for this I think misses that we still need to surface an error downstream.

This might be a good time to review our error strategy vis-a-vis with streams spec.

Even though the streams spec allows `any` "reason", my guess is best practice is to create error (DOMException) objects, just like we do for throwing things?

Similarly, calling streams methods with error code numbers to reset a QUIC stream [seems a bit off](https://pr-preview.s3.amazonaws.com/jan-ivar/webtransport/pull/248.html#protocol-mapping):

Method | QUIC Protocol Action
---- | ----
 writable.abort(42) | send RESET_STREAM with 42

Might it be better to do something like this?
```js
writable.abort(new WebTransportError({errorCode: 42}))
```

This would also let us streamline the the wt.[closed](https://w3c.github.io/webtransport/#dom-webtransport-closed) attribute by rejecting instead of resolving on error:
```js
try {
  await wt.closed;
  // success!
} catch (e) {
  if (e.name != "WebTransportError") throw e;
  console.log("Server error: e.errorCode); // 42
}
```
vs. today:
```js
const info = await wt.closed;
if (info.errorCode == 0) {
  // success!
} else {
  console.log("Server error: info.errorCode); // 42
}
```

jan-ivar:
@ricea Thoughts?

ricea:
I seem to recall reading somewhere that we should try to avoid adding new exception types to the platform, but I can't find a reference to it now. @domenic ?

When `Error`s are transferred, only the `message` and `name` properties are cloned. `DOMException` is a special case, and it's possible that `WebTransportError` could be a special case too.

Requiring the user to pass a `WebTransportError` object to `abort()` to customise the `errorCode` seems a little strict. For `WebSocketStream` I made `abort({code: 4200})` work, which seems like a good compromise.

[...]

Issue #214, opened 2021-03-03
Error codes that don't fit


martinthomson:
We need text when the error code value doesn't fit in the protocol.  2<sup>62</sup> would not fit in the protocol, but it is a valid value for `unsigned long long`.

jan-ivar:
Take the lower bits?

martinthomson:
I was thinking `throw`, but whatever works.

yutakahirano:
close() and abort() are cleanup operations - for example they are often called in catch clauses. I think such functions shouldn't throw whenever possible. I prefer defaulting to 0.

jan-ivar:
As @ricea mentioned in https://github.com/w3c/webtransport/pull/288#discussion_r657531877 I think this boils down to us deciding between `[EnforceRange]` and `[Clamp]`.

jan-ivar:
Meeting:
- Go with `[Clamp]`

yutakahirano:
At https://github.com/w3c/webtransport/pull/288#discussion_r657531877 suggested `[Clamp]` because we used `octet` there. Now we need a wider range, and we cannot use `[Clamp]`.

I'm going to introduce a custom `[Clamp]`-like logic. @ricea, please let me know if I'm wrong.

ricea:
IIUC, `[Clamp]` will work for any integer type. https://heycam.github.io/webidl/#Clamp

yutakahirano:
But it doesn't have arguments - we cannot set min/max.

In this case we want to define a custom range - [0, 2 ** 62), so we cannot use `[Clamp]`.

(Moreover, 2 ** 62 is not valid as `unsigned long long` - we need to use bigint).

My proposal is the following.

```
[Exposed=(Window,Worker), SecureContext]
interface WebTransportError : DOMException {
  constructor(optional WebTransportErrorInit init = {});

  readonly attribute (unsigned long long or bigint)? applicationProtocolCode;
};

dictionary WebTransportErrorInit {
  ([Clamp] unsigned long long) or bigint applicationProtocolCode;
  DOMString message;
};
```

Things that are not clear to me:

 - Is it better to have `WebTransportError.applicationProtocolCode` as `(unsigned long long or bigint)?` or `bigint?`?

cc: @yuki3 as a WebIDL expert.

ricea:
Oh, I see. sgtm.

yuki3:
[...]

PR #316, opened 2021-07-26
Introduce WebTransportError

yutakahirano:
Introduce WebTransportError, an interface for error information
related to WebTransport.

This is related to #252. This is based on #288. This PR is aiming to land
a minimal change in order to unblock other issues. For example, we're
 talking about `errorSource` in #288. This PR doesn't contain it.

In the last WG meeting, no one was against introducing this interface, and
people prefered having one interface and multiple properties to having
multiple error interfaces and fewer properties. Hence the contents
of this shouldn't be super controversial I believe.

This change addresses most of comments made by @ricea at #288. One exception
is how to handle out-of-range value. I think it is the same problem as #214, and
should be solved altogether.

I want to merge this PR at the next WG meeting (8/3), but comments
before the meeting are welcomed.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/pull/316.html" title="Last updated on Jul 30, 2021, 12:26 PM UTC (7499229)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/316/723a113...7499229.html" title="Last updated on Jul 30, 2021, 12:26 PM UTC (7499229)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1375,7 +1375,7 @@ object |transport|, run these steps.
 <dfn interface>WebTransportError</dfn> is a subclass of {{DOMException}} that represents
 
  - An error coming from the server or the network, or
- - An reason for a client initiated abort operation.
+ - A reason for a client-initiated abort operation.
 
 <pre class="idl">
 [Exposed=(Window,Worker), SecureContext]
--- a/index.bs
+++ b/index.bs
@@ -1444,6 +1444,8 @@ To <dfn for="WebTransportError">set up</dfn> a {{WebTransportError}} |error| wit
    Note: This name does not have a mapping to a legacy code, so |error|'s {{DOMException/code}}
    is 0.
 
+</div>
+
 ## Configuration ## {#web-transport-error-configuration}
 
 <pre class="idl">
--- a/index.bs
+++ b/index.bs
@@ -73,6 +73,7 @@ spec:html; type:dfn; for:/; text:origin
 spec:fetch; type:dfn; for:/; text:fetch
 spec:url; type:dfn; text:scheme
 spec:url; type:dfn; text:fragment
+spec:webidl; type:dfn; for:/; text:new DOMException(message, name)
 </pre>
 <pre class="anchors">
 urlPrefix: http://www.ecma-international.org/ecma-262/6.0/index.html; spec: ECMASCRIPT-6.0
@@ -1402,7 +1403,6 @@ A {{WebTransportError}} has the following internal slots.
    <td><dfn>\[[ApplicationProtocolCode]]</dfn>
    <td class="non-normative">The application protocol error code for this error.
   </tr>
-  <tr>
  </tbody>
 </table>
 
--- a/index.bs
+++ b/index.bs
@@ -84,7 +84,7 @@ urlPrefix: http://www.ecma-international.org/ecma-262/6.0/index.html; spec: ECMA
     text: settled; url: sec-promise-objects
 urlPrefix: https://heycam.github.io/webidl/; spec: WEBIDL
   type: dfn
-    text: new DOMException(message, name); for:DOMException; url: ref-for-dom-domexception-domexception
+    text: new DOMException(message, name); for:DOMException; url: dom-domexception-domexception
 </pre>
 
 # Introduction #    {#introduction}
--- a/index.bs
+++ b/index.bs
@@ -1410,7 +1410,8 @@ A {{WebTransportError}} has the following internal slots.
 
 <div algorithm>
 
-The <dfn constructor for="WebTransportError">new
+The <dfn constructor for="WebTransportError"
+lt="WebTransportError(applicationProtocolCode, message)">new
 WebTransportError(applicationProtocolCode, message)</dfn> constructor steps are:
 
 1. Let |error| be [=this=].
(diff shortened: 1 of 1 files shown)


yutakahirano:
cc: @jan-ivar

jan-ivar (changes requested):
Thanks @yutakahirano  for picking this up, and narrowing its scope. I've noted some changes I think it still needs.

Now that the scope has been narrowed, it would be nice if the PR also included using this new interface for RESET_STREAM and STOP_SENDING.

If this is to be done in a follow-up, it would be nice to merge them at the same time or within a short period of time, to not leave the document inconsistent.

index.bs
```diff
@@ -1370,6 +1370,74 @@ object |transport|, run these steps.
...
+   <th>Internal Slot
+   <th>Description (<em>non-normative</em>)
+  </tr>
+ </thead>
+ <tbody>
+  <tr>
+   <td><dfn>\[[ApplicationProtocolCode]]</dfn>
+   <td class="non-normative">The application protocol error code for this error.
+  </tr>
+  <tr>
```
ricea, line 1405: Should this empty `<tr>` be here?

index.bs
```diff
@@ -1370,6 +1370,74 @@ object |transport|, run these steps.
...
+## Attributes ## {#web-transport-error-methods}
+
+: <dfn for="WebTransportError" attribute>applicationProtocolCode</dfn>
+:: The getter steps are to return [=this=]'s [=[[ApplicationProtocolCode]]=].
+
+## Procedures ## {#web-transport-error-procedures}
+
+<div algorithm>
+
+To <dfn for="WebTransportError">set up</dfn> a {{WebTranspoerError}} |error| with a DOMString
```
ricea, line 1432:
```suggestion
To <dfn for="WebTransportError">set up</dfn> a {{WebTransportError}} |error| with a DOMString
```

index.bs
```diff
@@ -1370,6 +1370,74 @@ object |transport|, run these steps.
...
+|message|, run these steps:
+
+1. Run {{new DOMException(message, name)}} constructor steps with |error|, |message| and
```
ricea, line 1435: Looks like this should link to https://heycam.github.io/webidl/#dom-domexception-domexception. I think we can do it in the anchors sections.

yutakahirano: Sorry, what do you mean by "in the anchors sections"?

ricea: Sorry that wasn't clear. I meant the `<pre class="anchors">` section at the top of the file. Since it is defined but not exported in WebIDL, I think putting it in the `<pre class="link-defaults">` section would actually be sufficient.

index.bs
```diff
@@ -1370,6 +1370,74 @@ object |transport|, run these steps.
...
+
+<dfn interface>WebTransportError</dfn> is a subclass of {{DOMException}} that represents
+
+ - An error coming from the server or the network, or
+ - An reason for a client initiated abort operation.
+
+<pre class="idl">
+[Exposed=(Window,Worker), SecureContext]
+interface WebTransportError : DOMException {
+  constructor(octet applicationProtocolCode, DOMString message);
```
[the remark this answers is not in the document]
ricea: I agree we probably need to have `WebTransportErrorInit` here.

index.bs
yutakahirano: No, thank you. Removed.

index.bs

index.bs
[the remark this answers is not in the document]
yutakahirano: Added WebTransportInit. `message` is also included in it.

yutakahirano:
Thank you for the comments, I addressed them. PTAL again.

index.bs
```diff
@@ -1370,6 +1373,85 @@ object |transport|, run these steps.
...
+ <thead>
+  <tr>
+   <th>Internal Slot
+   <th>Description (<em>non-normative</em>)
+  </tr>
+ </thead>
+ <tbody>
+  <tr>
+   <td><dfn>\[[ApplicationProtocolCode]]</dfn>
+   <td class="non-normative">The application protocol error code for this error, or null.
```
jan-ivar, line 1411: The spec has two previous mentions of "application protocol code", with language referring to [QUIC] — which is missing here — but I had to search the document for them.

It would be nice to link them together, and point them to a new definition under concepts.

index.bs
```diff
@@ -1370,6 +1373,85 @@ object |transport|, run these steps.
...
+1. [=WebTransportError/Set up=] |error| with |message|.
+1. Set |error|'s [=[[ApplicationProtocolCode]]=] to |init|'s
+   {{WebTransportErrorInit/applicationProtocolCode}} if it exists.
+
+</div>
+
+## Attributes ## {#web-transport-error-attributes}
+
+: <dfn for="WebTransportError" attribute>applicationProtocolCode</dfn>
+:: The getter steps are to return [=this=]'s [=[[ApplicationProtocolCode]]=].
```
jan-ivar, line 1438: This seems a bit terse in describing what the new `error.applicationProtocolCode` is, or (for the constructor) how it should be used.

But I'm happy to iterate on that, since I suppose we still haven't fully decided (if it's the exact QUIC definition, or if we're going to reserve/overload some values (e.g. "other" instead of null)?

jan-ivar (approved):

jan-ivar:
Meeting:
- Consider s/octet/uint64/ or whatever WebIDL calls it. If we want to reuse this attribute for connection close.
- Iterate on `errorSource` etc in other issues.

PR #323, opened 2021-08-05
Use [Clamp] for WebTransportErrorInit.applicationProtocolCode

### LINKED ISSUE #214: Error codes that don't fit

yutakahirano:
Fixes #214.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/pull/323.html" title="Last updated on Aug 17, 2021, 11:41 AM UTC (e4ee7a5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/323/2fe564d...e4ee7a5.html" title="Last updated on Aug 17, 2021, 11:41 AM UTC (e4ee7a5)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1475,7 +1475,7 @@ interface WebTransportError : DOMException {
 };
 
 dictionary WebTransportErrorInit {
-  octet applicationProtocolCode;
+  [Clamp] octet applicationProtocolCode;
   DOMString message;
 };
 </pre>


index.bs
```diff
@@ -1471,11 +1471,11 @@ object |transport|, run these steps.
 interface WebTransportError : DOMException {
   constructor(optional WebTransportErrorInit init = {});
 
-  readonly attribute octet? applicationProtocolCode;
+  readonly attribute (unsigned long long or bigint)? applicationProtocolCode;
 };
 
 dictionary WebTransportErrorInit {
-  octet applicationProtocolCode;
+  (([Clamp] unsigned long long) or bigint) applicationProtocolCode;
```
ricea, line 1478: Maybe we should always return bigint? It would avoid callers needing to special case depending on the return type.

yutakahirano:
Removed bigint based on the discussion between @ricea and @vasilvv. PTAL again.

@vasilvv can you take a look?

ricea (approved):
lgtm

martinthomson:
what discussion?

yutakahirano:
> what discussion?

Supporting big numbers implies using [bigint](https://heycam.github.io/webidl/#idl-bigint).

 - It is a bit hard to use (e.g., comparison with usual numbers).
 - Big numbers cannot be stored in JSON.

We can avoid these problems by removing the support, i.e., restricting the code space.

index.bs
yutakahirano: (Removed bigint)

martinthomson:
So 2^53 is a big number, so you are justified in terms of limiting space.

...however.  The registry rules for QUIC explicitly recommend using random selection for codepoints.  That means that you have just cut out most (99.8%) of the codepoints people might choose unless they explicitly choose smaller encodings.  That's probably OK for webtransport-first protocols where it is easy to ensure that you don't choose incompatible codepoints, but it's not good otherwise.  What's wrong with just using bigint uniformly?

yutakahirano:
> That means that you have just cut out most (99.8%) of the codepoints people might choose unless they explicitly choose smaller encodings. That's probably OK for webtransport-first protocols where it is easy to ensure that you don't choose incompatible codepoints, but it's not good otherwise.

Is it reasonable to offset the codes similar to what we're doing at https://github.com/ietf-wg-webtrans/draft-ietf-webtrans-http3/pull/59? We map [0, 255] to [0x52e4a40fa8db, 0x52e4a40fa9e2] there.

> What's wrong with just using bigint uniformly?

- It is a bit hard to use (e.g., `1n === 1` returns false).
- Big numbers cannot be stored in JSON.

martinthomson:
> Is it reasonable to offset the codes[...]?

This depends on whether you think that the codes can be subsetted for WebTransport use.  If they were, then the 256 codes you mention would be enough.  No point in trying to capture other values.

However, I think that the entire purpose of this affordance is to capture all possible error codes.  If the transport generates an error in the [2^53,2^62) range, we won't be able to represent that.  If an existing application wants to use WebTransport, then it will be unable to use existing codes that fall in that range.

Though bignums are annoying (in a great many ways), the alternative is to deny some number of QUIC-based protocols easy access to WebTransport.  Or, we create incentives to avoid the high parts of the error code range that are hidden in a place that people won't find, perhaps not until it is difficult for them to adapt.

yutakahirano:
@ricea @vasilvv do you have any opinions?

vasilvv:
> ...however. The registry rules for QUIC explicitly recommend using random selection for codepoints. That means that you have just cut out most (99.8%) of the codepoints people might choose unless they explicitly choose smaller encodings. That's probably OK for webtransport-first protocols where it is easy to ensure that you don't choose incompatible codepoints, but it's not good otherwise. What's wrong with just using bigint uniformly?

I am confused by this claim.  None of the error codes that would go into `applicationProtocolCode` in any way overlap with QUIC or any other IANA error spaces.

martinthomson:
Not sure what @vasilvv is talking about.  

Let me restate my position: if the purpose of the field is to carry a QUIC error code, then it needs to carry all possible values.  

If the purpose of the field is to carry a WebTransport error code (for which we have agreed there are just 256), then a byte will do.

I thought that we were talking about the former, not the latter.

vasilvv:
> If the purpose of the field is to carry a WebTransport error code (for which we have agreed there are just 256), then a byte will do.

The purpose is to provide two error codes: one in RESET_STREAM (8-bit), another is CLOSE_WEBTRANSPORT_SESSION as defined in ietf-wg-webtrans/draft-ietf-webtrans-http3#60 (size TBD).  I don't think we want to communicate QUIC connection errors directly outside of just some variation of network error.

yutakahirano:
> another is CLOSE_WEBTRANSPORT_SESSION as defined in ietf-wg-webtrans/draft-ietf-webtrans-http3#60 (size TBD)

Should it be represented by a `WebTransportError` (as a rejection of `.closed`) or `WebTransportCloseInfo` (as a fulfillment of `.closed`)?

vasilvv:
> > another is CLOSE_WEBTRANSPORT_SESSION as defined in [ietf-wg-webtrans/draft-ietf-webtrans-http3#60](https://github.com/ietf-wg-webtrans/draft-ietf-webtrans-http3/pull/60) (size TBD)
> 
> Should it be represented by a `WebTransportError` (as a rejection of `.closed`) or `WebTransportCloseInfo` (as a fulfillment of `.closed`)?

Depends on the context.  I think `.closed` should be fulfilled, not rejected, since this is a normal closure.  That said, for the *streams* that are still open when that happens, this will be an error.

yutakahirano:
Makes sense, thank you!

yutakahirano:
I removed the `unsigned long long` from the PR. Now this PR simply adds `[Clamp]`, given the @vasilvv's comment

> I don't think we want to communicate QUIC connection errors directly outside of just some variation of network error.

If we need more than 8 bits for the code for CLOSE_WEBTRANSPORT_SESSION, I'll create another PR.

> The purpose is to provide two error codes: one in RESET_STREAM (8-bit), another is CLOSE_WEBTRANSPORT_SESSION as defined in ietf-wg-webtrans/draft-ietf-webtrans-http3#60 (size TBD).

jan-ivar (approved):

Issue #317, opened 2021-07-26
WebTransportError.isInitiatedByClient


yutakahirano:
I think it'd be benefitial to have this boolean attribute in `WebTransportError`.

The attribute returns whether the error comes from the client. This will allow the script author to distinguish whether the stream (or session? There is no way to abort a session/connection from the script I believe, though...) is aborted because of the abort request from the client, or some issues in the server or network.

I don't know which milestone is appropriate - On the one hand, this is a nice to have. On the other hand, we may want to check this attribute in cancelArgorithm and abortAlgorithm, and that will be an incompatible change and hence we may want to do that before in the initial release.

What do you think?

vasilvv:
This would make sense, as we have similar boolean in our QUIC API.  One thing I will note though is that it should be possible to communicate this information via `errorSource` field (or whatever we call it), so we have two approaches here to choose from.

jan-ivar:
Meeting:
- Preference for enum
- Open to bikeshedding further.  `something = "client" | "server"` ? `initiator` not the best since clients don't technically initiate errors, they just happen

jan-ivar:
> (or session? There is no way to abort a session/connection from the script I believe, though...)

The script can call [`wt.close(closeInfo)`](https://w3c.github.io/webtransport/#dom-webtransport-close). But I don't think this spec should solve a script not knowing it did that.

I'm tempted to say the same about client stream errors from `stream.abort(new WebTransportError({streamErrorCode}))`.

wilaw:
Closed by #329

PR #329, opened 2021-08-13
Introduce WebTransportError/source

### LINKED ISSUE #252: Add custom error property for Application Protocol Error Codes?

yutakahirano:
...to allow web developers to distinguish
stream/session/connection errors.

Fixes #252.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/pull/329.html" title="Last updated on Aug 17, 2021, 11:50 PM UTC (9b6f389)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/329/2fe564d...9b6f389.html" title="Last updated on Aug 17, 2021, 11:50 PM UTC (9b6f389)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1483,7 +1483,6 @@ dictionary WebTransportErrorInit {
 enum WebTransportErrorSource {
   "stream",
   "session",
-  "connection",
 };
 </pre>
 
--- a/index.bs
+++ b/index.bs
@@ -1472,11 +1472,18 @@ interface WebTransportError : DOMException {
   constructor(optional WebTransportErrorInit init = {});
 
   readonly attribute octet? applicationProtocolCode;
+  readonly attribute WebTransportErrorSource source;
 };
 
 dictionary WebTransportErrorInit {
   octet applicationProtocolCode;
   DOMString message;
+};
+
+enum WebTransportErrorSource {
+  "stream",
+  "session",
+  "connection",
 };
 </pre>
 
@@ -1496,6 +1503,10 @@ A {{WebTransportError}} has the following internal slots.
    <td><dfn>\[[ApplicationProtocolCode]]</dfn>
    <td class="non-normative">The application protocol error code for this error, or null.
   </tr>
+  <tr>
+   <td><dfn>\[[Source]]</dfn>
+   <td class="non-normative">A {{WebTransportErrorSource}} indicating the source of this error.
+  </tr>
  </tbody>
 </table>
 
@@ -1512,7 +1523,7 @@ lt="WebTransportError(init)">new WebTransportError(init)</dfn> constructor steps
     :: null
 1. Let |message| be `""`.
 1. Set |message| to |init|'s {{WebTransportErrorInit/message}} if it exists.
-1. [=WebTransportError/Set up=] |error| with |message|.
+1. [=WebTransportError/Set up=] |error| with |message| and `"stream"`.
 1. Set |error|'s [=[[ApplicationProtocolCode]]=] to |init|'s
    {{WebTransportErrorInit/applicationProtocolCode}} if it exists.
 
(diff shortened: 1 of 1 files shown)


yutakahirano:
@jan-ivar @ricea PTAL.

vasilvv (approved):

index.bs
```diff
@@ -1472,12 +1472,19 @@ interface WebTransportError : DOMException {
   constructor(optional WebTransportErrorInit init = {});
 
   readonly attribute octet? applicationProtocolCode;
+  readonly attribute WebTransportErrorSource source;
 };
 
 dictionary WebTransportErrorInit {
```
ricea, line 1478: We're probably going to be asked by the TAG to add `source` to `WebTransportErrorInit`. We might as well do it now, unless it will complicate matters.

ricea (approved):
lgtm

index.bs
yutakahirano: We don't have a plan to introduce a feature that initiates a sesson/connection level error. Is it reasonable to allow to create such an error?

jan-ivar: While I could see some marginal benefit of not being able to fake this value, I think the ability to shim errors is generally considered a good thing.

index.bs
```diff
@@ -1472,12 +1472,19 @@ interface WebTransportError : DOMException {
...
   octet applicationProtocolCode;
   DOMString message;
 };
+
+enum WebTransportErrorSource {
+  "stream",
+  "session",
+  "connection",
```
jan-ivar, line 1486: I think these values need explanations. What's the difference between `"session"` vs. `"connection"`, and how might an application react differently?

index.bs
jan-ivar: > We don't have a plan to introduce a feature that initiates a sesson/connection level error.

Hmm, does this mean we could split `"stream"` into `"client"` and `"server"` to solve https://github.com/w3c/webtransport/issues/317?

index.bs
jan-ivar, line 1486:
```suggestion
```

PR #346, opened 2021-09-02
Replace TypeError with WebTransportError

### LINKED ISSUE #344: Use WebTransportError instead of TypeError for connection init failure
In https://github.com/w3c/webtransport/issues/338#issuecomment-904694997 @yutakahirano wrote what looks like reasonable parsing of WebTransport errors (but it has a bug):
> ```js
> try {
>   await stream_reader.closed;
> } catch (e1) {
>   // The stream is errored!
>   if (e1.source === 'stream') {
>     // The stream is reset. the code is e1.applicationProtocolCode and there is no reason text.
>   } else if (e1.source == 'session') {
>     assert_equals(e1.applicationProtocolCode, null);
>     try {
>       const close_info = await transport.closed;
>       // The session is closed cleanly. close_info.code is the code and close_info.reason is the reason text.
>     } catch (e2) {
>       // The session is closed uncleanly. Neither code nor reason text is provided in this case.
>     }
>   } else {
>     NOTREACHED();
>   }
> }
> ```

The bug is `NOTREACHED()` since we can get `TypeError` here if the WebTransport connection hasn't established yet [and fails](https://w3c.github.io/webtransport/#initialize-webtransport-over-http).

Instead of `TypeError`, I propose we fail with `WebTransportError` with `source` = `'session'` and `streamErrorCode` = `null`.

This would fix this particular footgun, by aligning this failure path with later connection failures and https://github.com/w3c/webtransport/pull/335.

yutakahirano:
Replace some TypeError occurences with WebTransportError. I didn't
replace type related errors such as the case where writeDatagrams is
called with a value that is not a BufferSource.

Fixes #344.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/pull/346.html" title="Last updated on Sep 6, 2021, 5:13 AM UTC (1a7b710)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/346/bf1ec50...1a7b710.html" title="Last updated on Sep 6, 2021, 5:13 AM UTC (1a7b710)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -955,6 +955,8 @@ Whenever a [=WebTransport session=] which is associated with a {{WebTransport}} 
 1. Let |cleanly| be a boolean representing whether the HTTP/3 stream associated with the CONNECT
    request that initiated |transport|'s [=[[Session]]=] is in the "Data Recvd" state. [[!QUIC]]
 1. [=Queue a network task=] with |transport| to run these steps:
+  1. Let |code| be TBD.
+  1. Let |reasonString| be TBD.
   1. If |transport|'s [=[[State]]=] is `"closed"` or `"failed"`, abort these steps.
   1. Let |error| be the result of [=WebTransportError/creating=] a {{WebTransportError}} with
      `"session"`.
--- a/index.bs
+++ b/index.bs
@@ -653,7 +653,7 @@ agent MUST run the following steps:
 1. Let |serverCertificateFingerprints| be {{WebTransport/constructor(url, options)/options}}'s
    {{WebTransportOptions/serverCertificateFingerprints}} if it exists, and null otherwise.
 1. If |dedicated| is false and |serverCertificateFingerprints| is non-null, then [=throw=]
-   the result of [=WebTransportError/creating=] a {{WebTransportError}} with `"session"`.
+   a {{TypeError}}.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -651,8 +652,8 @@ agent MUST run the following steps:
 1. Let |dedicated| be the negation of |allowPooling|.
 1. Let |serverCertificateFingerprints| be {{WebTransport/constructor(url, options)/options}}'s
    {{WebTransportOptions/serverCertificateFingerprints}} if it exists, and null otherwise.
-1. If |dedicated| is false and |serverCertificateFingerprints| is non-null, then [=throw=] a
-   {{TypeError}}.
+1. If |dedicated| is false and |serverCertificateFingerprints| is non-null, then [=throw=]
+   the result of [=WebTransportError/creating=] a {{WebTransportError}} with `"session"`.
```
ricea, line 656: I feel like this one should stay a TypeError, since it refers to the parsing of the arguments rather than any property of the session, and it happens synchronously during construction.

yutakahirano: Makes sense, done.

ricea (approved):
lgtm

PR #358, opened 2021-09-22
Use BufferSource for the hash value

### LINKED ISSUE #355: RTCDtlsFingerprint values are strings
This makes sense in WebRTC as the expected source of this information is SDP.  But parsing and validating this string is annoying.  And the most natural form for a fingerprint (ahem hash) is binary data.  I suggest we move to a new dictionary:

```webidl
dictionary WTHash {
  DOMString algorithm;
  BufferSource hash;
}
````

Now, I'm not sure what the convention is for taking binary inputs.  [`BufferSource`](https://heycam.github.io/webidl/#BufferSource) might work here.  I was concerned that maybe the lack of a clear endianness would be an issue here, but it seems like WebCrypto uses that with no issue, so I'm going to assume that this is the right answer.

### PR DESCRIPTION
**Pull request by yutakahirano**
Currently it's DOMString, but that led to some complexity such as case sensitivity. This PR changes
the type to BufferSource, so that we can use the simple byte sequence comparison.

Fixes #355.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/pull/358.html" title="Last updated on Sep 24, 2021, 1:11 AM UTC (763c7a7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/358/2a34434...763c7a7.html" title="Last updated on Sep 24, 2021, 1:11 AM UTC (763c7a7)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1048,10 +1048,9 @@ that determine how WebTransport connection is established and used.
    trusted if and only if it can successfully [=verify a certificate hash=] against
    {{WebTransportOptions/serverCertificateHashes}}
    and satisfies [=custom certificate requirements=].  The user agent SHALL
-   ignore any hash that uses an unknown {{WebTransportHash/algorithm}}
-   or has a malformed {{WebTransportHash/value}}.  If empty, the user agent
-   SHALL use certificate verification procedures it would use for normal
-   [=fetch=] operations.
+   ignore any hash that uses an unknown {{WebTransportHash/algorithm}}.
+   If empty, the user agent SHALL use certificate verification procedures it would
+   use for normal [=fetch=] operations.
 :: This cannot be used with {{WebTransportOptions/allowPooling}}.
 
 <div algorithm>
--- a/index.bs
+++ b/index.bs
@@ -1024,7 +1024,7 @@ The user agent MUST NOT garbage collect a {{WebTransport}} object whose [=[[Stat
 <pre class="idl">
 dictionary WebTransportHash {
   DOMString algorithm;
-  ByteSource value;
+  BufferSource value;
 };
 
 dictionary WebTransportOptions {
--- a/index.bs
+++ b/index.bs
@@ -1022,9 +1022,14 @@ The user agent MUST NOT garbage collect a {{WebTransport}} object whose [=[[Stat
 ## Configuration ##  {#web-transport-configuration}
 
 <pre class="idl">
+dictionary WebTransportHash {
+  DOMString algorithm;
+  ByteSource value;
+};
+
 dictionary WebTransportOptions {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by yutakahirano** [comment]
@martinthomson @vasilvv PTAL. I'm not planning to land this PR before the next WG meeting (Sep 28 or Sep 29), but I want to speed up the specification process and that's why I made this PR.

File: index.bs [hunk 1]
```diff
@@ -1043,8 +1048,8 @@ that determine how WebTransport connection is established and used.
    trusted if and only if it can successfully [=verify a certificate hash=] against
    {{WebTransportOptions/serverCertificateHashes}}
    and satisfies [=custom certificate requirements=].  The user agent SHALL
-   ignore any hash that uses an unknown {{RTCDtlsFingerprint/algorithm}}
-   or has a malformed {{RTCDtlsFingerprint/value}}.  If empty, the user agent
+   ignore any hash that uses an unknown {{WebTransportHash/algorithm}}
+   or has a malformed {{WebTransportHash/value}}.  If empty, the user agent
```
martinthomson, line 1052: We'll want to define what "malformed" means now.  I believe that this is - now - only that it is shorter or longer than the output of the selected algorithm.

index.bs
```diff
@@ -1053,8 +1058,7 @@ that determine how WebTransport connection is established and used.
 To <dfn>compute a certificate hash</dfn>, do the following:
```
martinthomson, line 1058: This should really take the algorithm as an input.  For another issue, perhaps.

index.bs
yutakahirano: "verify a certificate hash" contains the logic. We ignore _hash_ if
 1. _hash_'s value is null,
 2. _hash_'s value doesnt match with _referenceHash_.

The case where the length doesn't match is contained in 2. If we need to define "malformed" we can remove this sentence - What do you think?

martinthomson: WFM.

index.bs
yutakahirano: Filed #359.

yutakahirano:
@martinthomson do you have further comments?

martinthomson (approved):
Nope.

index.bs
```diff
@@ -1022,9 +1022,14 @@ The user agent MUST NOT garbage collect a {{WebTransport}} object whose [=[[Stat
 ## Configuration ##  {#web-transport-configuration}
 
 <pre class="idl">
+dictionary WebTransportHash {
+  DOMString algorithm;
+  ByteSource value;
```
jan-ivar, line 1027:
```suggestion
  BufferSource value;
```

Issue #59, opened 2019-09-18
CSP policies for WebTransport


mfoltzgoogle:
Currently, data exchange capabilities that are exposed to the Web are restricted through [CSP connect-src](https://www.w3.org/TR/CSP3/#directive-connect-src), which restricts which origins can be connected to through WebSockets, XHR, etc.

1. Will every WebTransport destination have an origin that can be verified by the UA (through Web PKI or some other means)?
2. If the answer to "1" is yes, will WebTransport use connect-src or some other CSP policy mechanism?
3. If the answer to "1" is no, how will application authors be able to enforce a WebTransport connection policy in their application?

mfoltzgoogle:
Possibly related to (or a specific issue in) Issue #13

vasilvv:
I think this is yet another issue that can be addressed by providing QuicTransport endpoints with URLs associated with it.  So if we add URL for #22 and #26, we would probably get CSP support for ~free.

aboba:
Discussed on December 8, 2020 WebTransport WG meeting. Agreement to support CSP for the Web PKI case. Concerns over impact of fingerprint verification on CSP support.

jan-ivar:
@vasilvv to write text and reach out to webappsec to discuss some of the more complex cases.

vasilvv:
So I think the answer for WebPKI-authenticated connections is fairly clear: we will check `connect-src` for WebTransport connections.

For the connections that are anchored against a custom certificate, I can see a few approaches:

1. Make this orthogonal to CSP connect-src.  One approach would be to add a new header, say, `Sec-WebTransport-Allow-Custom-Certificates`, that would opt the website into using custom certificates, or somehow encode that into CSP.
1. Encode the fingerprint within the domain name.  An approach I'd gp with is where a domain like `example.com` with certificate fingerprint `0x1234abcd` gets translated into `example.com.domain-1234abcd.fpr.arpa` (I think @ekr proposed something among those lines a few years ago when I discussed that problem with him).

The first approach has the advantage of being very simple, but less powerful, as it's just an all-or-nothing approach.  The second approach lets developers control the specific fingerprints, but it does require us to do more work.  Notably, I have no idea what the policies are wrt assigning `.arpa` domain names, and what are all of the implications.

cc @mikewest since he expressed interest in this topic in the past.
cc @martinthomson and @DavidSchinazi as they probably know more about `.arpa` since

martinthomson:
[...]

PR #367, opened 2021-10-07
Add Content Security Policy check on request

yutakahirano:
We only need the pre-request check. We don't need the post-request
check given there is no redirects on WebTransport.

This is for #59.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/pull/367.html" title="Last updated on Oct 20, 2021, 7:50 AM UTC (e1a7f10)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/367/0a62f0c...e1a7f10.html" title="Last updated on Oct 20, 2021, 7:50 AM UTC (e1a7f10)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -708,6 +708,17 @@ To <dfn>initialize WebTransport over HTTP</dfn>, given a {{WebTransport}} object
 
 1. Let |client| be |transport|'s [=relevant settings object=].
 1. Let |origin| be |client|'s [=environment settings object/origin=].
+1. Let |request| be a new [=request=] whose [=request/URL=] is |url|, [=request/client=] is
+   |client|, [=request/policy container=] is |client|'s
+   [=environment settings object/policy container=], [=request/destination=] is an empty string,
+   and [=request/origin=] is |origin|.
+1. Run <a>report Content Security Policy violations for |request|</a>.
+1. If [=should request be blocked by Content Security Policy?=] with |request| returns
+   <b>blocked</b>, then abort the remaining steps and [=queue a network task=] with |transport|
+   to run these steps:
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -703,6 +703,17 @@ agent MUST run the following steps:
 To <dfn>initialize WebTransport over HTTP</dfn>, given a {{WebTransport}} object
 <var>transport</var>, a [=URL record=] |url|, and a boolean |dedicated|, run these steps.
 
+1. Let |client| be |transport|'s [=relevant settings object=].
+1. Let |request| be a new [=request=] whose [=request/URL=] is |url|, [=request/client=] is
+   |client|, [=request/policy container=] is a clone of |client|'s
```
antosart, line 708: I don't think you need a clone. You can just write `[=request/policy container=] is |client|'s [=environment settings object/policy container=]`.

For context: we need a clone at other points in order to make all CSP checks with a consistent snapshot of the policies. Since things are synchronous here, it should not be needed.

index.bs
```diff
@@ -703,6 +703,17 @@ agent MUST run the following steps:
...
+   [=environment settings object/policy container=], [=request/destination=] is an empty string,
+   and [=request/origin=] is |client|'s [=environment settings object/origin=].
+1. If [=should request be blocked by Content Security Policy?=] with |request| returns
```
martinthomson, line 711: How do I follow this link?  I want to review that process to see if things match.

annevk: @martinthomson in OP there's a link called "Preview" and a link called "Diff". They will have the links.

jan-ivar (approved):

PR #375, opened 2021-11-18
Elaborate requirements on custom certificate hashes

### LINKED ISSUE #349: serverCertificateFingerprints security properties
As discussed offline, it would be good to address the following points in the spec:

 - The use of `serverCertfificateFingerprints` effectively downgrades the security properties of the resulting transport
    - It is not as strong as regular HTTPS
    - The short expiration time does mitigate the risk for certificate compromise to a certain extent
    - This would likely sit well in the [privacy and security considerations section](https://w3c.github.io/webtransport/#privacy-security)
 - User agents should be allowed to reject pinned certificates due to ~arbitrary policy
    - This would allow revoking certificates, or applying strict checks to the algorithms/key sizes of pinned certificates
    - Maybe this would be best injected in the [custom certificate requirements section](https://w3c.github.io/webtransport/#custom-certificate-requirements)

vasilvv:
This adds extra requirements on custom certificates used by
WebTransportHash API, as well as adds a discussion concerning the
security properties of those into Security Considerations section.

Fixes #349.

***
<a href="https://pr-preview.s3.amazonaws.com/vasilvv/web-transport/pull/375.html" title="Last updated on Dec 9, 2021, 10:50 PM UTC (e35741a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/375/5b0a9ec...vasilvv:e35741a.html" title="Last updated on Dec 9, 2021, 10:50 PM UTC (e35741a)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/vasilvv/web-transport/pull/375.html" title="Last updated on Feb 4, 2022, 6:14 PM UTC (5d6b3d2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/375/5b0a9ec...vasilvv:5d6b3d2.html" title="Last updated on Feb 4, 2022, 6:14 PM UTC (5d6b3d2)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1794,7 +1794,7 @@ Normally, a user agent authenticates a TLS connection between itself and a
 remote endpoint by verifying the validity of the TLS server certificate
 provided against the server name in the URL.  This is accomplished by chaining
 server certificates to one of the trust anchors maintained by the browser; the
-trust anchors in question are responisble for authenticating the server names
+trust anchors in question are responsible for authenticating the server names
 in the certificates.  We will refer to this system as Web PKI.
 
 This API provides web applications with a capability to connect to a remote
@@ -1821,19 +1821,25 @@ certificate used is ephemeral, such a mechanism is not necessary.  In other
 cases, the Web application has to consider the mechanism by which the
 certificate hashes are provisioned; for instance, if the hash is provided as a
 cached HTTP resource, the cache needs to be invalidated if the corresponding
-certificate has been rotated due to compromise.
+certificate has been rotated due to compromise.  Another security feature
+provided by the Web PKI are safeguards against certain issues with key
+generation, such as rejecting certificates with well-known weak keys; while
+this specification does not provide any specific guidance, browsers MAY reject
+those as a part of implementation-defined behavior.
 
 Web PKI enforces an expiry period requirement on the certificates.  This
 requirement limits the scope of potential key compromise; it also forces server
 operators to design systems that support and actively perform key rotation.
 For this reason, WebTransport imposes a similar expiry requirement; as the
 certificates are expected to be ephemeral or short-lived, the expiry period is
-limited to two weeks.
+limited to two weeks.  The two weeks limit is a balance between setting the
+expiry limit as low as possible to minimize consequences of a key compromise,
+and maintaining it sufficiently high to lower the operational costs of
+provisioning certificates.
 
-The WebTransport API
-lets the application specify multiple certificate hashes at once, allowing
-the client to accept multiple certificates for a period in which a new
-certificate is being rolled out.
+The WebTransport API lets the application specify multiple certificate
+hashes at once, allowing the client to accept multiple certificates for a
+period in which a new certificate is being rolled out.
 
 Unlike a similar mechanism in WebRTC, the server certificate hash API in
 WebTransport does not provide any means of authenticating the client; the fact
--- a/index.bs
+++ b/index.bs
@@ -1101,7 +1101,7 @@ period MUST NOT exceed two weeks.  The user agent MAY impose additional
 
 The exact list of <dfn>allowed public key algorithms</dfn> used in the Subject
 Public Key Info field (and, as a consequence, in the TLS CertificateVerify
-message) is [=implementation-defined=]; however, it SHOULD include ECDSA with
+message) is [=implementation-defined=]; however, it MUST include ECDSA with the
 secp256r1 (NIST P-256) named group ([[!RFC3279]], Section 2.3.5; [[!RFC8422]])
 to provide an interoperable default.  It MUST NOT contain RSA keys
 ([[!RFC3279]], Section 2.3.1).
@@ -1828,15 +1828,15 @@ requirement limits the scope of potential key compromise; it also forces server
 operators to design systems that support and actively perform key rotation.
 For this reason, WebTransport imposes a similar expiry requirement; as the
 certificates are expected to be ephemeral or short-lived, the expiry period is
-limited to two weeks.  This also improves the cryptographic agility of the
-system, as the keys in the Subject Public Key Info field can be upgraded to a
-different algorithm as a part of the key rotation process.  WebTransport API
+limited to two weeks.
+
+The WebTransport API
 lets the application to specify multiple certificate hashes at once, allowing
-the client to accept multiple certificates for a perioid in which a new
+the client to accept multiple certificates for a period in which a new
 certificate is being rolled out.
 
 Unlike a similar mechanism in WebRTC, the server certificate hash API in
-WebTransport does not provide any form of authenticating the client; the fact
+WebTransport does not provide any means of authenticating the client; the fact
 that the client knows what the server certificate is or how to contact it is
 not sufficient.  The application has to establish the identity of the client
 in-band if necessary.
--- a/index.bs
+++ b/index.bs
@@ -1817,7 +1817,7 @@ algorithm-hash pairs to be specified.
 It is important to note that Web PKI provides additional security
 mechanisms in addition to simply establishing a chain of trust for a server
 name.  One of them is handling certificate revocation.  In cases where the
-certificate used is ephemeral, such mechanism is not necessary.  In other
+certificate used is ephemeral, such a mechanism is not necessary.  In other
 cases, the Web application has to consider the mechanism by which the
 certificate hashes are provisioned; for instance, if the hash is provided as a
 cached HTTP resource, the cache needs to be invalidated if the corresponding
@@ -1831,7 +1831,7 @@ certificates are expected to be ephemeral or short-lived, the expiry period is
 limited to two weeks.
 
 The WebTransport API
-lets the application to specify multiple certificate hashes at once, allowing
+lets the application specify multiple certificate hashes at once, allowing
 the client to accept multiple certificates for a period in which a new
 certificate is being rolled out.
 
--- a/index.bs
+++ b/index.bs
@@ -1795,14 +1795,14 @@ between itself and a remote endpoint by verifying the validity of the TLS
 server certificate provided against the server name in the URL; this approach
 will be further referred to as Web PKI.  The full details of how the Web PKI
 functions are implementation-defined, and go far beyond the scope of this
-docuement; in practice, those tend to be fairly complex, and involve
+document; in practice, those tend to be fairly complex, and involve
 procedures for maintaining roots of trust, handling revocations of previously
 issued certificates, as well as defending against key compromise, certificate
 mis-issuance, and numerous other security risks.
 
 This API provides web applications with a capability to connect to a remote
 network endpoint authenticated by the hash of its server certificate, rather
-than its server name.  This enables connections to the endpoints that are not
+than its server name.  This enables connections to endpoints that are not
 capable of getting long-term certificates, including hosts that are ephemeral
 in nature (e.g. short-lived virtual machines), or that are not publicly
 routable.  Since this mechanism substitutes Web PKI-based authentication for an
--- a/index.bs
+++ b/index.bs
@@ -1823,7 +1823,7 @@ certificate hashes are provisioned; for instance, if the hash is provided as a
 cached HTTP resource, the cache needs to be invalidated if the corresponding
 certificate has been rotated due to compromise.  Another security feature
 provided by the Web PKI are safeguards against certain issues with key
-generation, such as rejecting certificates with well-known weak keys; while
+generation, such as rejecting certificates with known weak keys; while
 this specification does not provide any specific guidance, browsers MAY reject
 those as a part of implementation-defined behavior.
 
--- a/index.bs
+++ b/index.bs
@@ -1790,21 +1790,21 @@ willing to accept connections from the Web.
 
 ## Authentication using Certificate Hashes {#certificate-hashes}
 
-In the vast majority of cases, a user agent authenticates a TLS connection
-between itself and a remote endpoint by verifying the validity of the TLS
-server certificate provided against the server name in the URL.  This is
-accomplished by chaining server certificates to one of the trust anchors
-maintained by the browser; the trust anchors in question are responisble for
-authenticating the server names in the certificates.  We will refer to this
-system as Web PKI.
+Normally, a user agent authenticates a TLS connection between itself and a
+remote endpoint by verifying the validity of the TLS server certificate
+provided against the server name in the URL.  This is accomplished by chaining
+server certificates to one of the trust anchors maintained by the browser; the
+trust anchors in question are responisble for authenticating the server names
+in the certificates.  We will refer to this system as Web PKI.
 
 This API provides web applications with a capability to connect to a remote
 network endpoint authenticated by a specific server certificate, rather than
-its server name.  This mechanism enables connections to endpoints that are not
-capable of getting long-term certificates, including hosts that are ephemeral
-in nature (e.g. short-lived virtual machines), or that are not publicly
-routable.  Since this mechanism substitutes Web PKI-based authentication for an
-individual connection, we need to compare the security properties of both.
+its server name.  This mechanism enables connections to endpoints for which
+getting long-term certificates can be challenging, including hosts that are
+ephemeral in nature (e.g. short-lived virtual machines), or that are not
+publicly routable.  Since this mechanism substitutes Web PKI-based
+authentication for an individual connection, we need to compare the security
+properties of both.
 
 A remote server will be able to successfully perform a TLS handshake only if it
 posesses the private key corresponding to the public key of the certificate
--- a/index.bs
+++ b/index.bs
@@ -1792,10 +1792,10 @@ willing to accept connections from the Web.
 
 Normally, a user agent authenticates a TLS connection between itself and a
 remote endpoint by verifying the validity of the TLS server certificate
-provided against the server name in the URL.  This is accomplished by chaining
-server certificates to one of the trust anchors maintained by the user agent; the
-trust anchors in question are responsible for authenticating the server names
-in the certificates.  We will refer to this system as Web PKI.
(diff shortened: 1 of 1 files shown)


vasilvv:
cc @letitz @estark37 @sleevi

index.bs
```diff
@@ -1784,6 +1788,57 @@ endpoint is verified to be a WebTransport endpoint; thus, the Web application
 cannot distinguish between a non-existing endpoint and the endpoint that is not
 willing to accept connections from the Web.
 
+## Authentication using Certificate Hashes {#certificate-hashes}
+
+In the vast majority of cases, a user agent authenticates a TLS connection
+between itself and a remote endpoint by verifying the validity of the TLS
+server certificate provided against the server name in the URL; this approach
```
martinthomson, line 1795: I think that you are missing a key component in this definition: the certification path construction to trust anchors defined by the CA/Browser Forum.  It's not that the certificate is valid (that's almost trivial) and contains the right name (actually trivial), but that the certificate chains to a trust anchor.

index.bs
```diff
@@ -1784,6 +1788,57 @@ endpoint is verified to be a WebTransport endpoint; thus, the Web application
...
+will be further referred to as Web PKI.  The full details of how the Web PKI
+functions are implementation-defined, and go far beyond the scope of this
+docuement; in practice, those tend to be fairly complex, and involve
```
martinthomson, line 1798:
```suggestion
document; in practice, those tend to be fairly complex, and involve
```

index.bs
```diff
@@ -1784,6 +1788,57 @@ endpoint is verified to be a WebTransport endpoint; thus, the Web application
...
+procedures for maintaining roots of trust, handling revocations of previously
+issued certificates, as well as defending against key compromise, certificate
+mis-issuance, and numerous other security risks.
```
martinthomson, line 1801: I don't think that you need to elaborate on why Web PKI is big, important, and complex.  The choice of words here is not sufficiently neutral: "far beyond", "fairly complex", ...

You don't have to draw a comparison. You only have to define the fingerprint-based scheme in terms of how it works.  To that end, most of this paragraph can go.

index.bs
```diff
@@ -1784,6 +1788,57 @@ endpoint is verified to be a WebTransport endpoint; thus, the Web application
...
+
+This API provides web applications with a capability to connect to a remote
+network endpoint authenticated by the hash of its server certificate, rather
+than its server name.  This enables connections to the endpoints that are not
```
martinthomson, line 1805:
```suggestion
than its server name.  This enables connections to endpoints that are not
```

index.bs
```diff
@@ -1784,6 +1788,57 @@ endpoint is verified to be a WebTransport endpoint; thus, the Web application
...
+server certificate provided against the server name in the URL; this approach
+will be further referred to as Web PKI.  The full details of how the Web PKI
+functions are implementation-defined, and go far beyond the scope of this
+docuement; in practice, those tend to be fairly complex, and involve
+procedures for maintaining roots of trust, handling revocations of previously
+issued certificates, as well as defending against key compromise, certificate
+mis-issuance, and numerous other security risks.
+
+This API provides web applications with a capability to connect to a remote
+network endpoint authenticated by the hash of its server certificate, rather
```
martinthomson, line 1804: At this level, you want to talk about the connection being authenticated toward a *specific* certificate.  The hash is a method of achieving that identification.

index.bs
```diff
@@ -1784,6 +1788,57 @@ endpoint is verified to be a WebTransport endpoint; thus, the Web application
...
+than its server name.  This enables connections to the endpoints that are not
+capable of getting long-term certificates, including hosts that are ephemeral
+in nature (e.g. short-lived virtual machines), or that are not publicly
+routable.  Since this mechanism substitutes Web PKI-based authentication for an
+individual connection, we need to compare the security properties of both.
+
+Validating against a specific hash has certain desirable security properties
+that Web PKI normally does not provide.  For instance, the hashes used for
```
martinthomson, line 1812: Again, this is unnecessarily biased.

index.bs
```diff
@@ -1784,6 +1788,57 @@ endpoint is verified to be a WebTransport endpoint; thus, the Web application
...
+verification are scoped to the context of a single connection, which can limit
+the consequences of any individual certificate being compromised.  In addition,
+hash-based verification has a single unambiguous root of trust (the hash
+provided by the Web application), while the Web PKI allows names to be
+authenticated by multiple certificate authorities;  the latter fact arises from
+a variety of operational and ecosystem concerns that are not usually
+relevant for the use cases where hash-based verification is used.
```
martinthomson, line 1819: "not usually relevant" seems like a value judgment.

martinthomson (changes requested):
I don't think that this is what we want for this section.

My suggestion is to concentrate on the properties of the system that identifies specific certificates.  Acknowledge that this system is deliberately rigid in the sense that there is precisely one way to satisfy the requirement: control the private key that is associated with the public key in the certificate.  

Point out that the reason this is allowed is to make it possible to build services that might find it challenging to use the Web PKI.  Short-lived virtual machines, peer-to-peer communication, hosts without public names, etc...

Observe that this depends on the hash function (SHA-256) having good second preimage resistance and the secure delivery of the identifier to the browser.

Note the inherent shortcomings of this system.  A weak key might be chosen that nevertheless is accepted.  Duplicate keys might be used.  The choice of certificate is made by the site, so a failure in the delivery of script for the site can be exploited by an attacker to perform impersonation attacks. Moreover, there is no auditing system that might allow parties external to the communication observe these failures.  

Finally, there is the lack of revocation, the remedy for which is enforcement of short validity period.

I might also observe that the system does not provide a way to authenticate the client (unlike WebRTC) which exposes services to attacks if they do not use application-level authentication and assume that knowledge of contact points is sufficient to authorize access.  That's a new observation that is probably worth a new issue and discussion.

That is don't write this as a comparison to the Web PKI, but base it on the properties of the system that is being defined.

index.bs
vasilvv: Done.

index.bs
vasilvv: Removed the paragraph.

index.bs
vasilvv: Removed the paragraph.

index.bs
vasilvv: Deleted that part.

index.bs
vasilvv: From my perspective, the role of WebPKI as far as immediate network logic is concerned is establishing a validity of a `(server identity, public key)` pair; to some extent, the fact that we have trust anchors at all is an implementation detail, as the key observation here is that fingerprint-based verifier sidesteps this problem by making the first and the second elements of this pair essentially the same thing.

vasilvv:
I added some paragraphs on security properties, and removed a lot of direct comparison to the Web PKI.  Please take another look.

martinthomson (commented):
Much better, thanks.

index.bs
```diff
@@ -1784,6 +1788,59 @@ endpoint is verified to be a WebTransport endpoint; thus, the Web application
...
+maintained by the browser; the trust anchors in question are responisble for
+authenticating the server names in the certificates.  We will refer to this
+system as Web PKI.
+
+This API provides web applications with a capability to connect to a remote
+network endpoint authenticated by a specific server certificate, rather than
+its server name.  This mechanism enables connections to endpoints that are not
+capable of getting long-term certificates, including hosts that are ephemeral
+in nature (e.g. short-lived virtual machines), or that are not publicly
+routable.  Since this mechanism substitutes Web PKI-based authentication for an
```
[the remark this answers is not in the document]
vasilvv: Good point.  Reworded.

index.bs
```diff
@@ -1784,6 +1788,59 @@ endpoint is verified to be a WebTransport endpoint; thus, the Web application
 cannot distinguish between a non-existing endpoint and the endpoint that is not
 willing to accept connections from the Web.
 
+## Authentication using Certificate Hashes {#certificate-hashes}
+
+In the vast majority of cases, a user agent authenticates a TLS connection
```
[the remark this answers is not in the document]
vasilvv: I would lean on a reference if there were a good reference to lean on (I am not sure RFC 6125 quite captures what I'm trying to say here).

Fair point on "vast majority", though.  Reworded.

vasilvv:
cc @davidben

index.bs
```diff
@@ -1784,6 +1788,59 @@ endpoint is verified to be a WebTransport endpoint; thus, the Web application
 cannot distinguish between a non-existing endpoint and the endpoint that is not
 willing to accept connections from the Web.
 
+## Authentication using Certificate Hashes {#certificate-hashes}
+
+Normally, a user agent authenticates a TLS connection between itself and a
+remote endpoint by verifying the validity of the TLS server certificate
+provided against the server name in the URL.  This is accomplished by chaining
+server certificates to one of the trust anchors maintained by the browser; the
+trust anchors in question are responisble for authenticating the server names
```
davidben, line 1797: s/responisble/responsible/

index.bs
```diff
@@ -1784,6 +1788,59 @@ endpoint is verified to be a WebTransport endpoint; thus, the Web application
...
+certificate hashes are provisioned; for instance, if the hash is provided as a
+cached HTTP resource, the cache needs to be invalidated if the corresponding
+certificate has been rotated due to compromise.
+
+Web PKI enforces an expiry period requirement on the certificates.  This
+requirement limits the scope of potential key compromise; it also forces server
+operators to design systems that support and actively perform key rotation.
+For this reason, WebTransport imposes a similar expiry requirement; as the
+certificates are expected to be ephemeral or short-lived, the expiry period is
+limited to two weeks.
```
davidben, line 1831: Maybe connect this to the revocation thing above; this also bounds the consequences of cache mishaps from the previous paragraph.

davidben (approved):

index.bs
```diff
@@ -1092,15 +1092,19 @@ To <dfn>verify a certificate hash</dfn>, do the following:
...
-
-Issue: Reconsider the time period above.  We want it to be sufficiently large
-that applications using this for ephemeral certificates can do so without
-having to fight the clock skew, but small enough to discourage long-term use
-without key rotation.
+MUST be an X.509v3 certificate as defined in [[!RFC5280]], the key used in the
+Subject Public Key field MUST be one of the [=allowed public key algorithms=],
+the current time MUST be within the validity period of the certificate as
+defined in Section 4.1.2.5 of [[!RFC5280]] and the total length of the validity
+period MUST NOT exceed two weeks.  The user agent MAY impose additional
```
mikewest, line 1099: Does the time period need to be specified in a way that's robust against leap days/seconds/etc? Perhaps specifying 14 days, and referring to [6.3.2 of the Baseline Requirements](https://cabforum.org/wp-content/uploads/CA-Browser-Forum-BR-1.8.0.pdf#page=70)' definition of a day, and recommendations for issuance?

index.bs
```diff
@@ -1784,6 +1788,59 @@ endpoint is verified to be a WebTransport endpoint; thus, the Web application
...
+certificate has been rotated due to compromise.
+
+Web PKI enforces an expiry period requirement on the certificates.  This
+requirement limits the scope of potential key compromise; it also forces server
+operators to design systems that support and actively perform key rotation.
+For this reason, WebTransport imposes a similar expiry requirement; as the
+certificates are expected to be ephemeral or short-lived, the expiry period is
+limited to two weeks.
+
+The WebTransport API
```
mikewest, line 1833: Nit: Somewhat weird linebreak here.

index.bs
```diff
@@ -1784,6 +1788,59 @@ endpoint is verified to be a WebTransport endpoint; thus, the Web application
...
+lets the application to specify multiple certificate hashes at once, allowing
```
mikewest, line 1834:
```suggestion
lets the application specify multiple certificate hashes at once, allowing
```
```suggestion
lets the application to specify multiple certificate hashes at once, allowing
```

index.bs
mikewest: Also: two weeks is probably a reasonable compromise, but it would be helpful to understand how you landed on it. The discussion in the section below is light on detail...

index.bs
```diff
@@ -1784,6 +1788,59 @@ endpoint is verified to be a WebTransport endpoint; thus, the Web application
...
+posesses the private key corresponding to the public key of the certificate
+specified.  The API identifies the certificates using their hashes.  That is
+only secure as long as the cryptographic hash function used has second-preimage
+resistance.  The only function defined in this document is SHA-256; the API
+provides a way to introduce new hash functions through allowing multiple
+algorithm-hash pairs to be specified.
+
+It is important to note that Web PKI provides additional security
+mechanisms in addition to simply establishing a chain of trust for a server
+name.  One of them is handling certificate revocation.  In cases where the
```
letitz, line 1819: Another is that CAs enforce minimum standards of security on keys - checking for debian weak keys, for example. They also guard against key sharing, I believe, which is somewhat helped by the short expiration time. We should mention this.

index.bs
letitz: +1.

index.bs
```diff
@@ -1784,6 +1788,59 @@ endpoint is verified to be a WebTransport endpoint; thus, the Web application
...
+certificate used is ephemeral, such mechanism is not necessary.  In other
```
letitz, line 1820:
```suggestion
certificate used is ephemeral, such a mechanism is not necessary.  In other
```

letitz (changes requested):
Thanks for working on this, and apologies for the silence. Left a couple comments below, looking good otherwise!

index.bs
davidben: I believe two weeks, 14 days, and 14 * 24 * 3600 seconds are all equivalent. None of those measures are affected by leap days. Leap seconds are horrible, but all timestamps in certificates (and the web platform, and most of software in general) uses UTC, not TAI. That means leap seconds aren't counted in these sorts of measures. They just silently repeat/smear relative to physical time.

index.bs
vasilvv: Added a mention of weak keys.

index.bs
vasilvv: Added a comment on the rationale behind two weeks.

letitz (approved):
Looks good to me! Thanks for working on this :)

jan-ivar:
Meeting:
- Looks ready to merge. @martinthomson can you take a look to make sure this looks good to you?

index.bs
```diff
@@ -1790,21 +1790,21 @@ willing to accept connections from the Web.
...
-between itself and a remote endpoint by verifying the validity of the TLS
-server certificate provided against the server name in the URL.  This is
-accomplished by chaining server certificates to one of the trust anchors
-maintained by the browser; the trust anchors in question are responisble for
-authenticating the server names in the certificates.  We will refer to this
-system as Web PKI.
+Normally, a user agent authenticates a TLS connection between itself and a
+remote endpoint by verifying the validity of the TLS server certificate
+provided against the server name in the URL.  This is accomplished by chaining
+server certificates to one of the trust anchors maintained by the browser; the
```
martinthomson, line 1796:
```suggestion
server certificates to one of the trust anchors maintained by the user agent; the
```

index.bs
```diff
@@ -1790,21 +1790,21 @@ willing to accept connections from the Web.
...
+trust anchors in question are responsible for authenticating the server names
+in the certificates.  We will refer to this system as Web PKI.
```
martinthomson, line 1798: cite RFC 6125 here.

index.bs
```diff
@@ -1817,26 +1817,32 @@ algorithm-hash pairs to be specified.
...
 name.  One of them is handling certificate revocation.  In cases where the
-certificate used is ephemeral, such mechanism is not necessary.  In other
+certificate used is ephemeral, such a mechanism is not necessary.  In other
 cases, the Web application has to consider the mechanism by which the
 certificate hashes are provisioned; for instance, if the hash is provided as a
 cached HTTP resource, the cache needs to be invalidated if the corresponding
-certificate has been rotated due to compromise.
+certificate has been rotated due to compromise.  Another security feature
+provided by the Web PKI are safeguards against certain issues with key
+generation, such as rejecting certificates with well-known weak keys; while
```
martinthomson, line 1826:
```suggestion
generation, such as rejecting certificates with known weak keys; while
```

index.bs
```diff
@@ -1817,26 +1817,32 @@ algorithm-hash pairs to be specified.
...
-limited to two weeks.  This also improves the cryptographic agility of the
-system, as the keys in the Subject Public Key Info field can be upgraded to a
-different algorithm as a part of the key rotation process.  WebTransport API
-lets the application to specify multiple certificate hashes at once, allowing
-the client to accept multiple certificates for a perioid in which a new
-certificate is being rolled out.
+limited to two weeks.  The two weeks limit is a balance between setting the
+expiry limit as low as possible to minimize consequences of a key compromise,
+and maintaining it sufficiently high to lower the operational costs of
+provisioning certificates.
```
martinthomson, line 1838: This reasoning seems a little odd.  Part of the point of using hashes is so that you don't need to provision certificates in the usual fashion and this might read as saying that.

What I think you are talking about is the potential costs of synchronizing configuration across the resources that depend on the service (and so have certificate hashes embedded) and the systems that terminate connections.  Having to synchronize updates of these systems is made easier if the validity period is longer.

You also don't mention the problem of client clocks being horrible out of sync, which is relevant here.

martinthomson (commented):
WFM

index.bs
vasilvv: Done.

index.bs
vasilvv: "Synchronization" is probably more appropriate here than "provisioning", indeed.  Reworded and added a mention of clock sync.

jan-ivar:
@martinthomson goes this LGTY now?

martinthomson (approved):
LGTM

Issue #229, opened 2021-03-18
Port blocking


ricea:
We need to ensure that port blocking is applied to avoid security issues like the [NAT Slipstreaming attack](https://samy.pl/slipstream/). This will probably be delegated to the Fetch standard, so this issue is just here to track it and make sure it isn't forgotten.

yutakahirano:
I'm sure that WebTransport people are happy with the current block list, but having an allow-list instead (whatwg/fetch#1189) can be concerning.

annevk:
https://github.com/whatwg/fetch/issues/889 has some thoughts on this. We probably need to distinguish between local and remote ports as well.

jan-ivar:
@yutakahirano What's our "current block list"? I couldn't find any reference to "port" or "ports" in this spec.

ricea:
https://fetch.spec.whatwg.org/#port-blocking is what we're blocking in practice.

jan-ivar:
We probably should call that out in the spec explicitly.

jan-ivar:
Meeting:
- Adam: Probably W3C's responsibility to track this, since IETF may have native implementations where port blocking is not an issue.
- Marking ready for PR to add link to fetch spec per https://github.com/w3c/webtransport/issues/229#issuecomment-972584267
- As follow-up, push on https://github.com/whatwg/fetch/issues/1268 to possibly add a list to source ports as well to the block list that would be shared with WebTransport (more important for UDP/HTTP3 that we control the source port)
- Might be worth sending a mail to IETF (WebTrans & QUIC?) mailing lists requesting clarity on where this belongs?

PR #378, opened 2021-12-06
Block ports on fetch's bad ports list & use WebTransportError for it and CSP.

jan-ivar:
Fixes https://github.com/w3c/webtransport/issues/229.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webtransport/pull/378.html" title="Last updated on Dec 7, 2021, 4:44 PM UTC (1bdb11e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/378/3ddc1a8...jan-ivar:1bdb11e.html" title="Last updated on Dec 7, 2021, 4:44 PM UTC (1bdb11e)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -715,7 +715,8 @@ To <dfn>initialize WebTransport over HTTP</dfn>, given a {{WebTransport}} object
    and [=request/origin=] is |origin|.
 1. Run <a>report Content Security Policy violations for |request|</a>.
 1. If [=should request be blocked by Content Security Policy?=] with |request| returns
-   <b>blocked</b>, then abort the remaining steps and [=queue a network task=] with |transport|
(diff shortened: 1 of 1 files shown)


jan-ivar:
cc @annevk

yutakahirano:
Should we use `SecurityError` here?

annevk:
Why does CSP not use `WebTransportError`? I think ideally this uses the same signal as obtaining a connection returning failure. In the future we might want to move the port check there. And potentially we could move some of the CSP checking there as well, at least in theory.

annevk (approved):

yutakahirano:
@jan-ivar, can you update the test as well?

Issue #107, opened 2020-03-26
Fallback support


aboba:
If QUIC is not available what happens? Does the WebTransport constructor return an error?  That's what happens in the current implementation. Or can a fallback transport be invoked?  What if some functionality is not available (e.g. no datagram support)? 

In general, the specification needs to be more clear about fallback (e.g. whether it is transparent or up to the application to handle).

Related: Issue https://github.com/w3c/webtransport/issues/164

ekinnear:
Agreed that it needs to be more clear about fallback. I think we'd discussed the desire to have a fallback with the understanding that QUIC is not going to be available on all networks. That aligns well with work in other areas where we've seen that fallback to TCP-based protocols is workable, if not as nice as using QUIC -- and is certainly better than not having any connectivity for _most_ use cases.

jan-ivar:
Based in the most recent IETF meeting, it sounds like fallback is coming back into favor. If fallback is left to web developers, the concern is a large number of them will settle on the lowest common denominator, e.g. some websocket-like calback-based API in a js library.

At the same time, apps would need to learn what they got. This might look like something like this:
```js
await transport.ready;
if (transport.mode = "http2") 
```

aboba:
At IETF 110, we also discussed an optional argument in the constructor to indicate "QUIC only", so if the application would prefer to fail rather than fallback to HTTP/2, it can request that.

vasilvv:
I'm not sure if checking a specific transport is a good idea, since that's not future proof.  We should instead allow querying for properties, e.g. "are datagrams unreliable".

jan-ivar:
Agree. Also, pardon my ignorance, but when do we know QUIC is not available? E.g. should this be a static method, or does it depend on the server?

aboba:
At the May 20, 2021 IETF WEBTRANS WG meeting, there was support for WebTransport over HTTP/2.  Meeting minutes, slides, etc. are [here](https://datatracker.ietf.org/meeting/interim-2021-webtrans-02/session/webtrans).

yutakahirano:
I think this is out of Minimum Viable Product.

jan-ivar:
Meeting:
- Enums are good. But location should be higher. Maybe `wt.reliability` = new | reliable | unreliable ?
- If you don't want it you write `if (wt.reliability != "unreliable") wt.close()` ? Too costly?
  - new hostname, you attempt toconnect using QUIC, host doesn't support or can't detect, it fails, then you fall back to H2.
[...]

PR #385, opened 2022-02-01
Add wt.reliability attr and {requireUnreliable} constructor option.

jan-ivar:
Fixes https://github.com/w3c/webtransport/issues/107.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webtransport/pull/385.html" title="Last updated on Mar 31, 2022, 3:48 PM UTC (73518d8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/385/e3f77e5...jan-ivar:73518d8.html" title="Last updated on Mar 31, 2022, 3:48 PM UTC (73518d8)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -599,7 +599,7 @@ A {{WebTransport}} object has the following internal slots.
    failed.
   </tr>
   <tr>
-   <td><dfn>\`[[Reliability]]`</dfn>
+   <td><dfn>`[[Reliability]]`</dfn>
    <td class="non-normative">A {{WebTransportReliabilityMode}} indicating whether
    unreliable (UDP) transport is supported or whether only reliable (TCP fallback)
    transport is used. Returns `"pending"` until a connection has been established.
--- a/index.bs
+++ b/index.bs
@@ -2006,7 +2006,7 @@ connect.onclick = async () => {
     wt = new WebTransport(url);
     addToEventLog('Initiating connection...');
     await wt.ready;
-    addToEventLog(\`HTTP/${(wt.reliability == "reliable-only")? 2 : 3} connection ready.\`);
+    addToEventLog(\`${(wt.reliability == "reliable-only")? "TCP" : "UDP"} connection ready.\`);
     wt.closed
       .then(() => addToEventLog('Connection closed normally.'))
       .catch(() => addToEventLog('Connection closed abruptly.', 'error'));
--- a/index.bs
+++ b/index.bs
@@ -670,9 +670,8 @@ agent MUST run the following steps:
    {{WebTransportOptions/serverCertificateHashes}} if it exists, and null otherwise.
 1. If |dedicated| is false and |serverCertificateHashes| is non-null, then [=throw=] a
    {{TypeError}}.
-1. Let |disallowFallback| be {{WebTransport/constructor(url, options)/options}}'s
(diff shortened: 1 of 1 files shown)


martinthomson:
I think that I would have framed this as `tcpFallback: true/false` rather than talking about a "datagram mode".  Or, to phrase it more positively "quicTransport: true/false" or "http3: true/false".

index.bs
```diff
@@ -1043,6 +1067,7 @@ dictionary WebTransportHash {
 
 dictionary WebTransportOptions {
   boolean allowPooling;
+  boolean disallowFallback;
```
jan-ivar, line 1070: From @martinthomson 
```suggestion
  boolean requireUnreliable;
```

index.bs
```diff
@@ -582,6 +583,11 @@ interface WebTransport {
   readonly attribute ReadableStream incomingUnidirectionalStreams;
 };
 
+enum WebTransportReliabilityMode {
+  "new",
```
jan-ivar, line 587:
```suggestion
  "pending",
```

index.bs
jan-ivar: Meeting:
- Depends on whether we want fallback to be the default or not.
- Early adopters can specify `requireUnreliable: true` now to avoid fallback in the future.
- Should get this in early so people won't be surprised by this behavior.
- Different from allowPooling... open new issue to discuss allowPooling default?

vasilvv:
I don't like the way this describes transports as "reliable" vs "unreliable".  WebTransport always supports reliable delivery.  It sometimes supports unreliable delivery.  So the options should be phrased as "reliableOnly" vs "supportsUnreliable".

index.bs
```diff
@@ -1043,6 +1066,7 @@ dictionary WebTransportHash {
 
 dictionary WebTransportOptions {
   boolean allowPooling;
```
jan-ivar, line 1068:
```suggestion
  boolean allowPooling = false;
```

index.bs
```diff
@@ -1978,8 +2006,7 @@ connect.onclick = async () => {
     wt = new WebTransport(url);
     addToEventLog('Initiating connection...');
     await wt.ready;
-    addToEventLog('Connection ready.');
-
+    addToEventLog(\`HTTP/${(wt.reliability == "reliable-only")? 2 : 3} connection ready.\`);
```
jan-ivar, line 2009: From meeting: avoid referring to specific HTTP/N. 
```suggestion
    addToEventLog(\`${(wt.reliability == "reliable-only")? "TCP" : "UDP"} connection ready.\`);
```

index.bs
```diff
@@ -1978,8 +2006,7 @@ connect.onclick = async () => {
     wt = new WebTransport(url);
     addToEventLog('Initiating connection...');
     await wt.ready;
-    addToEventLog('Connection ready.');
-
+    addToEventLog(\`${(wt.reliability == "reliable-only")? "TCP" : "UDP"} connection ready.\`);
```
jan-ivar, line 2009:
```suggestion
    addToEventLog(\`${(wt.reliability == "reliable-only")? "TCP" : "UDP"} \` +
                  \`connection ready.\`);
```

Issue #388, opened 2022-02-16
Datagram priority algorithm fails to account for chunk size of reliable streams


jan-ivar:
From https://github.com/w3c/webtransport/issues/62#issuecomment-1040952103:
> Mathis Engelbart has looked at impact of prioritizing datagrams within a QUIC connection operating with NewReno. This results in high datagram latency, since **the reliable stream fills the congestion window when datagrams aren't being sent.**
Results are [here](https://docs.google.com/presentation/d/16gxsLckg7mLzYhIiQm0wPgUJHXNPfivbTmln-hhxlJw/edit#slide=id.g1143eb38147_6_224).

A flaw in SendStream's [write](https://w3c.github.io/webtransport/#sendstream-write)'s primitive priority gating algorithm:
 - 7. Wait for _transport_’s [[[Datagrams]]](https://w3c.github.io/webtransport/#webtransport-datagrams)'s [[[OutgoingDatagramsQueue]]](https://w3c.github.io/webtransport/#datagramduplexstream-outgoingdatagramsqueue) to be empty
 
 ...is we fail to account for chunk size, which, being app-controlled, may be magnitudes larger than a ~1024 byte datagram. So a single opening in the datagram queue may kick off:

 - 8. [Send](https://w3c.github.io/webtransport/#stream-send) [mega]_bytes_ on stream’s [[[InternalStream]]](https://w3c.github.io/webtransport/#sendstream-internalstream) and wait for the operation to complete.

...which may take a while, and presumably compete for priority with subsequently sent datagrams (if not outright blocking them).

Even if we tried to control size through re-chunking, there might be N reliable streams ready to jump at this one opportunity.

jan-ivar:
### Proposed solution

A better solution might be:
 - 8. [Send](https://w3c.github.io/webtransport/#stream-send) [mega]_bytes_ on stream’s [[[InternalStream]]](https://w3c.github.io/webtransport/#sendstream-internalstream) _with priority X_ and wait for the operation to complete.

...where X is something we come up with, since [Send (section 2.2)](https://datatracker.ietf.org/doc/html/draft-ietf-quic-transport#section-2.2) appears to expect and _"[relies on receiving priority information from the application](https://datatracker.ietf.org/doc/html/draft-ietf-quic-transport#section-2.2)"_, where _"application"_ is us.

At minimum X should be "high" for datagrams and "low" for everything else, to solve this issue.

This may have further utility in solving https://github.com/w3c/webtransport/issues/62.

[...]

PR #394, opened 2022-03-15
Reduce priority algorithm to normative guidance.

jan-ivar:
Fixes https://github.com/w3c/webtransport/issues/388.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1269,11 +1269,11 @@ To <dfn for="SendStream">write</dfn> |chunk| to a {{SendStream}} |stream|, run t
 1. Let |bytes| be a copy of the [=byte sequence=] which |chunk| represents.
 1. Set |stream|'s [=[[PendingOperation]]=] to |promise|.
 1. Return |promise| and run the remaining steps [=in parallel=].
-1. Wait for |transport|'s [=[[Datagrams]]=]'s [=[[OutgoingDatagramsQueue]]=] to be empty.
-
-   Note: This means outgoing datagrams are prioritized over stream data.
 1. [=stream/Send=] |bytes| on |stream|'s [=[[InternalStream]]=] and wait for the operation to
    complete.
+   This sending MAY happen in parallel with sending of previously queued streams and datagrams,
+   as well as streams and datagrams yet to be queued to be sent over this transport.
+   When determining prioritization, the user agent SHOULD make sure datagrams get out quickly.
 1. If the previous step failed, abort the remaining steps.
 
   Note: We don't reject |promise| here because we handle network errors elsewhere, and those steps


yutakahirano (approved):
LGTM.

index.bs
```diff
@@ -1269,11 +1269,11 @@ To <dfn for="SendStream">write</dfn> |chunk| to a {{SendStream}} |stream|, run t
 1. Let |bytes| be a copy of the [=byte sequence=] which |chunk| represents.
 1. Set |stream|'s [=[[PendingOperation]]=] to |promise|.
 1. Return |promise| and run the remaining steps [=in parallel=].
-1. Wait for |transport|'s [=[[Datagrams]]=]'s [=[[OutgoingDatagramsQueue]]=] to be empty.
-
-   Note: This means outgoing datagrams are prioritized over stream data.
 1. [=stream/Send=] |bytes| on |stream|'s [=[[InternalStream]]=] and wait for the operation to
    complete.
+   This sending MAY happen in parallel with sending of previously queued streams and datagrams,
```
LPardue, line 1274: what does the term in parallel mean here? For a single connection, QUIC packets will still be emitted in some serial order. An implementation might timeslice (interleave) stream and datagram payload over different packets, but that might not align with some people's definition of parallel.

jan-ivar, line 1274: Good point, I meant that this write algorithm may end up being invoked multiple times which means these steps may run in parallel with past and future calls of it. But it might be simpler to describe the net result.
```suggestion
   This sending MAY be interleaved with sending of previously queued streams and datagrams,
```

LPardue: interleaved works for me thanks!

martinthomson (approved):
w/ the tweak, seems OK

index.bs
```diff
@@ -1269,11 +1269,11 @@ To <dfn for="SendStream">write</dfn> |chunk| to a {{SendStream}} |stream|, run t
...
 1. Set |stream|'s [=[[PendingOperation]]=] to |promise|.
 1. Return |promise| and run the remaining steps [=in parallel=].
-1. Wait for |transport|'s [=[[Datagrams]]=]'s [=[[OutgoingDatagramsQueue]]=] to be empty.
-
-   Note: This means outgoing datagrams are prioritized over stream data.
 1. [=stream/Send=] |bytes| on |stream|'s [=[[InternalStream]]=] and wait for the operation to
    complete.
+   This sending MAY be interleaved with sending of previously queued streams and datagrams,
+   as well as streams and datagrams yet to be queued to be sent over this transport.
+   When determining prioritization, the user agent SHOULD make sure datagrams get out quickly.
```
jan-ivar, line 1276:
```suggestion
```

Issue #365, opened 2021-09-30
Pluggable Congestion Control


aboba:
QUIC congestion control algorithms such as BBRv1 have issues when WebTransport is used for realtime communications.  For example, there is no equivalent to the probing supported in WebRTC implementations, which allows for rapid rampup in encoder rate when congestion is resolved. 

The problem could be solved by allowing alternate congestion control mechanisms to be specified when the WebTransport is constructed.

LPardue:
> default congestion control for QUIC

Can you clarify this a bit? While some types of QUIC frames are subject to congestion control, there is no default algorithm in QUIC (a reno-inspired congestion control algorithm is provided as an example, it is not required to implement that algorithm and many choose something like cubic or BBR).

Is the request to exempt some type of frames (e.g. STREAM or DATAGRAM) from the transport-layer congestion controller?

aboba:
The [IETF RMCAT WG](https://datatracker.ietf.org/wg/rmcat/about/) developed congestion control algorithms optimized for use with realtime communications.  The algorithms developed were published as [experimental RFCs](https://datatracker.ietf.org/wg/rmcat/documents/). 

The proposal would be to allow selection of one of these algorithms (or an alternative, developed with a similar goal). 

In particular, the algorithms developed in RMCAT (or alternatives implemented in WebRTC.org source) seek to avoid building queues in bottleneck devices by incorporating delay metrics.  They also are capable of probing the network to determine when simulcast or SVC layers can be added back after congestive losses cause the layers to be dropped.

LPardue:
Thanks @aboba, that's helpful. Would this proposal require the browser to implement those CCs on behalf of the API user? If so, are there cases where applications above the API want to run their own CC  (which sounds flexible but also like a footgun).

aboba:
It is currently not possible for realtime communications applications to implement the kind of CC algorithms specified in RMCAT on their own. Firstly, the information necessary to do this is not available (WebTransport does not expose statistics, such as information available to the QUIC implementation.  Also, an application cannot implement multiplicative increase on top of a QUIC congestion control algorithm designed to implement Multiplicate Decrease/Additive Increase. 

Note that this problem has been understood from the start (e.g. it was raised in the original design of RTCQuicTransport), but has fallen through the cracks.

LPardue:
Makes sense! Would this work for WebTransport over HTTP/2?

aboba:
[...]

PR #406, opened 2022-07-05
Add congestionControl contructor arg and readonly attribute

jan-ivar:
Fixes https://github.com/w3c/webtransport/issues/365.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webtransport/pull/406.html" title="Last updated on Aug 30, 2022, 2:20 PM UTC (3f4b036)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/406/71ee094...jan-ivar:3f4b036.html" title="Last updated on Aug 30, 2022, 2:20 PM UTC (3f4b036)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -609,7 +609,8 @@ A {{WebTransport}} object has the following internal slots.
    <td><dfn>`[[CongestionControl]]`</dfn>
    <td class="non-normative">A {{WebTransportCongestionControl}} indicating
    whether a preference for a congestion control algorithm optimized for
-   throughput or low latency was requested by the application.
+   throughput or low latency was requested by the application and satisfied
+   by the user agent, or `"default"`.
   </tr>
   <tr>
    <td><dfn>`[[Closed]]`</dfn>
@@ -840,8 +841,10 @@ these steps.
    The getter steps are to return [=this=]'s {{[[Reliability]]}}.
 
 : <dfn for="WebTransport" attribute>congestionControl</dfn>
-:: The application's preference, if requested in the constructor, for a
-   congestion control algorithm optimized for either throughput or low latency.
+:: The application's preference, if requested in the constructor, and satisfied
+   by the user agent, for a congestion control algorithm optimized for either
+   throughput or low latency for sending on this connection. If a preference was
+   requested but not satisfied, then the value is `"default"`
    The getter steps are to return [=this=]'s {{[[CongestionControl]]}}.
 
 ## Methods ##  {#webtransport-methods}
--- a/index.bs
+++ b/index.bs
@@ -56,7 +56,6 @@ Boilerplate: omit conformance
 <pre class="link-defaults">
 spec:infra; type:dfn; for:set; text:for each
 spec:infra; type:dfn; for:/; text:set
-spec:infra; type:dfn; for:list; text:exist
 spec:streams; type:interface; text:ReadableStream
 spec:fetch; type:dfn; for:/; text:fetch
 spec:url; type:dfn; text:scheme
@@ -534,7 +533,6 @@ defined in [[!WEB-TRANSPORT-HTTP3]].
 interface WebTransport {
   constructor(USVString url, optional WebTransportOptions options = {});
 
-  static sequence&lt;WebTransportCongestionControlAlgorithm&gt; getCongestionControlAlgorithms();
   Promise&lt;WebTransportStats&gt; getStats();
   readonly attribute Promise&lt;undefined&gt; ready;
   readonly attribute WebTransportReliabilityMode reliability;
@@ -609,8 +607,9 @@ A {{WebTransport}} object has the following internal slots.
   </tr>
   <tr>
    <td><dfn>`[[CongestionControl]]`</dfn>
-   <td class="non-normative">A {{WebTransportCongestionControl}} describing
-   the congestion control algorithm used for sending on this connection.
+   <td class="non-normative">A {{WebTransportCongestionControl}} indicating
+   whether a preference for a congestion control algorithm optimized for
+   throughput or low latency was requested by the application.
   </tr>
   <tr>
    <td><dfn>`[[Closed]]`</dfn>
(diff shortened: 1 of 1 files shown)


jan-ivar:
Meeting:
- Are the abstract enum values sufficient to pick between multiple congestion control algorithms, or would it be preferable to accept a string with name(s)? NewReno, BBR, SCReAM etc.
- Might need a registry approach instead
- note: getter is usable for feature detection immediately, i.e. this isn't a negotiated property dependent on server
- might be good to have both abstract ones and specific names.
- modify PR to propose a static getter getAvailableAlgorithms.

LPardue:
> If the user agent only implements the default congestion control algorithm in
   [[!RFC9002]] [section 7](https://datatracker.ietf.org/doc/html/rfc9002#section-7),
   set |congestionControl| to `"throughput"`.

> This configuration option is considered at risk due to the lack of
       implementation and specification of any congestion control algorithm
       other than [[!RFC9002]] [section 7](https://datatracker.ietf.org/doc/html/rfc9002#section-7)
       at the time of writing.
       
I don't think these statements are quite accurate. The Reno-like CC in RFC 9002 is exemplary, IIUC  most implementations don't use it and prefer CUBIC - including Firefox and Chrome.

index.bs
```diff
@@ -638,6 +645,11 @@ agent MUST run the following steps:
    {{TypeError}}.
 1. Let |requireUnreliable| be {{WebTransport/constructor(url, options)/options}}'s
    {{WebTransportOptions/requireUnreliable}}.
+1. Let |congestionControl| be {{WebTransport/constructor(url, options)/options}}'s
+   {{WebTransportOptions/congestionControl}}.
+1. If the user agent only implements the default congestion control algorithm in
+   [[!RFC9002]] [section 7](https://datatracker.ietf.org/doc/html/rfc9002#section-7),
```
martinthomson, line 651: that's fairly narrowly targeted, so maybe "... or an algorithm that optimizes for throughput ..."

index.bs
```diff
@@ -1058,6 +1085,18 @@ that determine how WebTransport connection is established and used.
    use for normal [=fetch=] operations.
 :: This cannot be used with {{WebTransportOptions/allowPooling}}.
 
+: <dfn for="WebTransportOptions" dict-member>congestionControl</dfn>
+:: When set, influences the congestion control algorithm used by the user agent
+   when sending data over this connection.
+   <div class="issue atrisk">
```
martinthomson, line 1091: 👍

index.bs
```diff
@@ -1058,6 +1085,18 @@ that determine how WebTransport connection is established and used.
...
+     <p>
+       This configuration option is considered at risk due to the lack of
+       implementation and specification of any congestion control algorithm
+       other than [[!RFC9002]] [section 7](https://datatracker.ietf.org/doc/html/rfc9002#section-7)
+       at the time of writing.
```
martinthomson, line 1096: What Lucas said.  The challenge here is in standardizing an algorithm that optimizes for latency.  So maybe say that instead.

jan-ivar:
OK I've reworked the API based on meeting feedback, with some liberties. Example usage:

Read which CC a connection uses (readonly):
```js
const wt = new WebTransport(url);
console.log(wt.congestionControl.name); // CUBIC
console.log(wt.congestionControl.aim); // throughput
```

Read the CC algorithms the browser supports, and create a WebTransport with the first one:
```js
const [congestionControl] = WebTransport.getCongestionControlAlgorithms();
console.log(JSON.stringify(congestionControl)); // {name: "CUBIC", aim: "throughput"}
const wt = new WebTransport(url, {congestionControl}); // always works
```
Feature-detection:
```js
if (!WebTransport.getCongestionControlAlgorithms().find(({aim}) => aim == "low-latency")) {
  console.log("Requires low-latency");
}
```
```js
if (!WebTransport.getCongestionControlAlgorithms().find(({name}) => name == "SCReAM")) {
  console.log("Requires SCReAM");
}
```

Probing with the constructor also works (throws if not supported):
```js
try {
  const wt = new WebTransport(url, {congestionControl: {aim: "low-latency"});
} catch (e) {
  if (e.name == "NotSupportedError") console.log("Requires low-latency");
  throw e;
}
```
```js
try {
  const wt = new WebTransport(url, {congestionControl: {name: "SCReAM"});
} catch (e) {
  if (e.name == "NotSupportedError") console.log("Requires SCReAM");
  throw e;
}
```
LMKWYT

martinthomson:
There was a running joke during the later stages of QUIC standardization.  It goes something like this:

> Given two choices, we will always choose the most complex one.

It sounds like this is continuing in that fine tradition.

jan-ivar:
Would a way forward be to go over the requirements, get the WG to agree what they are, and use that to pick?

The tension here seems to be how low-level we want this API to be, and how much abstraction such apps can tolerate, vs how much responsibility we expect them to handle.

One litmus question might be: if two browsers end up defaulting to a different CC for `"throughput"`, would that break apps?

jan-ivar:
I have an action to update this PR based on https://github.com/w3c/webtransport/issues/365#issuecomment-1202704563.

jan-ivar:
Note this PR deviates from https://github.com/w3c/webtransport/issues/365#issuecomment-1202704563 in one regard: the constructor takes one of 3 enum values: `"default"`, `"throughput"` or `"low-latency"`.

This seems necessary to let an application truly specify whether it values throughput or low-latency more.

If the choices were only between `"default"` and `"low-latency"`, then my concern is the user agent would have no way to detect an application that values throughput, unless it interprets `"default"` to mean this, which risks cementing throughput as the default, which seems undesirable.

jan-ivar:
Meeting:
- General agreement on this direction
- Preference expressed for throwing enum
- Preference for the readable property to be more useful: a value other than default signifies that you got what you wanted
- Craft some prose to let the user agent decide

index.bs
```diff
@@ -1058,6 +1091,18 @@ that determine how WebTransport connection is established and used.
...
 :: This cannot be used with {{WebTransportOptions/allowPooling}}.
 
+: <dfn for="WebTransportOptions" dict-member>congestionControl</dfn>
+:: Optionally specifies an application's preference for a congestion control
+   algorithm tuned for either throughput or low-latency to be used when sending
+   data over this connection. This is a hint to the user agent.
+   <div class="issue atrisk">
+     <p>
+       This configuration option is considered a feature at risk due to the lack
+       of standardization and implementation of a congestion control algorithm,
```
jan-ivar, line 1101:
```suggestion
       of implementation of a congestion control algorithm,
```

PR #421, opened 2022-09-27
Add rate control event

### LINKED ISSUE #21: Stats for congestion control and bandwidth estimation

pthatcher:
Fixes #21 by adding an event to give RateControlFeedback to let the application know the maximum rate it should send at to avoid queuing.

***
<a href="https://pr-preview.s3.amazonaws.com/pthatcher/webtransport/pull/421.html" title="Last updated on Feb 14, 2023, 4:04 PM UTC (adacfb9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/421/bb7a74b...pthatcher:adacfb9.html" title="Last updated on Feb 14, 2023, 4:04 PM UTC (adacfb9)">Diff</a>


index.bs
```diff
@@ -549,8 +549,16 @@ interface WebTransport {
   Promise&lt;WebTransportSendStream&gt; createUnidirectionalStream();
   /* a ReadableStream of WebTransportReceiveStream objects */
   readonly attribute ReadableStream incomingUnidirectionalStreams;
+
+  /* of type CongestionControlFeedbackEvent */
+  attribute EventHandler oncongestioncontrolfeedback;
 };
 
+interface CongestionControlFeedbackEvent: Event {
+  readonly attribute unsigned long? targetBps;
```
jan-ivar: The design principle [§ 7.7. Use plain Events for state](https://w3ctag.github.io/design-principles/#state-and-subclassing) would suggest:
```js
wt.oncongestioncontrolfeedback = () => foo(wt.targetBps);
```

...which I suppose begs the question of why not
```js
wt.oncongestioncontrolfeedback = async () => foo((await wt.getStats()).targetBps);
```
...which in turn leaves me wondering why we need an event if it's firing on a fixed cadence...
```js
setInterval(async () => foo((await wt.getStats()).targetBps), 100);
```
Or is the idea it only fires when congestion changes? That's really the only way events "avoids pulling".

aboba: The argument for having `targetBps` be an event rather than another statistic is that `targetBps` might be needed more frequently than the rest of the statistics. If so, then you'd avoid polling `getStats()` more frequently than necessary. 

[Salsify ](https://snr.stanford.edu/salsify/) argues for the ability to adjust bandwidth usage on frequent basis (possibly per frame), based on available bandwidth. The adjustment could be made by changing per-frame QP (in progress in WebCodecs) or dropping/adding an extension layer.  Salsify makes an argument that frequent adjustment yields better video quality than adjusting the average target bitrate every keyframe.

jan-ivar:
Meeting:
- Figure out what data we want and how it will be used before we work out how to deliver it
- expected size of congestion window (how much you can send in one roundtrip), remaining/total availability, rtt?, calculations just focused at the next frame, or projections multiple frames ahead? a complex problem, and the information we are able to provide isn't well suited
- iframe is 40kb, it's going to take me 4 rtts to send, -> lower resolution etc.
- how big a frame can I send?
- reporting congstion window leads may lead to odd outcomes, like underutilizing bandwidth.
- a leaky abstraction, whereas sites need a lot of context

wilaw:
@pthatcher  - while we debate this PR, can you fix the ipr issue signalled above? The system would like you to link your github accounts with your W3C account so that PRs can be automatically accepted.

pthatcher:
> @pthatcher - while we debate this PR, can you fix the ipr issue signalled above? The system would like you to link your github accounts with your W3C account so that PRs can be automatically accepted.

It's fixed now.

index.bs
pthatcher: The idea is that the event would only fire when the value changes, so calling .getStats() doesn't make sense, nor does a fixed cadence.

pthatcher: Which somewhat contradicts what I wrote in the PR :).  I don't remember why I put in an interval.  I think I'll just change it to "fire when it changes".

pthatcher:
> Meeting:
> 
> * Figure out what data we want and how it will be used before we work out how to deliver it

I think we want a bitrate that means "if you write this much or less, you won't cause queueing", so you can go throttle you video encoder, for example. 

> * expected size of congestion window (how much you can send in one roundtrip), remaining/total availability, rtt?, calculations just focused at the next frame, or projections multiple frames ahead? a complex problem, and the information we are able to provide isn't well suited

I don't think we need anything more complex than the bitrate you should send at that will avoid queueing.

> * iframe is 40kb, it's going to take me 4 rtts to send, -> lower resolution etc.

Does anyone do rate control that way (based on RTT)?   It seems strange to me, but I suppose you could calculate it using the RTT. 

> * how big a frame can I send?

I think it's less about an individual frame and more about bitrate over time.  If you exceed the bitrate for a brief period (with a key frame), it will queue, so make sure you go lower than the target rate for a while after to a make sure it drains.

> * reporting congstion window leads may lead to odd outcomes, like underutilizing bandwidth.
> * a leaky abstraction, whereas sites need a lot of context

I think we should stick with the rate, not the window.  In fact, many algorithms, like googcc calculate target rate directly, not a window.

index.bs
pthatcher: I made it a max interval so if it doesn't change, it doesn't fire, and if it changes a lot, it will be throttled.  Hopefully that's the best of both worlds (avoid polling and avoid too often of an event).

index.bs
```diff
@@ -549,8 +549,20 @@ interface WebTransport {
...
+  attribute unsigned long? congestionControlFeedbackMinIntervalMs;
+  /* of type CongestionControlFeedbackEvent */
+  /* fires when the value changes, but at most every congestionControlFeedbackMinIntervalMs */
+  attribute EventHandler oncongestioncontrolfeedback;
 };
 
+interface CongestionControlFeedbackEvent: Event {
+  /* The rate you should send, on average, to avoid queuing */
+  readonly attribute unsigned long? averageTargetBps;
+}
```
jan-ivar, line 563:
```
      FATAL ERROR: IDL SYNTAX ERROR LINE: 35 - expected ";"
```

index.bs
```diff
@@ -549,8 +549,20 @@ interface WebTransport {
   Promise&lt;WebTransportSendStream&gt; createUnidirectionalStream();
   /* a ReadableStream of WebTransportReceiveStream objects */
   readonly attribute ReadableStream incomingUnidirectionalStreams;
+
+  /* if unset, default is 1 second */
+  attribute unsigned long? congestionControlFeedbackMinIntervalMs;
+  /* of type CongestionControlFeedbackEvent */
+  /* fires when the value changes, but at most every congestionControlFeedbackMinIntervalMs */
+  attribute EventHandler oncongestioncontrolfeedback;
```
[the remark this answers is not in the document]
pthatcher: I changed to "RateControl" instead of "CongestionControl".

index.bs
pthatcher: I hope I fixed it.

index.bs
pthatcher: This if fixed (an event with no payload plus a property).

pthatcher:
I think this is ready for more review.

index.bs
```diff
@@ -549,6 +549,18 @@ interface WebTransport {
...
+  attribute unsigned long? rateControlFeedbackMinIntervalMs;
+  /* fires when the value changes, but at most every rateControlFeedbackMinIntervalMs */
+  attribute EventHandler onratecontrolfeedback;
+  readonly attribute RateControlFeedback rateControlFeedback;
+};
+
+[Exposed=(Window,Worker), SecureContext]
+interface RateControlFeedback {
+  /* The rate you should send, on average, to avoid queuing */
+  readonly attribute unsigned long? averageTargetBps;
```
martinthomson, line 563: bits or bytes?

over what period?

martinthomson (commented):
Do you have any plans to write some prose to go with this?

index.bs
```diff
@@ -549,6 +549,18 @@ interface WebTransport {
   Promise&lt;WebTransportSendStream&gt; createUnidirectionalStream();
   /* a ReadableStream of WebTransportReceiveStream objects */
   readonly attribute ReadableStream incomingUnidirectionalStreams;
+
+  /* if unset, default is 1 second */
+  attribute unsigned long? rateControlFeedbackMinIntervalMs;
+  /* fires when the value changes, but at most every rateControlFeedbackMinIntervalMs */
```
martinthomson, line 555: "the value" here refers to which now?

jan-ivar: Is there a lot of precedent for putting typing ("Ms") into names? I would have expected `rateControlFeedbackMinInterval`. 

I think milliseconds can be inferred in the web platform thanks to [§ Use milliseconds for time measurement](https://w3ctag.github.io/design-principles/#milliseconds).

pthatcher: I removed the Ms suffix.

index.bs
pthatcher: Bits, and I made the period question clear in the text backing up the IDL.

pthatcher:
> Do you have any plans to write some prose to go with this?

I have added prose.

pthatcher:
> Meeting:
> 
> * unsigned long long
> * time range? rtt? coming out of a congestion control algorithm
> * what about pooling? Is it over multiple sessions?
> * algorithms to back up webidl

All of these have been added/fixed.

pthatcher:
I changed to a dictionary due to some feedback during one of the meetings, but now I find that I can't do that with an attribute.  Should I use a getter method or change back to an interface?

pthatcher:
From discussion:
1. Change field from "dequeueRate" to "sendRate"
2. Change .rateConfrolFeedback back to an interface
3. Add something to the privacy section about this value when polling is happening.

index.bs
```diff
@@ -1765,6 +1786,68 @@ object |transport|, run these steps.
...
+  </tr>
+ </tbody>
+</table>
+
+## Attributes ##  {#rate-control-feedback-attributes}
+
+: <dfn for="WebTransportRateControlFeedback" attribute>sendRate</dfn>
+:: The getter steps are to return [=this=]'s {{WebTransportRateControlFeedback/[[SendRate]]}}.
+   This is the rate at which queued data will be dequeued and sent by the user agent, in bits per second.
+   If the application wishes to avoid queuing, it should not send more than this rate.
```
jan-ivar, line 1825: We try to avoid "should" except for normative statments, and we can't make normative statements on applications. How about:   
```suggestion
   If the application wishes to avoid queuing, it can send less than at this rate.
```
?

index.bs
```diff
@@ -2081,6 +2164,17 @@ that the client knows what the server certificate is or how to contact it is
 not sufficient.  The application has to establish the identity of the client
 in-band if necessary.
 
+## Pooled Sessions
+
+When pooling, if information from to session is affected by the activity of another
```
jan-ivar, line 2169:
```suggestion
When pooling, if information from one session is affected by the activity of another
```

index.bs
```diff
@@ -1765,6 +1786,68 @@ object |transport|, run these steps.
...
+send rate {sendRate}, run these steps.
+
+1. Let |feedback| be a [=new=] {{WebTransportRateControlFeedback}}, with:
+    : {{WebTransportRateControlFeedback/[[SendRate]]}}
+    :: |sendRate|
+1. Return |feedback|.
+</div>
+
+<div algorithm="updating a WebTransportRateControlFeedback">
+When the user agent changes the rate at which is dequeues and sends for a [=session=],
```
jan-ivar, line 1845:
```suggestion
When the user agent changes the rate at which it dequeues bytes for sending on a [=session=],
```

pthatcher:
I have resolved everything addressed in the meeting and the comments in the PR.

index.bs
```diff
@@ -56,10 +56,12 @@ Boilerplate: omit conformance
 <pre class="link-defaults">
 spec:infra; type:dfn; for:set; text:for each
 spec:infra; type:dfn; for:/; text:set
+spec:infra; type:dfn; text:implementation-defined
 spec:streams; type:interface; text:ReadableStream
 spec:fetch; type:dfn; for:/; text:fetch
 spec:url; type:dfn; text:scheme
 spec:url; type:dfn; text:fragment
+spec:webdriver-bidi; type:dfn; text:session
```
pthatcher, line 64: I needed to add this to make bikshed work happy, but it seems to have nothing to do with my PR and it seems wrong.

index.bs
```diff
@@ -855,6 +863,20 @@ these steps.
...
+: <dfn for="WebTransport" attribute>rateControlFeedbackMinInterval</dfn>
+:: The application's preference, if changed from the default,
+   for the minimum interval between any two times when
+   [=this=]'s onratecontrolfeedback is fired.
+   This effectively allows the application
+   to control the maximum frequency of this event.
+
+: <dfn for="WebTransport" attribute>onratecontrolfeedback</dfn>
+:: Fired when [=this=]'s rateControlFeedback changes.
+   May be delayed by  [=this=]'s rateControlFeedbackMinInterval.
```
jan-ivar, line 875: Spec should define the `ratecontrolfeedback` event, not just the `onratecontrolfeedback` event handler which are separate things.

jan-ivar:
Also make WebTransport an EventTarget.

index.bs
vasilvv: I think you need to replace all instances of `[=session=]` with `[=WebTransport session=]` in the PR.

pthatcher:
I've done my best to make WebTransport an EventTarget and define the ratecontrolfeedback event, copying off of the WebRTC and WebCodecs specs.  Please let me know what I did wrong :).

index.bs
```diff
@@ -663,6 +669,8 @@ agent MUST run the following steps:
    congestion control algorithms that optimize for |congestionControl|, as allowed by
    [[!RFC9002]] [section 7](https://datatracker.ietf.org/doc/html/rfc9002#section-7),
    then set |congestionControl| to `"default"`.
+1. Let |rateControlFeedbackMinInterval| be 1000 milliseconds.
+1. Let |rateControlFeedback| be a [=new=] {{WebTransportRateControlFeedback}}.
```
jan-ivar, line 673: These stack variables are unused in this algorithm. They probably need to be [[InternalSlots]] on the wt instead.

index.bs
```diff
@@ -855,6 +863,26 @@ these steps.
    requested but not satisfied, then the value is `"default"`
    The getter steps are to return [=this=]'s {{[[CongestionControl]]}}.
 
+: <dfn for="WebTransport" attribute>rateControlFeedbackMinInterval</dfn>
+:: The application's preference, if changed from the default,
+   for the minimum interval between any two times when
```
jan-ivar, line 868: As mentioned above, I think we should make the default readable instead to simplify, which would mean rephrasing this.

index.bs
```diff
@@ -855,6 +863,26 @@ these steps.
...
+   [=this=]'s onratecontrolfeedback is fired.
```
jan-ivar, line 869:
```suggestion
   the [=ratecontrolfeedback=] event is fired.
```

index.bs
```diff
@@ -1824,6 +1853,76 @@ object |transport|, and a |sendOrder|, run these steps.
 
 </div>
 
+# Interface `WebTransportRateControlFeedback` #  {#rate-control-feedback}
+
+A {{WebTransportRateControlFeedback}} contains information
+useful for the application to control the rate at which it sends data.
```
jan-ivar, line 1859:
```suggestion
useful for the application to adapt the rate at which it sends data.
```

index.bs
```diff
@@ -1824,6 +1853,76 @@ object |transport|, and a |sendOrder|, run these steps.
...
+  <tr>
+   <th>Internal Slot
+   <th>Description (<em>non-normative</em>)
+  </tr>
+ </thead>
+ <tbody>
+  <tr>
+   <td><dfn>`[[SendRate]]`</dfn>
+   <td class="non-normative">The rate at which the {{WebTransport}} dequeues and sends data,
+   which is the maximum rate the applcation should send at to avoid queuing.
```
jan-ivar, line 1883:
```suggestion
   which is the maximum rate the application can send at without inducing queue buildup.
```

index.bs
```diff
@@ -1824,6 +1853,76 @@ object |transport|, and a |sendOrder|, run these steps.
...
+</table>
+
+## Attributes ##  {#rate-control-feedback-attributes}
+
+: <dfn for="WebTransportRateControlFeedback" attribute>sendRate</dfn>
+:: The getter steps are to return [=this=]'s {{WebTransportRateControlFeedback/[[SendRate]]}}.
+   This is the rate at which queued data will be dequeued and sent by the user agent, in bits per second.
+   If the application wishes to avoid queuing, it can send at a rate less than this value.
+   This rate applies to all streams and datagrams that share a [=session=]
+   and is calculated using {{WebTransport/congestionControl}}. If this [=session=] is
```
jan-ivar, line 1895:
```suggestion
   and is calculated by the congestion control algorithm (potentially chosen by
   {{WebTransport/congestionControl}}). If this [=session=] is
```

index.bs
```diff
@@ -1824,6 +1853,76 @@ object |transport|, and a |sendOrder|, run these steps.
...
+   pooled with others in a shared [=connection=], it is the rate at which the user
+   agent will drain the queues for this [=session=].  If the user agent cannot
+   determine a value (perhaps because of pooling), it may leave the value unset.
```
jan-ivar, line 1898: Leaving the value unset seems unfortunate, and possibly bad for web compat. Should we encourage user agents to make up a value?

index.bs
```diff
@@ -1824,6 +1853,76 @@ object |transport|, and a |sendOrder|, run these steps.
...
+
+## Procedures ## {#rate-control-feedback-procedures}
+
+<div algorithm="create a WebTransportRateControlFeedback">
+
+To <dfn for=WebTransportRateControlFeedback>create</dfn> a {{WebTransportRateControlFeedback}} with a
+send rate {sendRate}, run these steps.
```
jan-ivar, line 1905:
```suggestion
To <dfn for=WebTransportRateControlFeedback>create</dfn> a {{WebTransportRateControlFeedback}}, given a |sendRate|, run these steps:
```

index.bs
```diff
@@ -1824,6 +1853,76 @@ object |transport|, and a |sendOrder|, run these steps.
...
+    : {{WebTransportRateControlFeedback/[[SendRate]]}}
+    :: |sendRate|
+1. Return |feedback|.
+
+</div>
+
+<div algorithm="updating a WebTransportRateControlFeedback">
+
+When the user agent changes the rate at which it dequeues bytes for sending on a [=session=],
+it MUST run these steps.
```
jan-ivar, line 1917:
```suggestion
When the user agent changes the rate at which it dequeues bytes for sending on a
{{WebTransport}} |transport|'s [=WebTransport/session=],
it MUST [=queue a network task=] with |transport| to run these steps:
```

index.bs
```diff
@@ -1824,6 +1853,76 @@ object |transport|, and a |sendOrder|, run these steps.
...
+
+1. Update the value of {{WebTransportRateControlFeedback/[[SendRate]]}} of the
+   {{WebTransport/rateControlFeedback}} of the [=session=].
+1. Fire an event named {{WebTransport/ratecontrolfeedback}} at the {{WebTransport}} of the [=session=].
```
jan-ivar, line 1921:
```suggestion
1. Set |transport|.{{WebTransport/rateControlFeedback}}.{{WebTransportRateControlFeedback/[[SendRate]]}}
    to the new rate for [=session=].
1. Fire an event named {{WebTransport/ratecontrolfeedback}} at |transport|.
```

index.bs
```diff
@@ -2140,6 +2239,17 @@ that the client knows what the server certificate is or how to contact it is
...
 in-band if necessary.
 
+## Pooled Sessions
+
+When pooling, if information from one session is affected by the activity of another
+session, one session may infer information about the activity of another session.
+For example, if the rate control feedback of one session is affected by the rate at
+which an application is sending data on another session, one session may learn about the
+the rate at which another application is sending data.  To avoid this, a user agent
+may restrict access to information or modify information provided.  For example, it may leave
```
jan-ivar, line 2249: This section is non-normative, so we want to avoid "may" which might be interpreted as MAY.
```suggestion
session, one session could infer information about the activity of another session.
For example, if the rate control feedback of one session is affected by the rate at
which an application is sending data on another session, one session could learn about the
the rate at which another application is sending data.  To avoid this, a user agent
can restrict access to information or modify information provided.  For example, it could leave
```

index.bs
```diff
@@ -2140,6 +2239,17 @@ that the client knows what the server certificate is or how to contact it is
...
+the send rate of the rate control feedback undefined, or modify it to make it more difficult
+to infer anything from it.
```
jan-ivar, line 2251: As mentioned earlier, I'm not sure I like the idea of leaving it undefined. How about something like:
```suggestion
the send rate at an artifical value to make it more difficult to infer anything from it.
```

jan-ivar (changes requested):
Thanks, this is coming together! Lots of detail to work out here, so pardon all the comments.

index.bs
pthatcher: Yes, that looks like it was it.  Thank you.

index.bs
pthatcher: I think that's the case now.

index.bs
pthatcher: OK, I reworded it so the default is readable.

index.bs
pthatcher: I added some internal slots to fix this.

index.bs
```diff
@@ -560,6 +562,10 @@ interface WebTransport {
       optional WebTransportSendStreamOptions options = {});
   /* a ReadableStream of WebTransportReceiveStream objects */
   readonly attribute ReadableStream incomingUnidirectionalStreams;
+
+  attribute unsigned long? rateControlFeedbackMinInterval;
+  attribute EventHandler onratecontrolfeedback;
+  readonly attribute WebTransportRateControlFeedback rateControlFeedback;
```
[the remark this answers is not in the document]
pthatcher: I made #470 and #471 to look at which one we like better (or this one as-is).

index.bs
```diff
@@ -623,6 +627,16 @@ A {{WebTransport}} object has the following internal slots.
    throughput or low latency was requested by the application and satisfied
    by the user agent, or `"default"`.
   </tr>
+  <tr>
+   <td><dfn>`[[RateControlFeedbackMinInterval]]`</dfn>
+   <td class="non-normative">A {{long}} indicating the maximum frequency
+   that the {{WebTransport/ratecontrolfeedback}} will be fired.
```
martinthomson, line 633: Isn't this the minimum time between events?  Or even the duration over which the event is suppressed after receiving another.

Sure interval is inversely proportional to rate, but only loosely.

index.bs
```diff
@@ -2140,6 +2251,16 @@ that the client knows what the server certificate is or how to contact it is
...
 
+## Pooled Sessions
+
+When pooling, if information from one session is affected by the activity of another
+session, one session could infer information about the activity of another session.
+For example, if the rate control feedback of one session is affected by the rate at
+which an application is sending data on another session, one session could learn about the
+the rate at which another application is sending data.  To avoid this, a user agent
+can restrict access to information or modify information provided.  For example, it could leave
+the send rate at an artifical value to make it more difficult to infer anything from it.
```
martinthomson, line 2262: This seems pretty loose.  This boils down to "maybe the browser could work something out", which is pretty unsatisfying.  We should be more concrete, either block access to the attribute (which reveals the presence of another entity, so that needs thought) or define an exemplar strategy for making it available.

index.bs
```diff
@@ -1824,6 +1864,77 @@ object |transport|, and a |sendOrder|, run these steps.
...
+    :: |sendRate|
+1. Return |feedback|.
+
+</div>
+
+<div algorithm="updating a WebTransportRateControlFeedback">
+
+When the user agent changes the rate at which it dequeues bytes for sending on a
+{{WebTransport}} |transport|'s [=WebTransport session=],
+it MUST [=queue a network task=] with |transport| to run these steps:
```
martinthomson, line 1929: This requirement is impossible, or nearly so.  How does the UA monitor the rate?

index.bs
```diff
@@ -1824,6 +1864,77 @@ object |transport|, and a |sendOrder|, run these steps.
...
+
+1. Set |transport|.{{WebTransport/rateControlFeedback}}.{{WebTransportRateControlFeedback/[[SendRate]]}}
+    to the new rate for [=WebTransport session=].
+1. Fire an event named {{WebTransport/ratecontrolfeedback}} at |transport|.
```
martinthomson, line 1933: Where is the event limiting feature here?

martinthomson (commented):
This is incomplete, so I object to it having been merged.

PR #415, opened 2022-08-24
Add Samples Directory (and WebCodecs Echo Sample)

aboba:
Fixes https://github.com/w3c/webtransport/issues/416

Live webcodecs echo sample: https://webrtc.internaut.com/wc/wtSender2/

PR #430, opened 2022-10-26
Add WebCodecs-Echo to the samples directory

aboba:
Rebase of https://github.com/w3c/webtransport/pull/415

Live demo: https://webrtc.internaut.com/wc/wtSender10/

Requires Chromium M108+ (to support BYOB reader)

Known issues: 

1. Device selection (different behavior in Chrome vs. Chrome Canary). 
2. FIXED: Non-optimal write strategy (added `await writer.ready`, removed `await` from `writer.write`, `writer.close`)
3. Server support for "cut-through" forwarding
4. Server support for `RESET_STREAM` forwarding
5. PARTIALLY FIXED: Non-optimal read strategy (removed awaits on read side, while {} loop). 
6. FIXED: Add writer.releaseLock() and reader.releaseLock()

### DIFF
--- a/samples/webcodecs-echo/css/main.css
+++ b/samples/webcodecs-echo/css/main.css
@@ -1,0 +1,267 @@
+/*
+ *  Copyright (c) 2015 The WebRTC project authors. All Rights Reserved.
+ *
+ *  Use of this source code is governed by a BSD-style license
+ *  that can be found in the LICENSE file in the root of the source
+ *  tree.
+ */
+.hidden {
+  display: none;
+}
+
+.highlight {
+  background-color: #eee;
+  font-size: 1.2em;
+  margin: 0 0 30px 0;
+  padding: 0.2em 1.5em;
+}
+
+.warning {
+  color: red;
+  font-weight: 400;
+}
+
+@media screen and (min-width: 1000px) {
+  /* hack! to detect non-touch devices */
+  div#links a {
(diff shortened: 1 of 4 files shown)


aboba:
Recent update removes `await` in `writer.write(chunk)`.  This appears to decrease glass-glass latency substantially and reordering is now observed on the receiver (requiring a reordering buffer).  However, partial reliability no longer works.

ricea:
> 3\. Tried using `writer.close().then()` instead of `await writer.close()`. This resulted in much more reordering, but also frame RTTs that grew over time to very high levels (2000+ ms), causing many more aborted writes.  Not exactly sure why this change made so much difference.

This really shouldn't make any difference. Very long RTTs would seem to imply that backpressure is not working effectively.

aboba:
@ricea `writer.close().then()` results in a considerably higher average bitrate than `await writer.close()` and the example is not decreasing the encoding rate based on the increasing frame RTTs. My initial hypothesis was that the sending bandwidth was exceeding the bottleneck bandwidth with `writer.close().then` but not with `await writer.close()`.  However, the increasing frame RTTs are showing up with `writer.close.then()` even when the average bitrate target is greatly reduced (e.g. 100 Kbps).  

WIth frame/stream transport running at 30 fps, a frame RTT of 2 seconds would imply 60 open streams with `writer.close().then()`.  Could that be (part of) the problem?

Issue #402, opened 2022-06-21
Q: When does writer.write() resolve?


jan-ivar:
[Write](https://w3c.github.io/webtransport/#webtransportsendstream-write) says: _"[Send](https://w3c.github.io/webtransport/#stream-send) bytes on stream.[[[InternalStream]]](https://w3c.github.io/webtransport/#dom-webtransportsendstream-internalstream-slot) and wait for the operation to complete."_
[Close](https://w3c.github.io/webtransport/#webtransportsendstream-close) says: _"[Send](https://w3c.github.io/webtransport/#stream-send) FIN on stream.[[[InternalStream]]](https://w3c.github.io/webtransport/#dom-webtransportsendstream-internalstream-slot) and wait for the operation to complete."_

But when is "the operation" "complete"? [Section 2.2](https://datatracker.ietf.org/doc/html/draft-ietf-quic-transport#section-2.2) doesn't say AFAIK. Is it when:
 1. The data has been handed off to some lower-level network queue for sending?
 2. The data has left the network card (once)?
 3. An ACK for this data has been received?

I can reason myself to an answer:
 - when it can no longer fail (3)

But we should perhaps write this down.

jan-ivar:
Meeting:
- For datagrams the answer (in Chrome) appears to be 1
- For streams, we're not sure yet.

jan-ivar:
Meeting:
- For write, the answer seems to be 1. The data has been handed off to some lower-level network queue for sending
- For close the algorithm also waits on _"Wait for stream.[[[InternalStream]]](https://w3c.github.io/webtransport/#dom-webtransportsendstream-internalstream-slot) to enter the "Data Recvd" state"_

aboba:
After removing the `await` before `writer.write(chunk)` in the WebCodecs echo sample, latency went down and reordering was observed on the receiver.  This is consistent with handoff to a lower-level network queue, but it is also indicates that `await writer.write()` is sub-optimal. We should demonstrate the recommended approach in examples and sample code.

ricea:
As a general rule it's better to `await writer.ready` rather than `await writer.write()` but I'm not sure how much difference it will make here.

aboba:
In practice, it seems to make a substantial difference. `await writer.write` or even `await writer.close()` appears to preclude concurrency (the interleaving of sends on multiple streams).

jan-ivar:
[...]

PR #434, opened 2022-11-09
Update examples to not await write(bytes)

jan-ivar:
Fixes https://github.com/w3c/webtransport/issues/402. Also adds note with a link to the streams spec on this.

Also tries to use `bytes` vs `data` consistently to distinguish encoded vs decoded data respectively.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webtransport/pull/434.html" title="Last updated on Nov 9, 2022, 9:25 PM UTC (d914daf)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/434/f2fe7de...jan-ivar:d914daf.html" title="Last updated on Nov 9, 2022, 9:25 PM UTC (d914daf)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2095,9 +2095,9 @@ following example datagrams are only sent if the transport is ready to send.
 async function sendDatagrams(url, datagrams) {
   const wt = new WebTransport(url);
   const writer = wt.datagrams.writable.getWriter();
-  for (const datagram of datagrams) {
+  for (const bytes of datagrams) {
     await writer.ready;
-    writer.write(datagram).catch(() => {});
+    writer.write(bytes).catch(() => {});
   }
 }
 </pre>
@@ -2117,8 +2117,8 @@ async function sendFixedRate(url, createDatagram, ms = 100) {
   const wt = new WebTransport(url);
   await wt.ready;
   const writer = wt.datagrams.writable.getWriter();
-  const datagram = createDatagram();
-  setInterval(() => writer.write(datagram).catch(() => {}), ms);
+  const bytes = createDatagram();
+  setInterval(() => writer.write(bytes).catch(() => {}), ms);
 }
 </pre>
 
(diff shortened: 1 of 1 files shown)


wilaw (commented):
Updates look good. While you are refreshing examples, WDYT about adding a new example to show sending a series of data objects as a stream of short lived streams. This mirrors the common use case we are seeing with WARP/RUSH of sending frames or GOPs of video over individual streams.

aboba:
Have questions about `async function sendData(url, ...data)`

Seems like this async function might be called in a loop where you might want to be able to send multiple messages concurrently, so that `await` usage could be problematic. 

"const writable = await wt.createUnidirectionalStream();"

Can we remove the  `await` here?  Also am concerned with blocking on

  `await writer.close();`

Might substitute `writer.close().catch()`

aboba:
@willaw I have examples of this in the webcodecs-echo sample.  It took a while to get the message/stream writing functioning well, particularly combined with partial reliability, so might be worth using as an example. 

I also wonder about how to get optimal concurrency on the receiving side, whether `await` usage there is a problem as well.

jan-ivar:
> After substituting `writer.close().catch()` there was an improvement in concurrency, indicating that `await writer.close()` will block.

Correct. [close()](https://streams.spec.whatwg.org/#ws-prototype) blocks until everything has been _"successfully written and the stream successfully closes, or rejects if an error is encountered during this process"_. — This is necessary to surface any errors from failure to send.

> Have questions about `async function sendData(url, ...data)` ... Seems like this async function might be called in a loop where you might want to be able to send multiple messages concurrently.

If `sendData(url, ...data)` fails to send everything, it rejects. That's its contract, and the promise surfaces that error.

To get concurrency, loop synchronously and gather up promises. There's no back pressure between streams.
```js
await Promise.all(messages.map(arrayOfData => sendData(url, ...arrayOfData)));
```

At least the example now shows how to queue multiple chunks of data efficiently. If we still find it confusing, one option might be to model it closer to the [streams spec's example](https://streams.spec.whatwg.org/#example-manual-write-dont-await), which avoids this problem by never succeeding (`writeSuppliedBytesForever` only rejects on failure).

jan-ivar:
> I also wonder about how to get optimal concurrency on the receiving side, whether await usage there is a problem as well.

Every `await` is an opportunity to do something else, i.e. the antithesis of concurrency. While `await p; f()` is just syntactic sugar to make `p.then(f)` read like blocking code, apps still need to think through which operations depend on one another (and need to form promise-chains to surface errors correctly) and which don't, and shouldn't. `async` IFFEs with `await` can be used to branch chains.

The streams spec introduces `pipeTo` and `pipeThrough` to guide apps with how to organize this difficult code.

aboba:
@jan-ivar I have been experimenting with removal of `await` on both the sending and receiving sides.  Replacing `await writer.write()` with `await writer.ready; writer.write()` went smoothly, but replacing `await writer.close()` with `writer.close().then()` has had unexpected side effects, resulting in growing frame RTTs, regardless of the sending rate.  I am wondering if this is the result of leaving too many streams open. 

`writer.close().then()`:  https://webrtc.internaut.com/wc/wtSender7/
`await writer.close()`:   https://webrtc.internaut.com/wc/wtSender8/

aboba (changes requested):
Line 2159: `await writer.close()`  Is this also a concurrency issue?  I have tried replacing this with `writer.close().then()` in the WebCodecs echo sample, and got some unexpected results.

PR #438, opened 2022-11-30
Add order option to createUnidirectionalStream().

### LINKED ISSUE #431: Support minimal prioritization necessary to enable IETF MoQ base streaming protocol
There is an ongoing effort within the recently chartered Media Over Quic ([MoQ](https://datatracker.ietf.org/wg/moq/about/)) work group to standardize a simple low-latency media delivery solution for ingest and distribution of media. The use-cases include live streaming, gaming, and media conferencing and these correlate highly with the target [use-cases for WebTransport](https://github.com/w3c/webtransport/blob/main/use-cases.md) within the W3C. As such, it seems reasonable that the initial version of the W3C Webtransport API should at a minimum provide an API surface that enables MoQ base protocol to be implemented in a browser user-agent. 

IETF#115 next week will look consider a [merged proposal between WARP and RUSH](https://www.ietf.org/archive/id/draft-lcurley-warp-02.txt). From the draft text, the requirement around prioritization is:

`This is a strict prioritization scheme, such that any available bandwidth is allocated to streams in ascending priority order.  The sender SHOULD prioritize streams based on the delivery order.  If two streams have the same delivery order, they SHOULD receive equal bandwidth (round-robin).`

Therefore the minimum send prioritization scheme necessary for the web API seems to be one which
1. Allows a strict numeric priority to be set on each WT stream at construction time
2. Allows similar priority to be set against all WT datagrams (as they will be competing with streams for available bandwidth).
3. Requires the user-agent to send the queued high priority data first.  Data with equal priority is round-robined (equal weight). 

This scheme could be extended to incorporate alternative weighting schemes, abstract flows of streams and datagrams, but none of these features are necessary for an initial support of base protocol MoQ and can be deferred to future releases of the API.

Related issues:
Weighted flows of datagrams and short-lived streams [#419](https://github.com/w3c/webtransport/issues/419) 
Datagram vs stream & relative stream prioritization [#62](https://github.com/w3c/webtransport/issues/62)

jan-ivar:
Fixed #431.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webtransport/pull/438.html" title="Last updated on Dec 20, 2022, 6:12 AM UTC (44b090b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/438/87df5e2...jan-ivar:44b090b.html" title="Last updated on Dec 20, 2022, 6:12 AM UTC (44b090b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -542,12 +542,13 @@ interface WebTransport {
 
   readonly attribute WebTransportDatagramDuplexStream datagrams;
 
-  Promise&lt;WebTransportBidirectionalStream&gt; createBidirectionalStream();
+  Promise&lt;WebTransportBidirectionalStream&gt; createBidirectionalStream(
+      optional WebTransportSendStreamOptions options = {});
   /* a ReadableStream of WebTransportBidirectionalStream objects */
   readonly attribute ReadableStream incomingBidirectionalStreams;
 
   Promise&lt;WebTransportSendStream&gt; createUnidirectionalStream(
-      optional WebTransportUnidirectionalStreamOptions options = {});
+      optional WebTransportSendStreamOptions options = {});
   /* a ReadableStream of WebTransportReceiveStream objects */
   readonly attribute ReadableStream incomingUnidirectionalStreams;
 };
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -546,7 +546,8 @@ interface WebTransport {
   /* a ReadableStream of WebTransportBidirectionalStream objects */
```
martinthomson, line 546: One line above, we need the same argument.

Regarding names, we should probably have this type be a union of types.  When creating a unidirectional stream, you include "sending" options, but when creating a bidirectional stream, you might include both "sending" and "receiving" options.

index.bs
```diff
@@ -1160,6 +1161,36 @@ The dictionary SHALL have the following attributes:
...
+   {{WebTransportSendStream}} in to participating in <dfn>strict ordering</dfn>.
+   Bytes currently queued on [=strict ordering|strictly ordered=]
+   {{WebTransportSendStream}}s MUST be sent ahead of bytes currently queued on
+   other [=strict ordering|strictly ordered=] {{WebTransportSendStream}}s
+   created with lower order numbers.
+
+   If no order number is provided, the created {{WebTransportSendStream}}
+   MUST NOT participate in [=strict ordering=], meaning the order in which the
+   user agent sends bytes from it relative to other {{WebTransportSendStream}}s
+   remains [=implementation-defined=].
```
martinthomson, line 1189: Did we not agree on default behaviour?  Even setting expectations would be good.  Something like streams created first will generally be sent first, with a nod toward the starvation text in RFC 9218...

martinthomson (changes requested):
For the missing bidirectional streams component, but this is OK, I guess.

index.bs
jan-ivar: I wasn't planning on introducing this to bidirectional streams (the dictionary is named WebTransportUnidirectionalStreamOptions). Figuring out their semantics seems best deferred to when we have a use case for that, which seems out of scope for https://github.com/w3c/webtransport/issues/431.

index.bs
jan-ivar: I've created a separate PR https://github.com/w3c/webtransport/pull/441 for that, though the two are perhaps inextricably linked, if we want to skip discussing alternatives not on the table.

jan-ivar:
Remaining: bikeshed on `order` vs `priority`

jan-ivar:
Meeting:
- Add to bidirectional streams
- make it signed
- Replace MUST NOT language with implementation-defined.

Issue #435, opened 2022-11-22
allow opening streams and sending datagrams in parallel to the CONNECT request


marten-seemann:
WebTransport allows the client to (optimistically) open streams and send datagrams right after issuing the upgrade request, while the CONNECT request is still in flight. This saves one network roundtrip, assuming that the server accepts the WebTransport session. In case the server rejects the upgrade, these streams will be reset and the datagrams will be ignored.

This was recently clarified in the HTTP/3 WebTransport spec: https://github.com/ietf-wg-webtrans/draft-ietf-webtrans-http3/pull/80

Currently, this is not possible using the browser API. For example, [`createBidirectionalStream()`](https://w3c.github.io/webtransport/#dom-webtransport-createbidirectionalstream) is specified to wait until the connection state is `"connected"`. It would be really nice if there was a way to make use of this feature from the browser.

martinthomson:
This is potentially an advantage.  If the server redirects, data sent in QUIC-native streams or datagrams might be hard - or even impossible - to replicate properly when re-attempting a connection.  There are any number of reasons why what works on one connection might not work on a subsequent attempt.

The latency hit, which is non-trivial, is unfortunate though.

jan-ivar:
Just because a method returns a promise doesn't mean JS needs to `await` it. I suppose we could reorder [these two steps](https://w3c.github.io/webtransport/#dom-webtransport-createbidirectionalstream):

1. Wait for transport.[[[State]]](https://w3c.github.io/webtransport/#dom-webtransport-state-slot) to be "connected".
2. Let internalStream be the result of [creating a bidirectional stream](https://w3c.github.io/webtransport/#session-create-a-bidirectional-stream) with transport.[[[Session]]](https://w3c.github.io/webtransport/#dom-webtransport-session-slot).

jan-ivar:
> Just because a method returns a promise doesn't mean JS needs to await it.

But yeah you do need to `await` it to get the stream object. 🤦 I'm not awake today.

aboba:
I vote for Option 1 (for `createBidirectionalStream` and `createUnidirectionalStream`). As far as I can tell, nothing else would need to be changed.

PR #440, opened 2022-12-04
Allow opening streams in parallel to the CONNECT request

aboba:
Fixes: https://github.com/w3c/webtransport/issues/435

Note: Writing datagrams while the transport's state  is `"connecting"` is already allowed.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/pull/440.html" title="Last updated on Dec 4, 2022, 11:55 PM UTC (84e4710)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/440/5841bfe...84e4710.html" title="Last updated on Dec 4, 2022, 11:55 PM UTC (84e4710)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -905,7 +905,6 @@ these steps.
    1. Run the following steps [=in parallel=], but abort them whenever |transport|'s
       {{[[State]]}} becomes `"closed"` or `"failed"`, and instead
       [=queue a network task=] with |transport| to [=reject=] |p| with an {{InvalidStateError}}.
-      1. Wait for |transport|.{{[[State]]}} to be `"connected"`.
       1. Let |internalStream| be the result of [=creating a bidirectional stream=] with
          |transport|.{{[[Session]]}}.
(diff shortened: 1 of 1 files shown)


martinthomson (commented):
Is it worth allowing user agents to cap the number of streams they allow to be created here?  That's usually a function of stream limits, but if we're going to have to support redirects, which amount to a retry, then a cap on the number of open streams will increase the chances of the retry working seamlessly.  Not only is it more likely that the streams can be opened in a new session, the user agent has less state to track.

jan-ivar (approved):

jan-ivar:
> Is it worth allowing user agents to cap the number of streams they allow to be created here?

Is this an issue specific to opening streams in parallel to the CONNECT request, or can it be discussed separately?

jan-ivar:
Opened #446 to discuss stream limits.

PR #441, opened 2022-12-05
Define default send order.

### LINKED ISSUE #428: Do we need to specify default send behavior for web compat?
At TPAC, I presented [slides](https://docs.google.com/presentation/d/141lSOntrraY3FACjIMqad0NlrYgKjEzrK-txxY2JpjI/edit#slide=id.g1532c26172c_1_8) that asked if we need to specify default send behavior. Here's that issue:
> Today the UA decides which writables to send from next to the server
> 
> Here’s a [simple file upload fiddle](https://jsfiddle.net/jib1/z965juL7/). If I pick two equally sized files, it shows Chrome
giving equal time (flow) to each upload stream:
> 
> image
> 
> Add a third file (send-stream) and it reduces all send-rates to be equal AFAICT.
This reasonable behavior is [implementation-defined](https://infra.spec.whatwg.org/#implementation-defined), which seems nuts. A poor foundation to build JS-setable priorities on top of? — Should we specify this?

Logically, use cases probably fall in one of 3 categories:
1. My use case works regardless of how UAs prioritize sending of my multiple streams
2. My use case requires new APIs to control priorities https://github.com/w3c/webtransport/issues/62
2. My use case relies on the default behavior as observed in Chrome right now

It sounds like it'll be a while until we address 2, but what about 3?

As pointed out in https://github.com/w3c/webtransport/issues/426, apps may come to rely on browsers' send behaviors, and even deadlock over them.

jan-ivar:
Fixes #428. Contains tentative language based on assumptions in https://github.com/w3c/webtransport/pull/438.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webtransport/pull/441.html" title="Last updated on Dec 7, 2022, 11:05 PM UTC (d40fceb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/441/5841bfe...jan-ivar:d40fceb.html" title="Last updated on Dec 7, 2022, 11:05 PM UTC (d40fceb)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1361,10 +1361,18 @@ To <dfn for="WebTransportSendStream">write</dfn> |chunk| to a {{WebTransportSend
 1. Return |promise| and run the remaining steps [=in parallel=].
 1. [=stream/Send=] |bytes| on |stream|.{{WebTransportSendStream/[[InternalStream]]}} and wait for the
    operation to complete.
-   This sending MAY be interleaved with sending of previously queued streams and datagrams,
-   as well as streams and datagrams yet to be queued to be sent over this transport.
+   By default, if no order is specified, the prioritization of this sending of
+   bytes is [=implementation-defined=], but SHOULD, to the extent allowed by
+   flow control, be interleaved with sending of any previously queued bytes on
+   other streams and datagrams, as well as bytes yet to be queued on other
+   streams and datagrams to be sent over this transport, regardless of whether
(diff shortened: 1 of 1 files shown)


jan-ivar:
Meeting:
- If no order specified then streams don't starve.
- If order is specified then it may starve, if another stream of higher priority goes first
- An unordered stream should never starve

PR #453, opened 2023-01-05
Update index.bs to clarify DOMString comparison method 

### LINKED ISSUE #413: Clarify DOMString comparison method in dictionaly WebTransportHash
5.9 Configuration, verify a certificate hash
https://www.w3.org/TR/2022/WD-webtransport-20220623/#verify-a-certificate-hash

3.1.2 of operation states:
> If [algorithm](https://www.w3.org/TR/2022/WD-webtransport-20220623/#dom-webtransporthash-algorithm) of hash is equal to "sha-256", and hashValue is equal to referenceHash, the certificate is valid. Return true.

User supplied value (via constructor of `interface WebTransport` within `WebTransportOptions`) is compared with spec defined value as `is equal to`, but this text is not clear how to compare by `equal`.
This part should use defined term(s) in INFRA and refer there, like [`is` or `is identical to`](https://infra.spec.whatwg.org/#string-is) or [`is an ASCII case-insensitive match`](https://infra.spec.whatwg.org/#ascii-case-insensitive)

wilaw:
Fixes #413

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1130,7 +1130,7 @@ To <dfn>verify a certificate hash</dfn>, do the following:
 1. For every hash |hash| in |hashes|:
    1. If |hash|.{{WebTransportHash/value}} is not null:
      1. Let |hashValue| be the byte sequence which |hash|.{{WebTransportHash/value}} represents.
-     1. If {{WebTransportHash/algorithm}} of |hash| is equal to "sha-256", and |hashValue| is equal
+     1. If {{WebTransportHash/algorithm}} of |hash| is an ASCII case-insensitive match with "sha-256", and |hashValue| is identical
         to |referenceHash|, the certificate is valid. Return true.
 1. Return false.
 


index.bs
```diff
@@ -1130,7 +1131,7 @@ To <dfn>verify a certificate hash</dfn>, do the following:
 1. For every hash |hash| in |hashes|:
    1. If |hash|.{{WebTransportHash/value}} is not null:
      1. Let |hashValue| be the byte sequence which |hash|.{{WebTransportHash/value}} represents.
-     1. If {{WebTransportHash/algorithm}} of |hash| is equal to "sha-256", and |hashValue| is equal
+     1. If {{WebTransportHash/algorithm}} of |hash| is an ASCII case-insensitive match with "sha-256", and |hashValue| is identical
```
wilaw, line 1134:
```suggestion
     1. If {{WebTransportHash/algorithm}} of |hash| is an [=ASCII case-insensitive=] match with "sha-256", and |hashValue| is equal
```

jan-ivar:
Follow-up to determine impact on implementation

Issue #410, opened 2022-08-02
"QUIC Protocol Action" in #protocol-mapping is misleading


yutakahirano:
In https://w3c.github.io/webtransport/#protocol-mapping, we have a table with "QUIC Protocol Action" column. For example, the table maps "writable.abort(errorCode)" to "send RESET_STREAM with errorCode". This is misleading because we have an error code conversion logic at https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3/#section-4.3, and QUIC protocol sees the converted number.

jan-ivar:
Meeting:
- Don't use same "errorCode" name on both sides

PR #467, opened 2023-02-08
Update protocol mapping table

nidhijaju:
This PR updates the protocol mapping table, so that we don't use the words `errorCode` on both sides. It also removes `STREAM_FINAL` which is not a frame specified in QUIC.

Fixes https://github.com/w3c/webtransport/issues/410.

***
<a href="https://pr-preview.s3.amazonaws.com/nidhijaju/webtransport/pull/467.html" title="Last updated on Feb 13, 2023, 9:57 AM UTC (2a95c5b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/467/d90cf90...nidhijaju:2a95c5b.html" title="Last updated on Feb 13, 2023, 9:57 AM UTC (2a95c5b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1931,8 +1931,11 @@ To <dfn for="WebTransportError">set up</dfn> a {{WebTransportError}} |error| wit
 *This section is non-normative.*
 
 This section describes the [[QUIC]] protocol behavior of methods defined
-in this specification, utilizing [[WEB-TRANSPORT-HTTP3]]. Cause and effect may
-not be immediate due to buffering.
+in this specification, utilizing [[WEB-TRANSPORT-HTTP3]]. The application
+{{WebTransportError/streamErrorCode}} in the {{WebTransportError}} error is
+converted to an httpErrorCode, and vice versa, as specified in [[!WEB-TRANSPORT-HTTP3]]
+[section 4.3](https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3/#section-4.3).
+Cause and effect may not be immediate due to buffering.
 
   <table class="data">
     <colgroup class="header"><col></colgroup>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1931,8 +1931,11 @@ To <dfn for="WebTransportError">set up</dfn> a {{WebTransportError}} |error| wit
 *This section is non-normative.*
 
 This section describes the [[QUIC]] protocol behavior of methods defined
-in this specification, utilizing [[WEB-TRANSPORT-HTTP3]]. Cause and effect may
-not be immediate due to buffering.
+in this specification, utilizing [[WEB-TRANSPORT-HTTP3]]. The application
+{{WebTransportError/streamErrorCode}} in the {{WebTransportError}} error is
+converted to an httpErrorCode, and vice versa, as specified in [[!WEB-TRANSPORT-HTTP3]]
+[section 4.3](https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3/#section-4.3).
+Cause and effect may not be immediate due to buffering.
```
jan-ivar, line 1938: I'm wondering, since the added text is specific only to a subset of methods, if we should add it to the end instead, so the "cause and effect" statement cannot be misread as applying only to the subset?

Doesn't read as well though. WDYT?
```suggestion
in this specification, utilizing [[WEB-TRANSPORT-HTTP3]]. Cause and effect may
not be immediate due to buffering. The application
{{WebTransportError/streamErrorCode}} in the {{WebTransportError}} error is
converted to an httpErrorCode, and vice versa, as specified in [[!WEB-TRANSPORT-HTTP3]]
[section 4.3](https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3/#section-4.3).
```

jan-ivar (approved):
Thanks! Nice improvement.

index.bs
nidhijaju: Hmmm, I agree that it doesn't read as well, but I can also see why one might misread it that way. I don't have strong opinions, so I'll add the text at the end instead as you suggested.

PR #485, opened 2023-03-02
Networking privacy considerations

### LINKED ISSUE #478: Rate control privacy considerations
It is not immediately obvious, but exposing details of rate control has privacy implications.  Some of these are preexisting in ways that are probably not in scope for us to fix, even if we could.  We should at least be prepared to acknowledge them.

None of this is new information.  Sites can measure network capacity already, sometimes with very high fidelity.  All that needs is fetch, WebRTC, or any of the many other things we do with networking.  But we are providing an API with real-time (or close to it) performance and precision.  That could make measurements easier to take and more reliable.

## Details

The fundamental problem is that the available send rate is a resource that is shared between mutually distrustful entities.  Within a browser, there are multiple origins that each access the network.  In many situations, the available upstream and downstream capacity at the scope of a household (which approximates a user in many cases) is a relatively fixed resource where multiple applications compete for access.

Firstly, revealing the approximate total available capacity and RTT could be used as a fingerprint.

This is not a particularly reliable or good fingerprint, because we only provide the *current* available capacity.  On the other hand, network use tends to be sporadic and so sites might be able to assume exclusive use of the shared resource and thereby consider the value they get to be a reasonable approximation of the total.

Second, providing information about *current* available capacity also provides information about active use.  This provides sites with a means to correlate user activity across origins.

Let's say that you have two sites with independent user databases.  The sites want to learn which of the users of site A correspond to users of site B.  One site can coordinate to observe link capacity for multiple users and then have the other generate a spike of network load for certain users.  Any users with a corresponding reduction in capacity likely correspond with those users that were tested.

This sort of trick might work less well if the network bottleneck is close to a user.  For instance, it might not be possible to distinguish people from the same household who share an Internet connection.  Some network configurations will therefore never produce anything that is truly useful for correlation purposes.  But for the most part, network bottlenecks tend to be close to people, so many people will be susceptible to this.  Also, reducing the candidate set size is of some value as it opens up options for more targeted cross-site correlation tricks, like navigation tracking.

## Recommendations

I suggest that we write something about how exposing rate control information has these properties.  Referencing existing text would also be fine.  That text would go in security/privacy considerations.

Furthermore, as the problems here arise from concurrent access to the API from multiple sites, we might make some recommendations about removing or degrading access to sites that are not currently in the foreground or focus.

martinthomson:
A rough attempt at putting some words down.

You will note that the privacy mitigations here are pretty unsatisfactory.  In short, we don't have any.  That's just how the web rolls right now and I don't think that we need to spend a lot of time on trying to invent a solution.  The only solution I'm aware of involves serious degradations in performance, either by disabling concurrency across sites or by artificially limiting network performance.  Even then, it's not clear that these tricks will do that much to help.

But at least we're not really making things worse.

In writing this, I also realized that much of the text in this section on pooled sessions was out of date.  Using the network partition key means that we don't need to worry about pooling happening in cases where it should not.  Also, the mitigation wasn't going to be very effective at preventing cross-site recognition when the shared connection is all you need for that.

Closes #478.

***
<a href="https://pr-preview.s3.amazonaws.com/martinthomson/webtransport/pull/485.html" title="Last updated on Apr 26, 2023, 6:28 AM UTC (e5b2a79)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/485/b2b4b07...martinthomson:e5b2a79.html" title="Last updated on Apr 26, 2023, 6:28 AM UTC (e5b2a79)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2291,13 +2291,14 @@ Information about the network is available to the server either directly through
 its own networking stack, indirectly through the rate at which data is consumed
 or transmitted by the client, or as part of the statistics provided by the API
 (see [[#web-transport-stats]]).  Consequently, restrictions on information in
-user agents is unlikely to be the only way to manage privacy risks.
+user agents is not the only mechanism that might be needed to manage these
+privacy risks.
 
 ### Static Observations ### {#fp-static}
 
 A site can observe available network capacity or round trip time (RTT) between a
 user agent and a chosen server.  This information can be identifying when
-combined with other tracking vectors.  RTT can also reveal something of the
+combined with other tracking vectors.  RTT can also reveal something about the
 physical location of a user agent, especially if multiple measurements can be
 made from multiple vantage points.
 
@@ -2336,8 +2337,8 @@ session.  One session could infer information about the activity of another
 session, such as the rate at which another application is sending data.
 
 The use of a shared connection already allows the server to correlate sessions.
-Use of a [[fetch#network-partition-keys|network partition key]] disables pooling where use of a shared
-session might enable unwanted cross-site recognition.
+Use of a [[fetch#network-partition-keys|network partition key]] disables pooling
+where use of a shared session might enable unwanted cross-site recognition.
 
 
 # Examples #  {#examples}
--- a/index.bs
+++ b/index.bs
@@ -1584,7 +1584,7 @@ The dictionary SHALL have the following attributes:
    {{WebTransportSendStream}} has been sent at least once. This
    number can only increase, and is always less than or equal to
    {{WebTransportSendStreamStats/bytesWritten}}.
-   
+
    Note: this is progress of app data sent on a single stream only, and does not
    include any network overhead.
 : <dfn for="WebTransportSendStreamStats" dict-member>bytesAcknowledged</dfn>
--- a/index.bs
+++ b/index.bs
@@ -2262,21 +2262,85 @@ The WebTransport API lets the application specify multiple certificate
 hashes at once, allowing the client to accept multiple certificates for a
 period in which a new certificate is being rolled out.
 
-Unlike a similar mechanism in WebRTC, the server certificate hash API in
+Unlike a similar mechanism in [=WebRTC=], the server certificate hash API in
 WebTransport does not provide any means of authenticating the client; the fact
 that the client knows what the server certificate is or how to contact it is
 not sufficient.  The application has to establish the identity of the client
 in-band if necessary.
 
-## Pooled Sessions
+## Fingerprinting and Tracking ## {#fingerprinting}
 
-When pooling, if information from one session is affected by the activity of another
-session, one session could infer information about the activity of another session.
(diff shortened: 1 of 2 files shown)


jan-ivar:
@vasilvv to take a look.

index.bs
```diff
@@ -2262,21 +2262,83 @@ The WebTransport API lets the application specify multiple certificate
...
+WebTransport provides sites with an ability to generate network activity toward
+a server of their choice and observe the effects.  Observations of both the
+stable properties of a network path and dynamic effect of network usage are
+possible.
+
+Information about the network is available to the server either directly through
+its own networking stack, indirectly through the rate at which data is consumed
+or transmitted by the client, or as part of the statistics provided by the API
+(see [[#web-transport-stats]]).  Consequently, restrictions on information in
+user agents is unlikely to be the only way to manage privacy risks.
```
vasilvv, line 2294: I'm not entirely sure what "unlikely to be the only way to manage privacy risks" means?  "unlikely to be sufficient"?  "unlikely to be effective"?

index.bs
```diff
@@ -2262,21 +2262,83 @@ The WebTransport API lets the application specify multiple certificate
...
+
+### Static Observations ### {#fp-static}
+
+A site can observe available network capacity or round trip time (RTT) between a
+user agent and a chosen server.  This information can be identifying when
+combined with other tracking vectors.  RTT can also reveal something of the
```
vasilvv, line 2300: "something about"?

vasilvv (approved):
This is really good.

index.bs
martinthomson: I don't know what I was thinking when I wrote this...

```suggestion
user agents is not the only mechanism that might be needed to manage these privacy
risks.
```

index.bs
martinthomson:
```suggestion
combined with other tracking vectors.  RTT can also reveal something about the
```

index.bs
```diff
@@ -2262,21 +2262,83 @@ The WebTransport API lets the application specify multiple certificate
...
+### Pooled Sessions ### {#fp-pooled}
+
+Similar to shared networking scenarios, when sessions are pooled on a single
+connection, information from one session is affected by the activity of another
+session.  One session could infer information about the activity of another
+session, such as the rate at which another application is sending data.
+
+The use of a shared connection already allows the server to correlate sessions.
+Use of a [[fetch#network-partition-keys|network partition key]] disables pooling where use of a shared
+session might enable unwanted cross-site recognition.
```
martinthomson, line 2340:
```suggestion
Use of a [[fetch#network-partition-keys|network partition key]] disables pooling
where use of a shared session might enable unwanted cross-site recognition.
```

aboba (approved):

jan-ivar:
Please update to remove index.html file before merging.

Issue #480, opened 2023-02-24
Clarify whether spec supports BYOB readers for Datagrams


jesup:
The spec includes no text supporting the use of BYOB readers for Datagrams (while it does have BYOB text for WebTransportReceiveStreams, which are byte-streams).   Datagrams always traffic in Uint8Arrays, not bytes, so BYOB readers should not work with them, I presume.   Note also that the current Datagram WPT tests use BYOB readers....

If they somehow should be supported, there will need to be  number of spec changes I think.

ricea:
Sorry, this was a serious oversight by us (Chromium). We shipped BYOB support for datagrams without standardising the feature. We will prepare a PR for the spec ASAP.

jan-ivar:
> Datagrams always traffic in Uint8Arrays, not bytes, so BYOB readers should not work with them, I presume

The streams [model](https://streams.spec.whatwg.org/#model) says they _can be_ byte streams (_"a byte stream might contain chunks consisting of 16 KiB [Uint8Array](https://tc39.es/ecma262/#sec-typedarray-objects)s"_, but they [aren't](https://w3c.github.io/webtransport/#webtransport-constructor): _"20. [Set up](https://streams.spec.whatwg.org/#readablestream-set-up) incomingDatagrams ..."_ calls [set up](https://streams.spec.whatwg.org/#readablestream-set-up), not [set up with byte reading support](https://streams.spec.whatwg.org/#readablestream-set-up-with-byte-reading-support)...

> (Chromium) ... shipped BYOB support for datagrams ...

One concern is whether "[byte reading](https://streams.spec.whatwg.org/#example-manual-read-bytes)" semantics gel with preserving datagram boundaries.

But looking at the [WPT/webtransport/datagrams.https.any.js](https://github.com/web-platform-tests/wpt/blob/6ed09904d89a672406b28360081ee48f6fefca51/webtransport/datagrams.https.any.js#L145), it looks like too-small BYOB buffers reject, which is good:
```js
  // Write datagrams of size 10, but only 1 byte buffer is provided for BYOB
  // reader. To avoid splitting a datagram, stream will be errored.
  const buffer = new ArrayBuffer(1);
  const [error, _] = await Promise.all([
    reader.read(new Uint8Array(buffer)).catch(e => {
      controller.abort();
      return e;
    }),
    write_large_datagrams(writer, signal)
  ]);
  assert_equals(error.name, 'RangeError');
[...]

PR #487, opened 2023-03-03
Support BYOB Readers for Datagrams

nidhijaju:
This change adds support for BYOB readers for datagrams in the spec, similar to WebTransportReceiveStreams.

Fixes https://github.com/w3c/webtransport/issues/480.

***
<a href="https://pr-preview.s3.amazonaws.com/nidhijaju/webtransport/pull/487.html" title="Last updated on Apr 24, 2023, 4:14 AM UTC (3ffe59d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/487/b2b4b07...nidhijaju:3ffe59d.html" title="Last updated on Apr 24, 2023, 4:14 AM UTC (3ffe59d)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -490,9 +490,9 @@ To <dfn>pullDatagrams</dfn>, given a {{WebTransport}} object |transport|, run th
         with |error|.
     1. Return [=a promise rejected with=] |error|.
   1. Set |offset| to |view|.\[[ByteOffset]].
-  1. Set |maxBytes| to |view|'s [=BufferSource/byte length=].
+  1. Set |maxBytes| to the size of |datagram|.
   1. Set |buffer| to |view|'s [=BufferSource/underlying buffer=].
-  1. Set |chunk| to |view|.
+  1. Set |chunk| to a new {{Uint8Array}} object with |buffer|, |offset|, and the size of |datagram|.
 1. Otherwise:
   1. Set |offset| to 0.
   1. Set |maxBytes| to the size of |datagram|.
--- a/index.bs
+++ b/index.bs
@@ -494,7 +494,7 @@ To <dfn>pullDatagrams</dfn>, given a {{WebTransport}} object |transport|, run th
     1. Set |chunk| to |view|.
 1. Otherwise:
   1. Set |offset| to 0.
-  1. Set |maxBytes| to an [=implementation-defined=] size.
+  1. Set |maxBytes| to the size of |bytes|.
   1. Set |buffer| be a [=new=] {{ArrayBuffer}} with |maxBytes| size. If allocating the
      {{ArrayBuffer}} fails, return [=a promise rejected with=] a {{RangeError}}.
   1. Set |chunk| to a new {{Uint8Array}} object with |buffer| and |offset|.
@@ -766,6 +766,7 @@ agent MUST run the following steps:
 Note: Using 64kB buffers with datagrams is recommended because the effective
 maximum WebTransport datagram frame size has an upper bound of the QUIC maximum datagram frame size
 which is recommended to be 64kB (See [[!QUIC-DATAGRAM]] [Section 3](https://datatracker.ietf.org/doc/html/rfc9221#section-3)).
+This will ensure the stream is not errored due to a datagram being larger than the buffer.
 
 1. [=ReadableStream/Set up with byte reading support=] |incomingDatagrams| with
    [=ReadableStream/set up with byte reading support/pullAlgorithm=] set to
--- a/index.bs
+++ b/index.bs
@@ -484,19 +484,12 @@ To <dfn>pullDatagrams</dfn>, given a {{WebTransport}} object |transport|, run th
    [=ReadableStream/current BYOB request view=] is not null, then:
   1. Let |view| be |datagrams|.{{WebTransportDatagramDuplexStream/[[Readable]]}}'s
      [=ReadableStream/current BYOB request view=].
-  1. If |view|'s [=BufferSource/byte length=] is less than the size of |datagram|, then:
-    1. Let |error| be a new {{RangeError}} exception.
-    1. [=ReadableStream/Error=] |datagrams|.{{WebTransportDatagramDuplexStream/[[Readable]]}}
-        with |error|.
-    1. Return [=a promise rejected with=] |error|.
+  1. If |view|'s [=BufferSource/byte length=] is less than the size of |datagram|, return
+     [=a promise rejected with=] a {{RangeError}}.
   1. Let |elementSize| be the element size specified in [=the typed array constructors table=] for
      |view|.\[[TypedArrayName]]. If |view| does not have a \[[TypedArrayName]] internal slot
      (i.e. it is a {{DataView}}), let |elementSize| be 0.
-  1. If |elementSize| is not 1, then:
-    1. Let |error| be a new {{TypeError}} exception.
-    1. [=ReadableStream/Error=] |datagrams|.{{WebTransportDatagramDuplexStream/[[Readable]]}}
-        with |error|.
-    1. Return [=a promise rejected with=] |error|.
+  1. If |elementSize| is not 1, return [=a promise rejected with=] a {{TypeError}}.
 1. [=ReadableStream/Pull from bytes=] |datagram| into |datagrams|.{{WebTransportDatagramDuplexStream/[[Readable]]}}.
 1. Return [=a promise resolved with=] undefined.
 
--- a/index.bs
+++ b/index.bs
@@ -478,13 +478,13 @@ To <dfn>pullDatagrams</dfn>, given a {{WebTransport}} object |transport|, run th
 1. If |queue| is empty, then:
   1. Set |datagrams|.{{[[IncomingDatagramsPullPromise]]}} to a new promise.
   1. Return |datagrams|.{{[[IncomingDatagramsPullPromise]]}}.
-1. Let |bytes| and |timestamp| be the result of [=dequeuing=] |queue|.
+1. Let |datagram| and |timestamp| be the result of [=dequeuing=] |queue|.
 1. Let |buffer|, |offset|, |maxBytes|, and |chunk| be null.
 1. If |datagrams|.{{WebTransportDatagramDuplexStream/[[Readable]]}}'s
    [=ReadableStream/current BYOB request view=] is not null, then:
   1. Let |view| be |datagrams|.{{WebTransportDatagramDuplexStream/[[Readable]]}}'s
      [=ReadableStream/current BYOB request view=].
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -450,8 +450,27 @@ To <dfn>pullDatagrams</dfn>, given a {{WebTransport}} object |transport|, run th
...
+     [=ReadableStream/error=] |datagrams|.{{WebTransportDatagramDuplexStream/[[Readable]]}}
+     with a {{RangeError}}.
+  1. Otherwise:
+    1. Set |offset| to |view|.\[[ByteOffset]].
+    1. Set |maxBytes| to |view|'s [=BufferSource/byte length=].
+    1. Set |buffer| to |view|'s [=BufferSource/underlying buffer=].
+    1. Set |chunk| to |view|.
+1. Otherwise:
+  1. Set |offset| to 0.
+  1. Set |maxBytes| to an [=implementation-defined=] size.
```
ricea, line 468: I think the size should probably just be |bytes|, unless there's a reason why some implementations would need it to be larger.

index.bs
```diff
@@ -725,8 +744,14 @@ agent MUST run the following steps:
     :: null
 1. Let |pullDatagramsAlgorithm| be an action that runs [=pullDatagrams=] with |transport|.
 1. Let |writeDatagramsAlgorithm| be an action that runs [=writeDatagrams=] with |transport|.
-1. [=ReadableStream/Set up=] |incomingDatagrams| with [=ReadableStream/set up/pullAlgorithm=]
-   set to |pullDatagramsAlgorithm|, and [=ReadableStream/set up/highWaterMark=] set to 0.
+
+Note: Using 64kB buffers with datagrams is recommended because the effective
+maximum WebTransport datagram frame size has an upper bound of the QUIC maximum datagram frame size
+which is recommended to be 64kB (See [[!QUIC-DATAGRAM]] [Section 3](https://datatracker.ietf.org/doc/html/rfc9221#section-3)).
```
ricea, line 750: Suggestion: Add "This will ensure the stream is not errored due to a datagram being larger than the buffer."

index.bs
```diff
@@ -465,7 +465,7 @@ To <dfn>pullDatagrams</dfn>, given a {{WebTransport}} object |transport|, run th
     1. Set |chunk| to |view|.
 1. Otherwise:
   1. Set |offset| to 0.
-  1. Set |maxBytes| to an [=implementation-defined=] size.
+  1. Set |maxBytes| to the size of |bytes|.
```
ricea, line 468: I think this can actually just be
```
1. Set |maxBytes| to |bytes|.
```

ricea (approved):
lgtm

aboba (approved):

jan-ivar:
@jesup to have a look, and if everything looks good, then this is `Editors can integrate`.

index.bs
```diff
@@ -450,8 +450,27 @@ To <dfn>pullDatagrams</dfn>, given a {{WebTransport}} object |transport|, run th
...
+    1. Set |buffer| to |view|'s [=BufferSource/underlying buffer=].
+    1. Set |chunk| to |view|.
+1. Otherwise:
+  1. Set |offset| to 0.
+  1. Set |maxBytes| to the size of |bytes|.
+  1. Set |buffer| be a [=new=] {{ArrayBuffer}} with |maxBytes| size. If allocating the
+     {{ArrayBuffer}} fails, return [=a promise rejected with=] a {{RangeError}}.
+  1. Set |chunk| to a new {{Uint8Array}} object with |buffer| and |offset|.
+1. Write |bytes| into |buffer| with offset |offset|, up to |maxBytes| bytes.
+1. [=ReadableStream/Enqueue=] |chunk| to |datagrams|.{{WebTransportDatagramDuplexStream/[[Readable]]}}.
```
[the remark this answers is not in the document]
saschanaz: But it might be surprising to always get Uint8Array if the reader receives Uint16Array for byob for example. The caller need to check the type of view anyway to properly retrieve non-byob chunks, so maybe not too much troublesome?

cc @domenic

index.bs
ricea: Sorry, I was wrong. The original text was correct.

index.bs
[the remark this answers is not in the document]
saschanaz: > What I'd like to happen is the if you pass a typed array with `BYTES_PER_ELEMENT > 1` to `read(view)` then it will error the stream. I'm not sure if the Streams standard exposes enough information to let us do that.

Should be possible as "current BYOB request view" gives a JS typed array object which should have that information. But the exception can only happen from the pull callback, not immediately from the read call. 🤔

saschanaz:
[The Streams spec says](https://streams.spec.whatwg.org/#readablestream-current-byob-request-view):

>They should only [create](https://webidl.spec.whatwg.org/#arraybufferview-create) a new [ArrayBufferView](https://webidl.spec.whatwg.org/#ArrayBufferView) to pass to [enqueue](https://streams.spec.whatwg.org/#readablestream-enqueue) when the [current BYOB request view](https://streams.spec.whatwg.org/#readablestream-current-byob-request-view) is null, or when they have more bytes on hand than the [current BYOB request view](https://streams.spec.whatwg.org/#readablestream-current-byob-request-view)'s [byte length](https://webidl.spec.whatwg.org/#buffersource-byte-length).

But we are instead throwing an exception here if the view is too small. Should it allocate a new Uint8Array?

index.bs
jan-ivar: I think this suggests the name `bytes` is confusing here now that we're doing BYOB. [[[IncomingDatagramsQueue]]](https://pr-preview.s3.amazonaws.com/nidhijaju/webtransport/pull/487.html#dom-webtransportdatagramduplexstream-incomingdatagramsqueue-slot) says it's a "datagram" which might be a better term to help separate inputs from outputs, or even "incomingDatagram"?

index.bs
```diff
@@ -450,8 +450,27 @@ To <dfn>pullDatagrams</dfn>, given a {{WebTransport}} object |transport|, run th
...
-1. [=ReadableStream/Enqueue=] |chunk| to |transport|.{{[[Datagrams]]}}.{{WebTransportDatagramDuplexStream/[[Readable]]}}.
+1. Let |buffer|, |offset|, |maxBytes|, and |chunk| be null.
+1. If |datagrams|.{{WebTransportDatagramDuplexStream/[[Readable]]}}'s
+   [=ReadableStream/current BYOB request view=] is not null, then:
+  1. Let |view| be |datagrams|.{{WebTransportDatagramDuplexStream/[[Readable]]}}'s
+     [=ReadableStream/current BYOB request view=].
+  1. If |view|'s [=BufferSource/byte length=] is less than the size of |bytes|,
+     [=ReadableStream/error=] |datagrams|.{{WebTransportDatagramDuplexStream/[[Readable]]}}
+     with a {{RangeError}}.
+  1. Otherwise:
```
jan-ivar, line 461: This needs to bail, otherwise the algorithm proceeds to "Write |bytes| into" `null` below. Unlike with _throw_, which implicitly does this, I don't think we can assume abort to be implicit here, at least it seems unclear to me, especially with the "Otherwise" directly following this statement.

I suggest something like, "and abort all steps", since we're technically inside nested steps here.

This should let us remove this "Otherwise" as well, and help flatten things.

index.bs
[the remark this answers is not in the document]
domenic: It's not clear to me from this conversation if https://github.com/whatwg/streams/pull/1263 suffices for your case. It'd be lovely if it did.

I plan to merge that PR as-is. If there are improvements we could make to it that would make it usable by both Fetch and WebTransport, we can fast-follow with such updates.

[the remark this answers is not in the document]
ricea: > I plan to merge that PR as-is. If there are improvements we could make to it that would make it usable by both Fetch and WebTransport, we can fast-follow with such updates.

sgtm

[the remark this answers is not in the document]
saschanaz: >It's not clear to me from this conversation if whatwg/streams#1263 suffices for your case. It'd be lovely if it did.

IMO it does, except the type and size of the BYOB view should be checked before calling it. That shouldn't be problematic.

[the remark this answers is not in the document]
saschanaz: Can we have an ISSUE paragraph here at least?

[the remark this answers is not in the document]
nidhijaju: > IMO it does, except the type and size of the BYOB view should be checked before calling it.

I tried doing this and fixing the issue in the latest changes. Would you mind taking a look again, @saschanaz? Thank you!

index.bs
```diff
@@ -478,9 +479,25 @@ To <dfn>pullDatagrams</dfn>, given a {{WebTransport}} object |transport|, run th
...
+1. Let |datagram| and |timestamp| be the result of [=dequeuing=] |queue|.
+1. If |datagrams|.{{WebTransportDatagramDuplexStream/[[Readable]]}}'s
+   [=ReadableStream/current BYOB request view=] is not null, then:
+  1. Let |view| be |datagrams|.{{WebTransportDatagramDuplexStream/[[Readable]]}}'s
+     [=ReadableStream/current BYOB request view=].
+  1. If |view|'s [=BufferSource/byte length=] is less than the size of |datagram|, then:
+    1. Let |error| be a new {{RangeError}} exception.
+    1. [=ReadableStream/Error=] |datagrams|.{{WebTransportDatagramDuplexStream/[[Readable]]}}
+        with |error|.
+    1. Return [=a promise rejected with=] |error|.
```
saschanaz, line 491: Wouldn't this promise-returning step do the erroring above automatically?

index.bs
```diff
@@ -478,9 +479,25 @@ To <dfn>pullDatagrams</dfn>, given a {{WebTransport}} object |transport|, run th
...
+  1. Let |elementSize| be the element size specified in [=the typed array constructors table=] for
+     |view|.\[[TypedArrayName]]. If |view| does not have a \[[TypedArrayName]] internal slot
+     (i.e. it is a {{DataView}}), let |elementSize| be 0.
+  1. If |elementSize| is not 1, then:
+    1. Let |error| be a new {{TypeError}} exception.
+    1. [=ReadableStream/Error=] |datagrams|.{{WebTransportDatagramDuplexStream/[[Readable]]}}
+        with |error|.
+    1. Return [=a promise rejected with=] |error|.
```
saschanaz, line 499: Same here.

saschanaz (approved):
LGTM, thanks!

nidhijaju:
Hi @jesup, friendly ping but would you be able to take a look at this as well?

jan-ivar:
`editors can integrate` pending @jesup approves.

jesup:
LGTM

Issue #456, opened 2023-01-11
serverCertificateHashes


jesup:
Comment from someone else at Mozilla: It seems weird that the handling of serverCertificateHashes is specified only as a side effect in Section 5.9.  There's no specification of _where_ the NotSupportedEerror or trust errors should occur.

jan-ivar:
Seems ready for PR. Is there a simple test to check which order Chrome uses?

Issue #359, opened 2021-09-24
"compute a certificate hash" / "verify a certificate hash" should have explicit parameters


yutakahirano:
Brought up at #358.

cc: @martinthomson

PR #489, opened 2023-03-10
Throwing serverCertificateHashes related errors

nidhijaju:
This change specifies _where_ the NotSupportedError and trust errors should be thrown when handling [serverCertificateHashes](https://w3c.github.io/webtransport/#dom-webtransportoptions-servercertificatehashes). It also adds explicit parameters to the certificate hash algorithms.

Fixes https://github.com/w3c/webtransport/issues/456 and fixes https://github.com/w3c/webtransport/issues/359.

***
<a href="https://pr-preview.s3.amazonaws.com/nidhijaju/webtransport/pull/489.html" title="Last updated on Apr 27, 2023, 1:33 AM UTC (e950163)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/489/85fc448...nidhijaju:e950163.html" title="Last updated on Apr 27, 2023, 1:33 AM UTC (e950163)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -792,7 +792,7 @@ sequence&lt;{{WebTransportHash}}&gt; |serverCertificateHashes|, run these steps.
      |networkPartitionKey|, |url|, false, |newConnection|, and |http3Only|. If the user agent
      supports more than one congestion control algorithm, choose one appropriate for
      |congestionControl| for sending of data on this |connection|. When obtaining a connection, if
-     |serverCertificateHashes| is specified instead of the default certificate version algorithm,
+     |serverCertificateHashes| is specified instead of the default certificate verification algorithm,
      validate the certificate against [=custom certificate requirements=], and then
      [=verify a certificate hash|verify the certificate hash=] against |serverCertificateHashes|.
      The certificate is considered valid if and only if both of those checks pass.
@@ -1192,11 +1192,10 @@ To <dfn>verify a certificate hash</dfn>, given a |certificate| and an array of h
 perform the following steps:
 1. Let |referenceHash| be the result of [=computing a certificate hash=] with |certificate|.
 1. For every hash |hash| in |hashes|:
-   1. If |hash|.{{WebTransportHash/value}} is not null:
+   1. If |hash|.{{WebTransportHash/value}} is not null and |hash|.{{WebTransportHash/algorithm}}
+      is an [=ASCII case-insensitive=] match with "sha-256":
      1. Let |hashValue| be the byte sequence which |hash|.{{WebTransportHash/value}} represents.
-     1. If |hash|.{{WebTransportHash/algorithm}} is an [=ASCII case-insensitive=] match with
-        "sha-256", and |hashValue| is equal to |referenceHash|, the |certificate| is valid.
-        Return true.
+     1. If |hashValue| is equal to |referenceHash|, return true.
 1. Return false.
 
 </div>
--- a/index.bs
+++ b/index.bs
@@ -795,7 +795,7 @@ sequence&lt;{{WebTransportHash}}&gt; |serverCertificateHashes|, run these steps.
      |serverCertificateHashes| is specified instead of the default certificate version algorithm,
      validate the certificate against [=custom certificate requirements=], and then
      [=verify a certificate hash|verify the certificate hash=] against |serverCertificateHashes|.
-     The certificate is considered valid only if both of those checks pass.
+     The certificate is considered valid if and only if both of those checks pass.
   1. If |connection| is failure, then abort the remaining steps and [=queue a network task=] with
      |transport| to run these steps:
     1. If |transport|.{{[[State]]}} is `"closed"` or `"failed"`, then abort these steps.
--- a/index.bs
+++ b/index.bs
@@ -802,7 +802,9 @@ sequence&lt;{{WebTransportHash}}&gt; |serverCertificateHashes|, run these steps.
      |serverCertificateHashes| returns false or the certificate does not satisfy [=custom certificate requirements=],
      then abort the remaining steps and [=queue a network task=] with |transport| to run these steps:
     1. If |transport|.{{[[State]]}} is `"closed"` or `"failed"`, then abort these steps.
-    1. [=Cleanup=] |transport| with a {{SecurityError}}.
+    1. Let |error| be a newly [=DOMException/created=] {{WebTransportError}} whose
+       {{WebTransportErrorOptions/source}} is `"session"`.
+    1. [=Cleanup=] |transport| with |error|.
   1. Wait for |connection| to receive the first SETTINGS frame, and let |settings| be a dictionary that
      represents the SETTINGS frame.
   1. If |settings| doesn't contain SETTINGS_ENABLE_WEBTRANPORT with a value of 1, or it doesn't
--- a/index.bs
+++ b/index.bs
@@ -1197,7 +1197,7 @@ perform the following steps:
      1. Let |hashValue| be the byte sequence which |hash|.{{WebTransportHash/value}} represents.
      1. If |hash|.{{WebTransportHash/algorithm}} is an [=ASCII case-insensitive=] match with
         "sha-256", and |hashValue| is equal to |referenceHash|, the |certificate| is valid.
-	Return true.
+        Return true.
 1. Return false.
 
 </div>
--- a/index.bs
+++ b/index.bs
@@ -795,6 +795,7 @@ sequence&lt;{{WebTransportHash}}&gt; |serverCertificateHashes|, run these steps.
      |serverCertificateHashes| is specified instead of the default certificate version algorithm,
      validate the certificate against [=custom certificate requirements=], and then
      [=verify a certificate hash|verify the certificate hash=] against |serverCertificateHashes|.
+     The certificate is considered valid only if both of those checks pass.
   1. If |connection| is failure, then abort the remaining steps and [=queue a network task=] with
      |transport| to run these steps:
     1. If |transport|.{{[[State]]}} is `"closed"` or `"failed"`, then abort these steps.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1169,22 +1175,23 @@ that determine how WebTransport connection is established and used.
...
+perform the following steps:
+1. Let |referenceHash| be the result of [=computing a certificate hash=] with |certificate|.
 1. For every hash |hash| in |hashes|:
    1. If |hash|.{{WebTransportHash/value}} is not null:
      1. Let |hashValue| be the byte sequence which |hash|.{{WebTransportHash/value}} represents.
-     1. If {{WebTransportHash/algorithm}} of |hash| is an [=ASCII case-insensitive=] match with "sha-256", and |hashValue| is equal
-        to |referenceHash|, the certificate is valid. Return true.
+     1. If |hash|.{{WebTransportHash/algorithm}} is an [=ASCII case-insensitive=] match with
+        "sha-256", and |hashValue| is equal to |referenceHash|, the |certificate| is valid.
+        Return true.
```
ricea, line 1194:
```suggestion
        "sha-256", and |hashValue| is equal to |referenceHash|, return true.
```

ricea (commented):
lgtm

jan-ivar:
Meeting:
- victor to provide comments:
   - seems to be in the wrong place but no better place at the time
   - don't reveal info ahead of acceptance of connection

index.bs
```diff
@@ -779,6 +780,11 @@ To <dfn>initialize WebTransport over HTTP</dfn>, given a {{WebTransport}} object
     1. Let |error| be a newly [=DOMException/created=] {{WebTransportError}} whose
        {{WebTransportErrorOptions/source}} is `"session"`.
     1. [=Cleanup=] |transport| with |error|.
+  1. If [=verify a certificate hash|verifying the certificate hash=] from |connection| against
+     |serverCertificateHashes| returns false or the certificate does not satisfy [=custom certificate requirements=],
+     then abort the remaining steps and [=queue a network task=] with |transport| to run these steps:
+    1. If |transport|.{{[[State]]}} is `"closed"` or `"failed"`, then abort these steps.
+    1. [=Cleanup=] |transport| with a {{SecurityError}}.
```
vasilvv, line 787: Note that we always have to return a generic "handshake failed" WebTransportError before the handshake succeeds in order to prevent network scanning (see [here](https://w3c.github.io/webtransport/#protocol-security)).

index.bs
```diff
@@ -779,6 +780,11 @@ To <dfn>initialize WebTransport over HTTP</dfn>, given a {{WebTransport}} object
     1. Let |error| be a newly [=DOMException/created=] {{WebTransportError}} whose
        {{WebTransportErrorOptions/source}} is `"session"`.
     1. [=Cleanup=] |transport| with |error|.
+  1. If [=verify a certificate hash|verifying the certificate hash=] from |connection| against
```
vasilvv, line 783: I'm not sure this is a good place to put it.  This is not a separate step, but rather a modification of `obtaining a connection` step (which is where cert check would normally happen).  The problem is, `obtaining a connection` does not actually accept an argument that would let you modify certificate verification, nor does it talk about it at all, so I'm afraid we would have to be somewhat vague about this.

nidhijaju: Do you think it makes sense to move this to the end of Step 7.2, or is there a better place you would suggest?

vasilvv: Yes, something along the lines of "When obtaining a connection, if serverCertificatHashes is specified, instead of the default certificate version algorithm, validate the certificate against custom cert requirements and then validate it against the provided hashes".

nidhijaju: Got it, thank you!

index.bs
```diff
@@ -790,7 +791,10 @@ To <dfn>initialize WebTransport over HTTP</dfn>, given a {{WebTransport}} object
   1. Let |connection| be the result of [=obtain a connection|obtaining a connection=] with
      |networkPartitionKey|, |url|, false, |newConnection|, and |http3Only|. If the user agent
      supports more than one congestion control algorithm, choose one appropriate for
-     |congestionControl| for sending of data on this |connection|.
+     |congestionControl| for sending of data on this |connection|. When obtaining a connection, if
+     |serverCertificateHashes| is specified instead of the default certificate version algorithm,
+     validate the certificate against [=custom certificate requirements=], and then
+     [=verify a certificate hash|verify the certificate hash=] against |serverCertificateHashes|.
```
vasilvv, line 797: I would clarify that certificate should be considered valid if and only if _both_ of those checks pass.

vasilvv (approved):

jan-ivar:
@martinthomson to have a look.

index.bs
martinthomson:
```suggestion
        "sha-256", and |hashValue| is equal to |referenceHash|, the |certificate| is valid.
     1. Return true.
```
Either this, or what Adam said.

index.bs
```diff
@@ -790,7 +791,11 @@ To <dfn>initialize WebTransport over HTTP</dfn>, given a {{WebTransport}} object
...
+     The certificate is considered valid if and only if both of those checks pass.
```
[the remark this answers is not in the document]
nidhijaju: I was referring to both the certificate meeting the custom certificate requirements and verifying the hash against the `serverCertificateHashes`.

I tried rephrasing the sentences, and mentioning that the connection should have failed if either of those conditions are not met, but does it look better?

index.bs
```diff
@@ -790,7 +791,12 @@ To <dfn>initialize WebTransport over HTTP</dfn>, given a {{WebTransport}} object
   1. Let |connection| be the result of [=obtain a connection|obtaining a connection=] with
      |networkPartitionKey|, |url|, false, |newConnection|, and |http3Only|. If the user agent
      supports more than one congestion control algorithm, choose one appropriate for
-     |congestionControl| for sending of data on this |connection|.
+     |congestionControl| for sending of data on this |connection|. When obtaining a connection, if
+     |serverCertificateHashes| is specified, instead of using the default certificate verification
+     algorithm, consider the certificate valid if it meets the [=custom certificate
+     requirements=] and if [=verify a certificate hash|verifying the certificate hash=] against
+     |serverCertificateHashes| returns true. If either condition is not met, let |connection| be
+     failure.
```
martinthomson, line 799: This bit about `serverCertificateHashes` might be better as a substep:

```
1. If |serverCertificateHashes| is specified, run the following steps:
    1.  If the certificate does not meet the [=custom certificate requirements=], mark the |connection| as failed.
    1. If the [=verify a certificate hash|verifying the certificate hash=] against |serverCertificateHashes| returns false, mark the |connection| as failed.
```

index.bs
```diff
@@ -790,7 +791,12 @@ To <dfn>initialize WebTransport over HTTP</dfn>, given a {{WebTransport}} object
...
   1. If |connection| is failure, then abort the remaining steps and [=queue a network task=] with
```
martinthomson, line 800: The phrasing here is a bit odd.  Maybe `if |connection| failed for any reason`.

index.bs
nidhijaju: I actually initially had this as substeps in earlier commits, but based on https://github.com/w3c/webtransport/pull/489#discussion_r1163085904, I changed it to be part of the same step. Since these steps are a modification of the "[obtain a connection](https://fetch.spec.whatwg.org/#concept-connection-obtain)" step in Fetch, I think we should keep this part of the same step so as to make it clear that these checks happen _during_ the "obtain a connection" step, as opposed to after.

index.bs
nidhijaju: "[Obtain a connection](https://fetch.spec.whatwg.org/#concept-connection-obtain)" actually returns "failure" if the connection fails, which is what this step is checking.

index.bs
martinthomson: I don't think that conceptualizing these as part of "obtain a connection" helps.  It is cleaner if you regard these as checks that occur after a connection was successfully obtained.

index.bs
martinthomson: Ah, I see what you are doing.  This is like a Rust `Result<T, E>` where you get a composite object that is either success (a connection) or an error.  You are checking if the result was an error.

The problem I have is that - while this fine if you are writing code, this is terribly stilted prose.  It reads as a grammatical error.

index.bs
nidhijaju: I slightly prefer keeping it part of the same step, but either seems workable, so I'll update it to whatever seems best to you and @vasilvv.

index.bs
nidhijaju: Yes, while I agree that it sounds odd, I think this is a fairly standard pattern throughout the spec ecosystem. For example, step 2 of https://url.spec.whatwg.org/#concept-url-parser and step 2 of https://encoding.spec.whatwg.org/#dom-textdecoder have similar text saying "if _x_ is failure". There's also an open issue about clarifying what "failure" is at https://github.com/whatwg/infra/issues/545, so perhaps when that is fixed, we can update it in the WebTransport spec.

index.bs
vasilvv: > I don't think that conceptualizing these as part of "obtain a connection" helps. It is cleaner if you regard these as checks that occur after a connection was successfully obtained.

I'm not sure I understand why this would be a separate check, since presumably you would want to perform it at the place you would normally perform certificate validation (as soon as you receive a certificate), which is in the middle of obtaining a connection (e.g. there might be steps after, like sending H/3 settings; also from error handling perspective, that means that a failure against certificate hash list is equivalent to failure against Web PKI in terms of how it's reported to the application).

ekinnear: It does seem like it ends up in a different subsystem and is something that is done during connection configuration at setup. 

In other words, in code, this isn't something that we check afterwards, but rather something that we configure as part of setting up a connection object and starting it.

Issue #484, opened 2023-03-01
Add sendRateEstimate to getStats()


jan-ivar:
From https://github.com/w3c/webtransport/issues/481#issuecomment-1449126622
> conclusion: change it to include sendRateEstimate in getStats()

wilaw:
Implementation should address questions raised in #476

pthatcher:
Do we really want major app behaviors like rate control to be based on stats?  I don't really care so much about whether it is or not; I just care about the value being in there at all.  But it does seem inconsistent with the general purpose of "stats".

BTW, Did I miss a meeting where this was discussed and decided on?

jan-ivar:
This was discussed and decided at the [Feb 28 meeting](https://www.w3.org/wiki/WebTransport/Meetings2023#WebTransport_Bi-weekly_Virtual_Meeting_.2355_late_-_Feb_28.2C_2023) (minutes and recording). But we can discuss it at again next meeting if you'd like.

I think the general sentiment was that the app could poll for these metrics. Can you elaborate on the difference from stats? E.g. WebRTC stats has [targetBitrate](https://www.w3.org/TR/2020/CR-webrtc-stats-20200114/#dom-rtcoutboundrtpstreamstats-targetbitrate).

pthatcher:
I have a PR: https://github.com/w3c/webtransport/pull/494

PR #494, opened 2023-03-29
Add estimatedSendRate to stats.

pthatcher:
Fixes https://github.com/w3c/webtransport/issues/484.

***
<a href="https://pr-preview.s3.amazonaws.com/pthatcher/webtransport/pull/494.html" title="Last updated on Apr 11, 2023, 1:41 PM UTC (3c7ed34)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/494/5e8422f...pthatcher:3c7ed34.html" title="Last updated on Apr 11, 2023, 1:41 PM UTC (3c7ed34)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1280,7 +1280,7 @@ dictionary WebTransportStats {
   DOMHighResTimeStamp rttVariation;
   DOMHighResTimeStamp minRtt;
   WebTransportDatagramStats datagrams;
-  unsigned long long estimatedSendRate;
+  unsigned long long? estimatedSendRate;
 };
 </pre>
 
--- a/index.bs
+++ b/index.bs
@@ -1320,9 +1320,9 @@ The dictionary SHALL have the following attributes:
    This rate applies to all streams and datagrams that share a [=WebTransport session=]
    and is calculated by the congestion control algorithm (potentially chosen by
    {{WebTransport/congestionControl}}). If this [=WebTransport session=] is
-   pooled with others in a shared [=connection=], the value is undefined.
-   If the user agent does not have an estimate, the value is undefined.
-   The value change from defined to undefined or undefined to defined at any time.
+   pooled with others in a shared [=connection=], or if the user agent does not
+   currently have an estimate, the member MUST be [=map/exist|absent=].
+   The member can be absent even if present in previous results.
 
 
 ## `WebTransportDatagramStats` Dictionary ##  {#web-transport-stats}
--- a/index.bs
+++ b/index.bs
@@ -1280,6 +1280,7 @@ dictionary WebTransportStats {
   DOMHighResTimeStamp rttVariation;
   DOMHighResTimeStamp minRtt;
   WebTransportDatagramStats datagrams;
+  unsigned long long estimatedSendRate;
 };
 </pre>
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1314,6 +1315,15 @@ The dictionary SHALL have the following attributes:
...
 : <dfn for="WebTransportStats" dict-member>minRtt</dfn>
 :: The minimum round-trip time observed on the entire connection.
+: <dfn for="WebTransportStats" dict-member>estimatedSendRate</dfn>
+:: The estimated rate at which queued data will be sent by the user agent, in bits per second.
+   This rate applies to all streams and datagrams that share a [=WebTransport session=]
+   and is calculated by the congestion control algorithm (potentially chosen by
+   {{WebTransport/congestionControl}}). If this [=WebTransport session=] is
+   pooled with others in a shared [=connection=], the value is undefined.
+   If the user agent does not have an estimate, the value is undefined.
+   The value change from defined to undefined or undefined to defined at any time.
```
jan-ivar, line 1325: For dictionary members, it's best practice to test for [existence](https://infra.spec.whatwg.org/#map-exists) (lack of existence only produces `undefined` in JS). Keeping the current semantics, this would be:
```suggestion
   pooled with others in a shared [=connection=], or if the user agent does not
   currently have an estimate, the member MUST be [=map/exist|absent=].
   The member can be absent even if present in previous results.
```
...or if we go with nullable (which I recommend), we could write:
```suggestion
   pooled with others in a shared [=connection=], the value is `null`.
   If the user agent does not have an estimate, the value is `null`.
   The value can be `null` even if it were non-null in previous results.
```

index.bs
```diff
@@ -1314,6 +1315,15 @@ The dictionary SHALL have the following attributes:
    [Section 5.3](https://www.rfc-editor.org/rfc/rfc9002#section-5.3).
 : <dfn for="WebTransportStats" dict-member>minRtt</dfn>
 :: The minimum round-trip time observed on the entire connection.
+: <dfn for="WebTransportStats" dict-member>estimatedSendRate</dfn>
+:: The estimated rate at which queued data will be sent by the user agent, in bits per second.
+   This rate applies to all streams and datagrams that share a [=WebTransport session=]
+   and is calculated by the congestion control algorithm (potentially chosen by
+   {{WebTransport/congestionControl}}). If this [=WebTransport session=] is
+   pooled with others in a shared [=connection=], the value is undefined.
```
jan-ivar, line 1323: I gather this is to satisfy @martinthomson's concern in https://github.com/w3c/webtransport/pull/421#discussion_r1107885111?

This might limit media ingest to non-pooled connections. Are we OK with that? Is there no strategy a user agent could employ instead (e.g. divide 50/50 and omit other traffic from the result)?

Or perhaps this is fine?

martinthomson: To start with, this speaks to the feature detection angle more strongly.

The primary concern here is with cross-origin WebTransport sessions.  That is, when example.com establishes a session toward https://wt.example/upload and wants to know stuff about that connection.  If that connection is used to load stuff for other purposes that wt.example might have (within the same top-level browsing context), then information about size and timing might leak through.  We might assume that example.com is not already in a position to observe that much and so adding this new information would be irresponsible.

After getting more comfortable with my own understanding of this situation, I believe that it is already the case that sites that can make queries of cross-site content can learn stuff about how the connection is being used.  So we're just changing the fidelity of the measurements and maybe changing/constraining how the browser might be able to act to protect that information from observation.  For instance, in the current regime, a browser could isolate requests from different sources in time, so that requests never contend for capacity, at the cost of timing contention; WebTransport might make that sort of trick harder to pull off.

I'm almost at the point where I would instead say that this requirement can be removed, though perhaps we can do that in a follow-up.

jan-ivar:
Meeting:
- Merging this and filing a new issue to discuss pooling.

Issue #326, opened 2021-08-11
Determine interaction with back forward cache


ricea:
We need to specify the behaviour when the page is navigated away and the document is placed in the browser's back forward cache. Also what happens if/when the page is restored from the cache.

rakina:
Thanks for filing this! The BFCache team and the Web Transport team discussed about this earlier but haven't reached any conclusions on what to do with existing WebTransport connections when the page is put into BFCache.

As it is currently specced, it will go through the [unloading steps](https://w3c.github.io/webtransport/#web-transport-context-cleanup-steps) which will fail existing connections (CMIIW). We might want to change this to be more useful/powerful if the page enters BFCache (loses full activity), e.g. maybe keep the connection in some form.

In case it might be useful, here's a link to the TAG design principles PR: https://github.com/w3ctag/design-principles/pull/317
Also we have a BFCache meta-bug in case you need examples of past spec issues: https://github.com/whatwg/html/issues/5880

(cc @fergald)

jan-ivar:
I think it's desirable and POLA to close connections when navigating away from a page. The BFCache is a user agent optimization that some browsers do, but we should consider that most pages that go on it are never to be heard from again.

 The current behavior matches [webrtc](https://w3c.github.io/webrtc-pc/#dfn-ice-connection-state) and [websocket](https://html.spec.whatwg.org/multipage/browsing-the-web.html#unloading-document-cleanup-steps) which seems right to me.

> Also what happens if/when the page is restored from the cache.

It stays closed?

fergald:
> > Also what happens if/when the page is restored from the cache.
> 
> It stays closed?

Ideally the close event was delivered when the page comes back out of BFCache. My understanding is that the close event is not delivered when the connections are closed due to navigating away, so there would be no change for the never-restored case.

jan-ivar:
Meeting
- Agreement to leave session in a terminal state, but reject promises only on revival from BFCache.

rakina:
Just want to chime in that running `cleanup()` in "[unloading document cleanup steps]"(https://html.spec.whatwg.org/multipage/document-lifecycle.html#unloading-document-cleanup-steps) is probably a good thing to do. Not sure if it's needed, but if there are things that needs to be done on navigation back to the page (BFCache restore) you can also listen to that through adding a step in "[reactivate a document](https://html.spec.whatwg.org/multipage/browsing-the-web.html#reactivate-a-document)"

jesup:
[...]

PR #500, opened 2023-04-25
Specify BFCache interaction

nidhijaju:
The proposal for WebTransport interacting with BFCache is that we allow pages using WebTransport to enter the BFCache, but close the connection when navigating away. Then, the various promises will resolve if/when the page is navigated back to. 

The spec currently already terminates the session in the [unloading document cleanup steps](https://www.w3.org/TR/webtransport/#web-transport-context-cleanup-steps), but this PR adds a step to [queue a network task](https://www.w3.org/TR/webtransport/#webtransport-queue-a-network-task) to run [cleanup](https://www.w3.org/TR/webtransport/#webtransport-cleanup) on a WebTransport object. The network task queue will be paused when the page is navigated away from, and resumed when the page is navigated back to.

Fixes https://github.com/w3c/webtransport/issues/326.

cc @rakina

***
<a href="https://pr-preview.s3.amazonaws.com/nidhijaju/webtransport/pull/500.html" title="Last updated on Apr 25, 2023, 12:40 PM UTC (c52beb2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/500/b2b4b07...nidhijaju:c52beb2.html" title="Last updated on Apr 25, 2023, 12:40 PM UTC (c52beb2)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1097,8 +1097,13 @@ This specification defines [=unloading document cleanup steps=] as the following
 
 1. Let |window| be |document|'s [=relevant global object=].
 1. For each {{WebTransport}} |transport| whose [=relevant global object=] is |window|:
-  1. If |transport|.{{[[State]]}} is `"connected"`, set |transport|.{{[[State]]}} to `"failed"`
-     and [=session/terminate=] |transport|.{{[[Session]]}} [=in parallel=].
+  1. If |transport|.{{[[State]]}} is `"connected"`, then:
+    1. Set |transport|.{{[[State]]}} to `"failed"`.
+    1. [=In parallel=], [=session/terminate=] |transport|.{{[[Session]]}}.
+    1. [=Queue a network task=] with |transport| to run the following steps:
+      1. Let |error| be a newly [=DOMException/created=] {{WebTransportError}} whose
+         {{WebTransportErrorOptions/source}} is `"session"`.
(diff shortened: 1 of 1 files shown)


aboba (approved):

rakina:
Thanks for working on this! This is great, but recently the guidance on adding [unloading document cleanup steps](https://html.spec.whatwg.org/multipage/document-lifecycle.html#unloading-document-cleanup-steps) had changed, so instead of using the hook like this PR does, to define a stand-alone set of steps, that the HTML spec explicitly calls out to. This makes the ordering of the steps clear for all things that run at unloading-document-cleanup time. 
(The [BFCache guide](https://html.spec.whatwg.org/multipage/document-lifecycle.html#unloading-document-cleanup-steps) was recently updated to reflect this as well)

So I think there are some more changes needed:
1. Instead of defining "unloading document cleanup steps" for HTML make it an independent thing (you can keep the name, or use "context cleanup steps" like the section heading says.
2. Update the HTML spec to call out to this step in the Web Transport spec from within the HTML spec's "unloading document cleanup steps"

Also, are there any implementation bugs filed for Chrome (and maybe Firefox) yet?

(thanks again!)

cc @domenic @fergald

fergald:
Ping any progress on this? I would really like WT to have it's BFCache story straight from the start.

PR #502, opened 2023-04-25
Add getAtomicWriter() API.

### LINKED ISSUE #426: When flow control is blocked
Our QUIC implementation takes a very particular approach to how it manages flow control.  In order to avoid flow control deadlocks, which are a serious problem, it does not accept data on streams that have insufficient flow control credit.  This helps us avoid deadlocks in our application protocol code because any code that creates dependent writes can use a rejected write as a signal.  For instance, in QPACK, we attempt to write out QPACK table inserts before we reference them, but if the insert fails, we change to using a literal value.

Our interface also has two write interfaces: one that takes a stream of bytes and reports how many bytes were accepted; and, one that takes a block of bytes and only does anything if the entire block can be written.  We found this to be essential.

What is the attitude of the WebTransport API when it comes to this pattern?  It really isn't clear from the [write](https://w3c.github.io/webtransport/#webtransportsendstream-write) definition.

jan-ivar:
Fixes #426

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webtransport/pull/502.html" title="Last updated on Jun 21, 2023, 9:04 PM UTC (002fcad)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/502/37dee3b...jan-ivar:002fcad.html" title="Last updated on Jun 21, 2023, 9:04 PM UTC (002fcad)">Diff</a>


index.bs
```diff
@@ -1395,6 +1396,20 @@ The {{WebTransportSendStream}}'s [=transfer steps=] and
            1. [=Resolve=] |p| with |stats|.
      1. Return |p|.
 
+: <dfn for="WebTransportSendStream" method>getAtomicWriter()</dfn>
+:: For transactional applications where blocking on flow control can be an
+   issue, the following alternative writer API gives the application
```
martinthomson, line 1401: A citation for https://datatracker.ietf.org/doc/html/rfc9204#section-2.1.3 might help here.  It's not a great citation, but it highlights the sort of issue this might be avoiding.

index.bs
```diff
@@ -1395,6 +1396,20 @@ The {{WebTransportSendStream}}'s [=transfer steps=] and
...
+   additional control.
+
+   Implemented on top of the default writer, except any data the sink cannot
+   send atomically is dropped without erroring the stream, causing the
+    promise to settle with false.
```
martinthomson, line 1406:
```suggestion
   promise to settle with false.
```

jan-ivar:
Meeting:
- what does app do when it gets false? Try again later?
- Does API need to expose more info, e.g. fewer bytes might work?

index.bs
```diff
@@ -1897,6 +1924,17 @@ object |transport|, and a |sendOrder|, run these steps.
...
 </div>
 
+# `WebTransportAtomicWriter` Interface #  {#web-transport-atomic-writer-interface}
+
+{{WebTransportAtomicWriter}} is a specialized version of {{WritableStreamDefaultWriter}}.
+
+<pre class="idl">
+[Exposed=*]
+interface WebTransportAtomicWriter : WritableStreamDefaultWriter {
+  Promise&lt;bool&gt; write(optional any chunk);
```
martinthomson, line 1934: Adding a `Promise<unsigned long long> getAvailable()` function might help in the recovery scenario.  That function would come with a pretty big caveat: the value you get isn't guaranteed to be the value that is applied when you subsequently write.  But it might help guide decision-making during recovery in the general case, provided that the browser and app are fast in processing and acting upon the result.

index.bs
martinthomson: Lucas reminded me of [RFC 9308](https://www.rfc-editor.org/rfc/rfc9308.html#section-4.4) which has a lot more and more useful text.

index.bs
jan-ivar: Would that number match the number of bytes that got sent? If so, we could instead resolve `write()` with the number of bytes that got sent... maybe?

LPardue: That's sort of how the API works in Cloudflare quiche, you can ask to send N bytes and we'll return X (between 1 and N) - https://docs.quic.tech/quiche/struct.Connection.html#method.stream_send

martinthomson: Returning how many bytes were written makes sense for a non-atomic API - we do the same in our stack - but it doesn't make sense here.  This is expressly an all-or-nothing option.  The idea was to provide a query that reports many bytes would/could have been written if a write were attempted at the time of the query.

jan-ivar: @martinthomson OK how about we redefine the number returned from `write` to be the number of bytes (or less) you should try instead? Simply to avoid a separate `getAvailable()` method.

martinthomson: Won't that just be confusing?  I know that it streamlines code, but it is still a bit weird.  You write N bytes, and get back M == N to indicate that it worked and M < N to indicate that it didn't?  I don't see why we would want to save the extra call when that is far easier to comprehend.

jan-ivar:
Meeting:
- No objections bu @vasilvv to review.

martinthomson (approved):
As always, the complexity here is a marvel.  Hopefully the implementation can be simpler.

index.bs
```diff
@@ -1939,6 +1984,71 @@ object |transport|, and a |sendOrder|, run these steps.
...
+cannot be sent in its entirety without blocking on flow control, then the data is
+dropped without erroring the stream, and the promise is resolved with false.
+
+A {{WebTransportAtomicWriter}} is always created by the
+[=WebTransportAtomicWriter/create=] procedure.
+
+<pre class="idl">
+[Exposed=*]
+interface WebTransportAtomicWriter : WritableStreamDefaultWriter {
+  Promise&lt;bool&gt; write(optional any chunk);
```
jan-ivar, line 2000:
```suggestion
  Promise&lt;boolean&gt; write(optional any chunk);
```

jan-ivar:
I've taken a stab at reducing the number of steps a bit.

index.bs
jan-ivar: I've reverted back to a boolean to resolve this. Let's discuss `Promise<unsigned long long> getAvailable()` in a new issue if needed.

jan-ivar:
This still isn't quite right. Needs to prevent sending of queued writes, not just resolve their promises.

martinthomson:
Good point.  A failed atomic write should probably cause all subsequent writes to fail, even if they are not atomic.

jan-ivar:
> A failed atomic write should probably cause all subsequent writes to fail, even if they are not atomic.

For atomic writes, where "fail" just means succeeding with `false`, that makes sense: the all-or-nothing guarantee couldn't be kept so this is desired behavior.

But regular writes have no mechanism to "fail" a single write that doesn't terminate the stream, which doesn't seem POLA. So I don't think we want that. I think normal writes have to still block, whether atomic writes are sprinkled in or not. Otherwise a single atomic write makes them all atomic effectively, which seems wrong.

martinthomson:
The problem with that is that a stream is an ordered construct and the thing that writes to the stream might have expectations about the order in which things appear in the stream.  I think that's a risk that people assume when writing, but the effect could be surprising.  Our HTTP implementation relies on being able to mix writes of different types (atomic/non-atomic).  However, it does tend to wait for one write to resolve before attempting the next.

At a minimum, we need to be very clear about what we expect to have happen if a write fails at any point.  If an application performs multiple writes with an atomic write mixed in, I would have expected all subsequent writes to abort also.

If, however, the stream semantics demands that writes only fail when the underlying stream breaks, then there are alternatives:

1. Block the stream until flow control becomes available for the atomic write.  This is a great way to deadlock your application, so it probably isn't a great way to go if the browser does it.  We should, however, encourage applications to wait for an atomic write to resolve before writing anything else.

2. Drop all subsequent writes, as I suggested.  This says that if you attempt and fail to write atomically, then you need to start over from the point at which the failure occurred.  I don't see why this wouldn't be a problem, unless this violates expectations for normal writes in some fundamental way.

3. Have bytes disappear from the sequence.  I think that this is what you suggested.  This likely breaks most protocols.  Imagine if I sent an HTTP header (as atomic) and then the body, but if the header disappeared and only the body got through.  The other end might be surprised.

I can't think of any other alternatives offhand.

jan-ivar:
Thanks for explaining. I agree 2 makes the most sense given the requirements in the OP.

> If, however, the stream semantics demands that writes only fail when the underlying stream breaks,

They do, as the streams spec even encourages apps to [ignore the write promise](https://streams.spec.whatwg.org/#example-manual-write-with-backpressure) for better throughput:
```js
  const writer = writableStream.getWriter();

  while (true) {
    await writer.ready;

    const bytes = new Uint8Array(writer.desiredSize);
    crypto.getRandomValues(bytes);

    // Purposefully don’t await; awaiting writer.ready is enough.
    writer.write(bytes).catch(() => {});
  }
```


> 2. Drop all subsequent writes, as I suggested. This says that if you attempt and fail to write atomically, then you need to start over from the point at which the failure occurred.

So what does "drop" mean here? Resolve the write promises without having written anything?

That could work, but doesn't seem low on the surprise scale either.

Let me try to rephrase 2 (without the word "fail") to see if I understand:

 2. Resolve all subsequent write promises without writing anything. This says that if the atomic write couldn't be written due to a flow control block, then you need to start over from the point you made the atomic write.

jan-ivar:
>  Our HTTP implementation relies on being able to mix writes of different types (atomic/non-atomic).

What's the utility of queuing a non-atomic write after an atomic one? The current API seems clunky for that:
```js
// one atomic write...
let writer = sendStream.getAtomicWriter();
await writer.ready;
const haveSent = writer.write(data1); // haveSent is a promise
writer.releaseLock();

// ...followed by a non-atomic write
writer = sendStream.getWriter();
writer.write(data2).catch(() => {});
writer.releaseLock();

console.log(`Atomic got sent = ${await haveSent}.`;
```

If mixing will be common, maybe we should simplify the API?
```js
await writer.ready;
const haveSent = writer.writeAtomic(data1); // one atomic write
writer.write(data2).catch(() => {});        // followed by a non-atomic write

console.log(`Atomic got sent = ${await haveSent}.`;
```

jan-ivar:
Meeting:
- @vasilvv  since we're subclassing the writer, we could make the promise from write reject instead of succeeding with false, if we want
- We could also invent a new AtomicBufferSource to reuse the .write() method, if we wanted. But seems strange for app to have to transform BufferSource objects just to send them atomically.
-

bitcoinfinancier (approved):

wilaw:
This PR referenced at TPAC 2023 meeting with this slide: https://docs.google.com/presentation/d/1PzshEzs8GPeoYvVboi-9D0n8ZtXBZHuM7uZe1blahSk/edit#slide=id.g27bc4bd45e1_0_125

jan-ivar:
Superseded by https://github.com/w3c/webtransport/pull/551.

PR #504, opened 2023-04-26
Add .gitignore to remove index.html

nidhijaju:
We don't want to update the index.html file every time we make a PR to the repo, so this adds a .gitignore to prevent that from accidentally happening, and also removes the existing index.html file.


martinthomson (approved):
I generally add a few extra things when adding a .gitignore:

```
*.swp
*~
```

But you don't have to.

jan-ivar (approved):

Issue #492, opened 2023-03-29
Allow updating send order for streams


vasilvv:
We currently only allow sendOrder to be set whenever a stream is created.  This is limiting, since we don't get to change the priority of incoming streams.  I think we should add an `updateSendOrder()` method to deal with that.

jan-ivar:
Meeting:
- We need this
- Bikeshedding on API
  - Do we want to keep things simpler for implementations?
  - stream.priority = n — keeps the API regular
  - what if you're running code divorced from where it was created
  - No other good alternatives come to mind

PR #510, opened 2023-05-09
Add sendStream.sendOrder = n attribute.

jan-ivar:
Fixes https://github.com/w3c/webtransport/issues/492.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webtransport/pull/510.html" title="Last updated on May 9, 2023, 12:41 PM UTC (339cc0a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/510/7bb74aa...jan-ivar:339cc0a.html" title="Last updated on May 9, 2023, 12:41 PM UTC (339cc0a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1372,6 +1372,7 @@ data to the server.
 <pre class="idl">
 [Exposed=(Window,Worker), SecureContext, Transferable]
 interface WebTransportSendStream : WritableStream {
+  attribute long long? sendOrder;
   Promise&lt;WebTransportSendStreamStats&gt; getStats();
 };
 </pre>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1491,6 +1500,10 @@ To <dfn for="WebTransportSendStream">write</dfn> |chunk| to a {{WebTransportSend
      [=WritableStream/Error | errored=] nor blocked by [=flow control=], have been
      sent.
 
+    Note: We access |stream|.{{[[SendOrder]]}} [=in parallel=] here. User agents SHOULD
+    respond to live updates of these values during sending, though the details are
+    [=implementation-defined=].
```
martinthomson, line 1505: So the implementation of this will be asynchronous in practice, but the interface is simple.  That means that browsers will need to retain a value for the stream objects it maintains.  It then needs to propagate changes asynchronously to the stack, as this mentions.

The alternative is to describe two values, with the one that is used here being updated asynchronously (and potentially fallibly, with no error reporting).

Issue #446, opened 2022-12-13
MAX_STREAM limit (ingest) (still need an event when stream ids become available again)


aboba:
Recently on the MoQ WG mailing list, there was a thread relating to interaction of frame/stream transport with the MAX_STREAMS limit.  Imagine a conferencing application implementing a 7x7 grid of videos from each participant.  If 10 QUIC streams needed to be simultaneously open per video stream, then this could yield a requirement for MAX_STREAMS >= 500 just for video. 

How would an application ensure the MAX_STREAMS limit is set to a suitable value for a WebTransport session? 

Pointer to the thread: https://mailarchive.ietf.org/arch/msg/moq/hecXJfwaysqbyp85ZUwAT9I5CM0/

martinthomson:
This would be something that can be adjusted at runtime, but you are right to note that this is probably higher than the default mode that an HTTP client would choose.

Obviously, for upload, the server can just expand as needed, but the client probably needs an API.

jan-ivar:
> Obviously, for upload, the server can just expand as needed, but the client probably needs an API.

Why can't the browser expand as needed?

martinthomson:
This is a limit that the server exercises.  Let's say that the client starts at 100, which is a reasonable starting point for a great many use cases.  But if the server needs 500, the client won't know that 500 unless it is somehow told.  This is true for any starting point, as it seems like any default will - in some cases - be exceeded.

jan-ivar:
Meeting:
- should we report these (server) limits to JS?
- default in Chrome may be 100 for sending?
- partially a resource constraint, partially an application concurrency control
- it is cumulative
- your quic stack would update the limits.
- for server to client direction, the application might want to raise its limit, as client is willing to accept those streams
- for client to server direction, you don't know what the limit is, especially if it's pooled
- example: conferencing app, might have 5 concurrent frames sent at the same time, per media stream. x 10 users, would hit server to client limit.
- concerned about resource attacks on the server, and "give me more" JS apis
- might want a specific error code to say you've exceeded the quota of number of streams.

wilaw:
Per the QUIC spec https://datatracker.ietf.org/doc/html/draft-ietf-quic-transport-18#section-19.11, the Maximum Streams are defined as "A count of the cumulative number of streams of the corresponding type that can be opened over the lifetime of the connection". 

In this thread @martinthomson suggests that 100 might be a reasonable starting point. 

[...]

Issue #469, opened 2023-02-14
API behavior when opening a new stream is blocked


vasilvv:
Our current behavior that is specified in the draft is that whenever the stream limit is exceeded, the API returns a promise that is blocked until we receive a new MAX_STREAMS, upon which the stream is opened and the promise is resolved.

Chrome's current implementation apparently does not do that (it immediately rejects a promise).  I was working on fixing that, but then I realized that we probably should not allow the JS code to request infinitely many streams (in our implementation, this is problematic since the requests get buffered on the more privileged side of the stack).

I think the correct answer is to allow *some* amount of pending promises (1x currently open streams? 2x?), after which we should start rejecting new streams immediately.

jesup:
So if I have 1 stream open, and need to open 5 more because someone just hit 'play', I can't, and have to pace myself to open them 1 (or a couple) at  time? I think that's going to cause all sorts of heartache.   And worse, a setup that work may suddenly fail to open streams because we closed a stream earlier.

jesup:
If this only kicks in at (say) 100 open streams, it's far less of an issue, of course (it wasn't clear from the above; it seems to state this limit based on open streams applies independent of the server stream limit).

Video with stream-per-frame could create quite a few streams in a short period (though probably not more than hundreds, even for high-framerate video).   I'd also note that the cases where we'd hit this are pretty much edge-cases, and if the resources required to allow it aren't all that large, I'd err on the high side.

ricea:
I don't think we should change the standard to work around the limitations in Chrome's implementation. Instead, Chrome should queue stream requests on the unprivileged side.

ekinnear:
On the QUIC side, we've also been pushing to ensure that streams are generally "cheap", and that does seem to have been especially valuable for a number of media use cases. So even if you attempt to open 100 streams all at the same time, is it actually a problem if those promises don't resolve for a while?

wilaw:
> So even if you attempt to open 100 streams all at the same time, is it actually a problem if those promises don't resolve for a while?

For a client sending frame-per-stream media at 30fps, if stream creation takes > 33ms, or even some meaningful fraction of that, it is going to fall behind. Especially true if user-agent has say a 100ms RTT to server and has to negotiate an increase in max streams before it can authorize new stream creation. We can anticipate media at 120fps which further exacerbates the consequences of any delay in stream creation.

martinthomson:
[...]

PR #528, opened 2023-07-19
Reject create(Uni|Bi)directionalStream() on stream ID exhaustion.

jan-ivar:
Fixes https://github.com/w3c/webtransport/issues/446.

***

### :boom: Error: 500 Internal Server Error :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Jul 19, 2023, 7:52 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [CSS Spec Preprocessor](https://api.csswg.org/bikeshed/) - CSS Spec Preprocessor is the web service used to build Bikeshed specs.

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2Fw3c%2Fwebtransport%2Fb6eef8f292a437fcd1a878389e1fb1dd3c0de92f%2Findex.bs&md-warning=not%20ready)

```
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>500 Internal Server Error</title>
</head><body>
<h1>Internal Server Error</h1>
<p>The server encountered an internal error or
misconfiguration and was unable to complete
your request.</p>
<p>Please contact the server administrator at 
 [no address given] to inform them of the time this error occurred,
 and the actions you performed just before this error.</p>
<p>More information about this error may be available
in the server error log.</p>
<hr>
<address>Apache/2.4.10 (Debian) Server at api.csswg.org Port 443</address>
</body></html>

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/webtransport%23528.)._
</details>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -996,8 +996,10 @@ these steps.
       {{[[State]]}} becomes `"closed"` or `"failed"`, and instead
       [=queue a network task=] with |transport| to [=reject=] |p| with an {{InvalidStateError}}:
       1. Let |streamId| be a new stream ID that is valid and unique for
-         |transport|.{{[[Session]]}}. If one is not immediately available due to exhaustion,
-         [=reject=] |p| with a {{QuotaExceededError}} and abort these steps. [[!QUIC]]
+         |transport|.{{[[Session]]}}, as defined in [[!QUIC]]
+         [Section 19.11](https://www.rfc-editor.org/rfc/rfc9000#section-19.11). If one is not
+         immediately available due to exhaustion, [=reject=] |p| with a {{QuotaExceededError}}
+         and abort these steps.
       1. Let |internalStream| be the result of [=creating a bidirectional stream=] with
            |transport|.{{[[Session]]}} and |streamId|.
       1. [=Queue a network task=] with |transport| to run the following steps:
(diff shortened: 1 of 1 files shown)


vasilvv:
I'm a little bit confused.  I remember last time we discussed this in #469, the conclusion was that we're supposed to block the promise until the next stream is available.  Have I missed something?

jan-ivar:
My understanding from the [May 23 meeting notes](https://www.w3.org/wiki/WebTransport/Meetings2023#WebTransport_Bi-weekly_Virtual_Meeting_.2360_late_-_May_23rd.2C_2023) and https://github.com/w3c/webtransport/issues/446#issuecomment-1560274258 was that we wanted to rethink this:
> TODO:
> - Need a PR to reject createBi/UniDirectionStream() in a reliable way, when MAX stream limit is hit. Let's app drop its request, OR
> - Fire some limitreached event

I believe the thinking was that blocking the promise made it difficult for the application to detect that the MAX_STREAM_LIMIT has been hit, and that waiting wasn't always the right cause of action. Rejecting instead would let the application know what happened so it can decide what to do next.

ricea:
But how can the app know when the limit has been increased and it is safe to call Create*Stream() again?

PR #551, opened 2023-10-10
Add writer.atomicWrite() method.

jan-ivar:
Fixes https://github.com/w3c/webtransport/issues/426, superseding #502.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webtransport/pull/551.html" title="Last updated on Jan 30, 2024, 11:57 PM UTC (fbc834b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/551/af7ea01...jan-ivar:fbc834b.html" title="Last updated on Jan 30, 2024, 11:57 PM UTC (fbc834b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1594,10 +1594,6 @@ To <dfn for="WebTransportSendStream">write</dfn> |chunk| to a {{WebTransportSend
 1. Set |stream|.{{[[PendingOperation]]}} to |promise|.
 1. Let |inFlightWriteRequest| be
    |stream|.<a href="https://streams.spec.whatwg.org/#writablestream-inflightwriterequest">inFlightWriteRequest</a>.
-
-   Note: Find a [better way](https://streams.spec.whatwg.org/#other-specs) to integrate
-   with the streams spec to identify the write request in flight.
-
 1. Let |atomic| be true if |inFlightWriteRequest| [=list/exists=]
    in [=stream=].{{WebTransportSendStream/[[AtomicWriteRequests]]}}, otherwise false.
 1. Run the following steps [=in parallel=]:
@@ -1616,7 +1612,7 @@ To <dfn for="WebTransportSendStream">write</dfn> |chunk| to a {{WebTransportSend
      except for [=flow control=] reasons or [=WritableStream/Error | error=].
 
      If the sending of |bytes| cannot be completed in its entirety within the current
-     [=flow control=] window, and |atomic| is true, then give up on sending |bytes| and
+     [=flow control=] window, and |atomic| is true, then stop sending |bytes| and
      [=queue a network task=] with |transport| to [=abort all atomic write requests=]
      on |stream|, before proceeding immediately to the next step without failing.
 
--- a/index.bs
+++ b/index.bs
@@ -2770,7 +2770,7 @@ async function receiveText(url, createWritableStreamForTextData) {
 
 *This section is non-normative.*
 
-Sending a transactional piece of data on a one-way stream only if it can be done
+Sending a transactional piece of data on a unidirectional stream, only if it can be done
 entirely without blocking on [=flow control=], can be achieved by using the
 {{WebTransportSendStream/getWriter}} function and the resulting writer.
 
--- a/index.bs
+++ b/index.bs
@@ -1598,7 +1598,7 @@ To <dfn for="WebTransportSendStream">write</dfn> |chunk| to a {{WebTransportSend
   1. [=Queue a network task=] with |transport| to run these steps:
     1. Set |stream|.{{[[PendingOperation]]}} to null.
     1. If |atomic| is true, [=map/remove=] |inFlightWriteRequest| from
-       |stream|.{{WebTransportSendStream/[[AtomicWriteRequests]]}}.
+       |stream|.{{WebTransportSendStream/[[AtomicWriteRequests]]}} if it exists.
     1. [=Resolve=] |promise| with undefined.
 1. Return |promise|.
 
@@ -2073,7 +2073,10 @@ on a {{WebTransportSendStream}} |stream|, given |chunk|, run these steps:
 1. Let |p| be the result of {{WritableStreamDefaultWriter/write(chunk)}}
     on {{WritableStreamDefaultWriter}} with |chunk|.
 1. Set |stream|.{{WebTransportSendStream/[[AtomicWriteRequests]]}} to |p|.
-1. Return |p|.
+1. Return the result of [=reacting=] to |p| with the following steps:
+   1. [=map/Remove=] |p| from |stream|.{{WebTransportSendStream/[[AtomicWriteRequests]]}} if it exists.
+   1. If |p| was rejected with reason |r|, then return [=a promise rejected with=] |r|.
+   1. Return undefined.
 
 </div>
 
--- a/index.bs
+++ b/index.bs
@@ -1501,7 +1501,7 @@ A {{WebTransportSendStream}} has the following internal slots.
   </tr>
   <tr>
    <td><dfn>`[[AtomicWriteRequests]]`</dfn>
-   <td class="non-normative">A [=list=] of promises, each representing an in-flight atomic
+   <td class="non-normative">An [=ordered set=] of promises, each representing an in-flight atomic
     write request to be processed by the underlying sink.
   </tr>
  <tbody>
@@ -2072,7 +2072,7 @@ To <dfn export for="WebTransportWriter" lt="write atomically">write atomically</
 on a {{WebTransportSendStream}} |stream|, given |chunk|, run these steps:
 1. Let |p| be the result of {{WritableStreamDefaultWriter/write(chunk)}}
     on {{WritableStreamDefaultWriter}} with |chunk|.
-1. Set |stream|.{{WebTransportSendStream/[[AtomicWriteRequests]]}} to |p|.
+1. [=set/Append=] |p| to |stream|.{{WebTransportSendStream/[[AtomicWriteRequests]]}}.
 1. Return the result of [=reacting=] to |p| with the following steps:
    1. [=map/Remove=] |p| from |stream|.{{WebTransportSendStream/[[AtomicWriteRequests]]}} if it exists.
    1. If |p| was rejected with reason |r|, then return [=a promise rejected with=] |r|.
--- a/index.bs
+++ b/index.bs
@@ -70,7 +70,6 @@ spec:url; type:dfn; text:scheme
 spec:url; type:dfn; text:fragment
 spec:infra; type:dfn; for:/; text:ASCII case-insensitive
 spec:infra; type:dfn; text:list
-spec:streams; type:dict-member; text:write
 </pre>
 <pre class="anchors">
 url: https://html.spec.whatwg.org/multipage/origin.html#concept-origin; type: dfn; text: origin; for:/
--- a/index.bs
+++ b/index.bs
@@ -1594,8 +1594,8 @@ To <dfn for="WebTransportSendStream">write</dfn> |chunk| to a {{WebTransportSend
 1. Set |stream|.{{[[PendingOperation]]}} to |promise|.
 1. Let |inFlightWriteRequest| be
    |stream|.<a href="https://streams.spec.whatwg.org/#writablestream-inflightwriterequest">inFlightWriteRequest</a>.
-1. Let |atomic| be true if |inFlightWriteRequest| [=list/exists=]
-   in [=stream=].{{WebTransportSendStream/[[AtomicWriteRequests]]}}, otherwise false.
+1. Let |atomic| be true if [=stream=].{{WebTransportSendStream/[[AtomicWriteRequests]]}}
+   [=list/contains=] |inFlightWriteRequest|, otherwise false.
 1. Run the following steps [=in parallel=]:
   1. [=stream/Send=] |bytes| on |stream|.{{WebTransportSendStream/[[InternalStream]]}} and wait for the
      operation to complete.
@@ -1634,8 +1634,7 @@ To <dfn for="WebTransportSendStream">write</dfn> |chunk| to a {{WebTransportSend
   1. Otherwise, [=queue a network task=] with |transport|
      to run these steps:
     1. Set |stream|.{{[[PendingOperation]]}} to null.
-    1. [=map/Remove=] |inFlightWriteRequest| from
-       |stream|.{{WebTransportSendStream/[[AtomicWriteRequests]]}} if it exists.
+    1. If |stream|.{{WebTransportSendStream/[[AtomicWriteRequests]]}} [=list/contains=] |inFlightWriteRequest|, [=list/remove=] |inFlightWriteRequest|.
     1. [=Resolve=] |promise| with undefined.
 1. Return |promise|.
 
@@ -1691,7 +1690,7 @@ To <dfn for="WebTransportSendStream">abort</dfn> a {{WebTransportSendStream}} |s
 <div algorithm>
 To <dfn for="WebTransportSendStream">abort all atomic write requests</dfn> on a {{WebTransportSendStream}} |stream|, run these steps:
   1. Let |requestsToAbort| be [=stream=].{{WebTransportSendStream/[[AtomicWriteRequests]]}}.
-  1. [=map/Clear=] [=stream=].{{WebTransportSendStream/[[AtomicWriteRequests]]}}.
+  1. [=list/Empty=] [=stream=].{{WebTransportSendStream/[[AtomicWriteRequests]]}}.
   1. [=For each=] |promise| in |requestsToAbort|, [=reject=] |promise| with {{AbortError}}.
   1. [=In parallel=], [=for each=] |promise| in |requestsToAbort|, abort the
      [=stream/send|sending=] of bytes associated with |promise|.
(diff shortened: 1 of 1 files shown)


jan-ivar:
Meeting:
- Use AbortError

jan-ivar:
Meeting:
- Remove "Whenever the sending of bytes becomes blocked ..." because semantically we want to allow queuing atomic writes behind non-atomic ones. I.e. atomic = all or nothing (but all or nothing may be queued)

index.bs
```diff
@@ -1591,6 +1615,11 @@ To <dfn for="WebTransportSendStream">write</dfn> |chunk| to a {{WebTransportSend
      This sending MUST NOT starve otherwise,
      except for [=flow control=] reasons or [=WritableStream/Error | error=].
 
+     If the sending of |bytes| cannot be completed in its entirety within the current
+     [=flow control=] window, and |atomic| is true, then give up on sending |bytes| and
```
martinthomson, line 1619:
```suggestion
     If |atomic| is true and |bytes| cannot all be sent within
     the available [=flow control=] window, then give up on sending |bytes| and
```

Does "give up" mean that the step fails?  That needs to more clear here, I think.

index.bs
```diff
@@ -2089,6 +2130,81 @@ object |transport|, and a |sendOrder|, run these steps.
...
+  Promise&lt;undefined&gt; write(optional any chunk);
+  Promise&lt;undefined&gt; atomicWrite(optional any chunk);
+};
+</pre>
+
+## Methods ##  {#web-transport-writer-procedures-methods}
+
+: <dfn for="WebTransportWriter" method>write(chunk)</dfn>
+:: The {{write}} method will not reject on [=flow control=] blockage
+   (unless queued behind one or more outstanding calls to {{atomicWrite}}).
```
martinthomson, line 2153:
```suggestion
   unless queued behind one or more outstanding calls to {{atomicWrite}}.
```

index.bs
```diff
@@ -2089,6 +2130,81 @@ object |transport|, and a |sendOrder|, run these steps.
...
+:: The {{atomicWrite}} method will reject if the |chunk| given to it
+   cannot be sent in its entirety within the [=flow control=] window that
+   is current at the time of sending. In other words, the sending of |chunk|
+   is all or nothing. This behavior is designed to satisfy niche transactional
+   applications sensitive to [=flow control=] deadlocks ([[RFC9308]]
+   [Section 4.4](https://datatracker.ietf.org/doc/html/rfc9308#section-4.4)).
+
+   Note: Atomic writes can still block if queued behind non-atomic writes. If
+   the write is rejected, everything queued behind it will be rejected as well,
+   including non-atomic writes. Applications are therefore encouraged to always
```
martinthomson, line 2173:
```suggestion
   including non-atomic writes, which are enqueued as atomic writes when an
   atomic write is outstanding. Applications are therefore encouraged to always
```

index.bs
```diff
@@ -2654,6 +2770,31 @@ async function receiveText(url, createWritableStreamForTextData) {
...
+
+<pre class="example" highlight="js">
+async function sendTransactionalData(wt, bytes) {
+  const writable = await wt.createUnidirectionalStream();
+  const writer = writable.getWriter();
+  await writer.ready;
+  try {
+    await writer.atomicWrite(bytes);
+  } catch (e) {
+    if (e.name != "TransactionInactiveError") throw e;
```
martinthomson, line 2789:
```suggestion
    if (e.name != "AbortError") throw e;
```

martinthomson (approved):

index.bs
jan-ivar: How about s/give up on/stop/? The sentence already ends with: _"before proceeding immediately to the next step without failing"_, so it is not failing. This is important to properly clear [[PendingOperation]], which the existing algorithm doesn't do on failure (didn't need to since algo-failure here remains terminal.

index.bs
```diff
@@ -1616,7 +1612,7 @@ To <dfn for="WebTransportSendStream">write</dfn> |chunk| to a {{WebTransportSend
      except for [=flow control=] reasons or [=WritableStream/Error | error=].
 
      If the sending of |bytes| cannot be completed in its entirety within the current
-     [=flow control=] window, and |atomic| is true, then give up on sending |bytes| and
+     [=flow control=] window, and |atomic| is true, then stop sending |bytes| and
```
martinthomson, line 1615: Maybe the right way to deal with this problem is to have three distinct outcomes from this step, so that you can execute the correct code on subsequent steps.

jan-ivar:
@vasilvv does this look good?

vasilvv:
I believe this does address my concerns regarding the ordering, though I am going to defer to @nidhijaju as to whether this makes sense from the streams API perspective.

jan-ivar:
@nidhijaju or @ricea could you take a look at this from a WHATWG streams perspective?

jan-ivar:
@nidhijaju would you mind reviewing this? We would like to include it in wide review, and this is holding that up.

nidhijaju:
This looks okay to me, but @ricea had some concerns about this maybe not being interoperable with other streams APIs (I'll let him comment with more details), so just waiting on that. Sorry for the delay.

ricea:
A less well-known way to get a writer is
```js
const writer = new WritableStreamDefaultWriter(stream);
```
In this case, obviously the writer wouldn't have an `atomicWrite()` method, but it's important that everything else still works properly, and there's no misbehaviour.

Currently, if you write something like
```js
const writer1 = stream.getWriter();
writer1.atomicWrite(block1);
writer1.releaseLock();
const writer2 = new WritableStreamDefaultWriter(stream);
writer2.write(block2);
```
`block2` will not get the atomic treatment. From my understanding, this won't cause anything to explode, and since no-one is likely to do that, it's not a big problem. A developer could also intentionally call the base class method by doing something like
```js
WritableStreamDefaultWriter.prototype.write.call(writer, chunk);
```
which is possibly more likely when some framework that wraps stream functionality is in use.

Another issue with overriding the `write()` method is that if we add an optional second argument in the Streams Standard in future, it will not be easily usable with WebTransport.

Piping after releasing the lock also won't set the _atomic_ flag. I don't know if this is a problem or not.
```js
const writer = stream.getWriter();
writer.atomicWrite(block);
writer.releaseLock();
await otherStream.pipeTo(stream);
```

If we add a bytes-optimised writer type to WritableStream in future it won't work with WebTransport and it may be surprising to developers that passing the optional `type` argument to SendStream's `getWriter()` doesn't do anything.

From a design perspective, `atomicWrite()` doesn't fit into the streaming model and won't interoperate nicely with other streams, but this seems to arise directly from the design requirements and is probably not fixable while accommodating this use case.

index.bs
```diff
@@ -1620,18 +1640,27 @@ To <dfn for="WebTransportSendStream">write</dfn> |chunk| to a {{WebTransportSend
 
      Note: The definition of fairness here is [=implementation-defined=].
 
-  1. If the previous step failed, abort the remaining steps.
+     Stop sending |bytes| if |atomic| is true and the current [=flow control=] window
```
ricea, line 1643: The fact that the instruction to stop comes after the instruction to send seems to imply that the chunk may be partially sent before stopping. Based on the name `atomic` I guess that is not the intention.

martinthomson:
> From a design perspective, atomicWrite() doesn't fit into the streaming model and won't interoperate nicely with other streams, but this seems to arise directly from the design requirements and is probably not fixable while accommodating this use case.

Yes, I don't see a good way out of that, though maybe this goes back to the streams API.  I don't know why stream locking and unlocking is atomic.

My experience with asynchronous systems elsewhere suggests that acquiring a mutex is not atomic.  It might be *presented* through a synchronous operation, which means that the call blocks the thread.  Analogous systems often provide asynchronous equivalents that do not block.  Here, it seems like we could hold the lock until the atomic write is resolved, but there is no way to delay the acquisition of a new lock.

For writes, like this, the idea that the write promise might take time to resolve is fine, but that this does not delay the release of the lock is a bit strange.

index.bs
jan-ivar: As written, nothing protects the server from receiving a partial message. How should it work differently, say with a 100 MB chunk?

jan-ivar:
AFAICT [IsWritableStreamLocked(stream)](https://streams.spec.whatwg.org/#is-writable-stream-locked) isn't a mutex, just a TypeError-throwing `stream.[[writer]]` state to help delineate the two writing APIs (piping vs writer.write) for web developers.

> For writes, like this, the idea that the write promise might take time to resolve is fine, but that this does not delay the release of the lock is a bit strange.

Interestingly, `pipeTo` [acquires](https://streams.spec.whatwg.org/#ref-for-acquire-writable-stream-default-writer) `[[writer]]` synchronously, and holds it until [it resolves its promise](https://streams.spec.whatwg.org/#rs-pipeTo-finalize).

Might we infer from that that apps calling `writer.releaseLock()` ahead of all its `writer.write` promises having settled are in the wrong?

jan-ivar:
> Piping after releasing the lock also won't set the atomic flag. I don't know if this is a problem or not.
> ```js
> const writer = stream.getWriter();
> writer.atomicWrite(block);
> writer.releaseLock();
> await otherStream.pipeTo(stream);
> ```

Flow control blocking `block` would [fail](https://streams.spec.whatwg.org/#ref-for-writable-stream-finish-in-flight-write-with-error) the `pipeTo` if it did, escalating to terminally erroring both `otherStream` and `stream`. The chances of that being desirable seem low.

> ... [new WritableStreamDefaultWriter(stream).write(block2);] will not get the atomic treatment.

> Another issue with overriding the write() method is that if we add an optional second argument...

> If we add a bytes-optimised writer type to WritableStream in future it won't work with WebTransport...

These three problems seem to stem from modifying `write`, a late addition to this PR. The inconsistency with pipeTo and the other way of obtaining the writer is perhaps also not terrific. Do we want to reexamine the benefit of that addition?

That addition was: any `write` following an unsettled `atomicWrite` is an atomic write. A convenience at best (workaround: use `atomicWrite` instead), and a bit surprising maybe?

jan-ivar:
I see a couple of options:
1. Allow writes after unsettled atomic writes and just warn users this may reorder things (skip the atomic data)
2. Error the stream with `QuotaExceededError` at the point where reordering would otherwise occur

This could all be done inside the underlying sink, without needing to override `write()`, removing that wrinkle.

jan-ivar:
I've updated the PR to go with option 2 above.

index.bs
```diff
@@ -1620,18 +1640,27 @@ To <dfn for="WebTransportSendStream">write</dfn> |chunk| to a {{WebTransportSend
...
     1. Set |stream|.{{[[PendingOperation]]}} to null.
+    1. If |stream|.{{WebTransportSendStream/[[AtomicWriteRequests]]}} [=list/contains=] |inFlightWriteRequest|, [=list/remove=] |inFlightWriteRequest|.
     1. [=Resolve=] |promise| with undefined.
 1. Return |promise|.
 
 Note: The [=fulfilled|fulfillment=] of the promise returned from this algorithm (or,
-{{WritableStreamDefaultWriter/write|WritableStreamDefaultWriter.write}}) does **NOT** necessarily mean that the chunk is acked by
+{{WritableStreamDefaultWriter/write(chunk)}}) does **NOT** necessarily mean that the chunk is acked by
 the server [[!QUIC]]. It may just mean that the chunk is appended to the buffer. To make sure that
 the chunk arrives at the server, use an application-level protocol.
```
martinthomson, line 1665:
```suggestion
the chunk arrives at the server, the server needs to send some form of acknowledgment message.
```

index.bs
martinthomson: I think that Adam's concern here is similar to mine.  You have a bunch of text about interleaving and whatnot, which implies that maybe you send atomic data in part.  The atomic bit really needs to come before all of that.

I would move the atomic piece up to line 1619, ahead of all of the other stuff.

index.bs
```diff
@@ -2101,6 +2144,65 @@ object |transport|, and a |sendOrder|, run these steps.
...
+:: The {{atomicWrite}} method will reject if the |chunk| given to it
+   could not be sent in its entirety within the [=flow control=] window that
+   is current at the time of sending. This behavior is designed to satisfy niche
+   transactional applications sensitive to [=flow control=] deadlocks ([[RFC9308]]
+   [Section 4.4](https://datatracker.ietf.org/doc/html/rfc9308#section-4.4)).
+
+   Note: {{atomicWrite}} might have partially sent a chunk before rejecting. It
+   does not guarantee atomicity on the wire (e.g. MTU) or at the server, nor does
+   it prevent interleaving of data with other send requests. Only the client will
+   know if it failed.
```
martinthomson, line 2174:
```suggestion
   Note: {{atomicWrite}} can still reject after sending some data. Though it
   provides atomicity with respect to flow control, other errors may occur. 
   {{atomicWrite}} does not prevent data from being split between packets
   or being interleaved with other data. Only the sender learns if
   {{atomicWrite}} fails due to lack of available flow control credit.
```

index.bs
```diff
@@ -2101,6 +2144,65 @@ object |transport|, and a |sendOrder|, run these steps.
...
+
+   Note: Atomic writes can still block if queued behind non-atomic writes. If
+   the atomic write is rejected, everything queued behind it at that moment
+   will be rejected as well. Any non-atomic writes rejected in this way will
+   [=WritableStream/error=] the stream. Applications are therefore encouraged to
+   always await atomic writes.
+
+   When atomicWrite is called, the user agent MUST run the following steps:
```
martinthomson, line 2182:
```suggestion
   When {{atomicWrite}} is called, the user agent MUST run the following steps:
```
?  Not sure about what the rules are for this.

martinthomson (approved):

index.bs
jan-ivar: @martinthomson Thanks for clarifying. So the intent truly is to guarantee nothing is sent unless all can be sent? This would seem to limit atomicWrite chunk sizes to < the current flow control window, but I'm OK with that.

index.bs
```diff
@@ -1510,6 +1512,12 @@ The {{WebTransportSendStream}}'s [=transfer steps=] and
            1. [=Resolve=] |p| with |stats|.
      1. Return |p|.
 
+: <dfn for="WebTransportSendStream" method>getWriter()</dfn>
+:: This method must be implemented the same as {{WritableStream/getWriter}}
```
nidhijaju, line 1516:
```suggestion
:: This method must be implemented in the same manner as {{WritableStream/getWriter}}
```

index.bs
```diff
@@ -1620,20 +1644,22 @@ To <dfn for="WebTransportSendStream">write</dfn> |chunk| to a {{WebTransportSend
...
+    1. If |stream|.{{WebTransportSendStream/[[AtomicWriteRequests]]}} [=list/contains=] |inFlightWriteRequest|, [=list/remove=] |inFlightWriteRequest|.
     1. [=Resolve=] |promise| with undefined.
 1. Return |promise|.
 
 Note: The [=fulfilled|fulfillment=] of the promise returned from this algorithm (or,
-{{WritableStreamDefaultWriter/write|WritableStreamDefaultWriter.write}}) does **NOT** necessarily mean that the chunk is acked by
+{{WritableStreamDefaultWriter/write(chunk)}}) does **NOT** necessarily mean that the chunk is acked by
 the server [[!QUIC]]. It may just mean that the chunk is appended to the buffer. To make sure that
-the chunk arrives at the server, use an application-level protocol.
+the chunk arrives at the server, the server needs to send some form of acknowledgment message.
```
nidhijaju, line 1662:
```suggestion
the chunk arrives at the server, the server needs to send an application-level acknowledgment message.
```
I think we should keep "application-level" to make it clear that a transport-level ack is not sufficient.

index.bs
```diff
@@ -2101,6 +2141,66 @@ object |transport|, and a |sendOrder|, run these steps.
 
 </div>
 
+# `WebTransportWriter` Interface #  {#web-transport-writer-procedures-interface}
```
nidhijaju, line 2144:
```suggestion
# `WebTransportWriter` Interface #  {#web-transport-writer-interface}
```
nit: I'm not sure we need procedures here (and below)

index.bs
```diff
@@ -2101,6 +2141,66 @@ object |transport|, and a |sendOrder|, run these steps.
...
+[=WebTransportWriter/create=] procedure.
+
+<pre class="idl">
+[Exposed=*, SecureContext]
+interface WebTransportWriter : WritableStreamDefaultWriter {
+  Promise&lt;undefined&gt; atomicWrite(optional any chunk);
+};
+</pre>
+
+## Methods ##  {#web-transport-writer-procedures-methods}
```
nidhijaju, line 2159:
```suggestion
## Methods ##  {#web-transport-writer-methods}
```

index.bs
```diff
@@ -2101,6 +2141,66 @@ object |transport|, and a |sendOrder|, run these steps.
...
+   1. Let |p| be the result of {{WritableStreamDefaultWriter/write(chunk)}}
+       on {{WritableStreamDefaultWriter}} with |chunk|.
+   1. [=set/Append=] |p| to |stream|.{{WebTransportSendStream/[[AtomicWriteRequests]]}}.
+   1. Return the result of [=reacting=] to |p| with the following steps:
+      1. If |stream|.{{WebTransportSendStream/[[AtomicWriteRequests]]}} [=list/contains=] |p|,
+         [=list/remove=] |p|.
+      1. If |p| was rejected with reason |r|, then return [=a promise rejected with=] |r|.
+      1. Return undefined.
+
+## Procedures ##  {#web-transport-writer-procedures-procedures}
```
nidhijaju, line 2190:
```suggestion
## Procedures ##  {#web-transport-writer-procedures}
```

index.bs
```diff
@@ -2666,6 +2766,31 @@ async function receiveText(url, createWritableStreamForTextData) {
 }
 </pre>
 
+## Sending a transactional chunk on a stream ##  {#example-transactional-stream}
+
+*This section is non-normative.*
+
+Sending a transactional piece of data on a one-way stream only if it can be done
```
nidhijaju, line 2773:
```suggestion
Sending a transactional piece of data on a unidirectional stream, only if it can be done
```

nidhijaju (approved):
Mostly looks good to me, other than nits.

PR #558, opened 2023-10-20
Reject getStats() promise when the connection is closed/failed.

### LINKED ISSUE #557: getStats() being rejected
Currently, we never reject getStats(). We should do that for failed connections, since those don't have any stats to report.  We should also probably do that for closed connections too, as the state associated with the connection is already likely gone.

vasilvv:
Fixes #557

***
<a href="https://pr-preview.s3.amazonaws.com/vasilvv/web-transport/pull/558.html" title="Last updated on Nov 20, 2023, 10:57 AM UTC (5dae276)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/558/967a740...vasilvv:5dae276.html" title="Last updated on Nov 20, 2023, 10:57 AM UTC (5dae276)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -987,6 +987,8 @@ these steps.
      1. Run the following steps [=in parallel=]:
          1. Gather the stats from the [=underlying connection=], including stats on datagrams.
          1. [=Queue a network task=] with |transport| to run the following steps:
+           1. If |transport|.{{[[State]]}} is `"closed"` or `"failed"`,
+              [=reject=] |p| with an {{InvalidStateError}} and abort these steps.
            1. Let |stats| be a [=new=] {{WebTransportConnectionStats}} object representing the gathered stats.
            1. [=Resolve=] |p| with |stats|.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -987,6 +987,8 @@ these steps.
      1. Run the following steps [=in parallel=]:
          1. Gather the stats from the [=underlying connection=], including stats on datagrams.
          1. [=Queue a network task=] with |transport| to run the following steps:
+           1. If |transport|.{{[[State]]}} is `"closed"` or `"failed"`,
+              [=reject=] |p| with an {{InvalidStateError}} and abort these steps.
```
jan-ivar, line 991: I think we should do the same state check in the synchronous steps (since the timing of promise state is synchronously observable in a JS debugger).

jan-ivar:
Meeting:
- Reject on failed
- Record final stats on close and return those after close.

PR #568, opened 2023-11-14
Add feature-detection of HTTP/2 fallback.

### LINKED ISSUE #518: WebTransport http2 idea's suggestions
I have been working on a node webtransport plugin (build around quiche (the google one)) and a polyfill for webtransport on unsupported browsers or networks.

The polyfill currently uses for every session and every stream has its own web socket connection. The streams and sessions are tied together by initial signed tokens, that authorize and tied together the websession connections. Data within the streams are sent as binary WebSocket messages and control messages (equivalents to quic messages) as text messages.
It does not suffer head-of-line blocking, but probably the higher number of TCP connections is not beneficial for large setups (but I have no experience with this, pure speculation).

Anyway, I did this, before finding the http2 proposal (I do not know if it was at this point already in the spec).
As my main motivation for the polyfill is to

1. Support browsers, that do not implement webtransport (yet?).
2. Work in environments, where UDP is potentially blocked. (This is true for most of the network, I will deploy my application, which is my motivation for the webtransport work)

So I took a look on the current http/2 spec for webtransport including the capsule protocol. I think it will not be hard to add it to my node plugin using node's native http/2 implementation. So I will probably do it in the next time. 
But my motivation is not so high, since there is currently no browser implementation, so the use case is pretty limited.
And the design of the capsule protocol prevents use case 1.

So I was wondering if it would make sense to add a capsule over websocket protocol (only for the server side) to the specifications.
The benefits are as follows:

1. The protocol's way of working would be very similar to the actual http/2, so one could easily test ideas without waiting for the browser implementation.
2. It will not be hard to implement it in a shim/polyfill (or I think it comes naturally out from my node plugins work)
3. Having it specified ensures, that one shim/polyfill is enough for all future servers deployed that comply to the spec.
4. And it will automatically address points 1. and 2. above.

My proposal would be as follows, the normal capsule protocol:
```
Capsule {
     Capsule Type (i),
     Capsule Length (i),
     Capsule Value (..),
   }
```
The idea would be to transport is as binary frame/payload of length `Length + 4 `, and the payload will start with capsule type and continue with capsule value. Maybe an initial text frame indicates, that it is a capsule over websocket connection.
The rest of the http/2 capsule protocol will remain unchanged. So besides a little bit extra overload, the mechanics of most of the protocol will remain unchanged.

One thing, good for a shim/polyfill would be a mechanism to identify, if the webtransport implementation actually supports the http/2 protocol, so that it can otherwise wrap around and provide a fallback mechanism. Or is this possible and I missed it?

jan-ivar:
Fixes #518 (draft for discussion).

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webtransport/pull/568.html" title="Last updated on Nov 14, 2023, 8:45 PM UTC (9e82560)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/568/c177c1e...jan-ivar:9e82560.html" title="Last updated on Nov 14, 2023, 8:45 PM UTC (9e82560)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -828,6 +828,12 @@ sequence&lt;{{WebTransportHash}}&gt; |serverCertificateHashes|, run these steps.
   1. If |connection| is failure, then abort the remaining steps and [=queue a network task=] with
      |transport| to run these steps:
     1. If |transport|.{{[[State]]}} is `"closed"` or `"failed"`, then abort these steps.
+    1. If the user agent supports [=WebTransport sessions=] over HTTP/2 [=connections=], set
+       |transport|'s {{[[Reliability]]}} to `"reliable-only"`, otherwise set it to
+       `"supports-unreliable"`.
+
+       Note: This allows applications to feature-detect HTTP/2 fallback.
+
     1. Let |error| be a newly [=DOMException/created=] {{WebTransportError}} whose
        {{WebTransportErrorOptions/source}} is `"session"`.
     1. [=Cleanup=] |transport| with |error|.


jan-ivar:
Meeting:
- Alternative idea: New API to return array of what was attempted "reliable" and/or "unreliable", or two booleans.
- Alternative idea: WebTransport.supportsReliableOnly

jan-ivar:
Overtaken by #575.

PR #573, opened 2023-11-21
Make estimatedSendRate be `null` instead of `undefined`

### LINKED ISSUE #554: estimatedSendRate absent vs null
The spec currently says "If the user agent does not currently have an estimate, the member MUST be absent", which I assume means it has to be `undefined`; but the type is declared as nullable, which I assume means it is supposed to be null, and not undefined (or otherwise we never null it).

vasilvv:
Fixes #554

***
<a href="https://pr-preview.s3.amazonaws.com/vasilvv/web-transport/pull/573.html" title="Last updated on Nov 21, 2023, 3:27 PM UTC (0794b98)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/573/c177c1e...vasilvv:0794b98.html" title="Last updated on Nov 21, 2023, 3:27 PM UTC (0794b98)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1395,8 +1395,8 @@ The dictionary SHALL have the following attributes:
    This rate applies to all streams and datagrams that share a [=WebTransport session=]
    and is calculated by the congestion control algorithm (potentially chosen by
    {{WebTransport/congestionControl}}). If the user agent does not
-   currently have an estimate, the member MUST be [=map/exist|absent=].
-   The member can be absent even if present in previous results.
+   currently have an estimate, the member MUST be the ECMAScript `null` value.
+   The member can be `null` even if it was not `null` in previous results.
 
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1395,8 +1395,8 @@ The dictionary SHALL have the following attributes:
    This rate applies to all streams and datagrams that share a [=WebTransport session=]
    and is calculated by the congestion control algorithm (potentially chosen by
    {{WebTransport/congestionControl}}). If the user agent does not
-   currently have an estimate, the member MUST be [=map/exist|absent=].
-   The member can be absent even if present in previous results.
+   currently have an estimate, the member MUST be the ECMAScript `null` value.
```
jan-ivar, line 1398:
```suggestion
   currently have an estimate, the member MUST be the `null` value.
```

jan-ivar (approved):

PR #580, opened 2024-01-23
Rename "Data Recvd" to "all data committed"

martinthomson:
As discussed in https://github.com/ietf-wg-webtrans/draft-ietf-webtrans-http2/issues/103

This is NOT the only place where this state is used.  That one needs a few more grey matter cycles before I have a resolution.  I'll open an issue instead.

***
<a href="https://pr-preview.s3.amazonaws.com/martinthomson/webtransport/pull/580.html" title="Last updated on Jan 14, 2025, 3:38 PM UTC (ba41dea)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/580/607bfa8...martinthomson:ba41dea.html" title="Last updated on Jan 14, 2025, 3:38 PM UTC (ba41dea)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1656,6 +1656,7 @@ To <dfn for="WebTransportSendStream">close</dfn> a {{WebTransportSendStream}} |s
 1. Run the following steps [=in parallel=]:
   1. [=stream/Send=] FIN on |stream|.{{WebTransportSendStream/[[InternalStream]]}} and wait for the operation to
      complete.
+  1. Wait for |stream|.{{WebTransportSendStream/[[InternalStream]]}} to enter the "all data committed" state. [[!QUIC]]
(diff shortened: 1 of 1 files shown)


ekinnear (approved):

vasilvv:
I believe this will mean that the API will be unable to reset streams that the sender has already sent a FIN on, which is why that step has been added in the first place.

jan-ivar:
Meeting:
- Useful for aborting a stream to create "partial reliable" sending (to abandon data you already sent)
- May need to keep a signal in HTTP/3 and not HTTP/2 (or add it, not useful there though)

jan-ivar:
Meeting:
- abandon PR?

PR #598, opened 2024-04-08
Add protocols constructor argument to WebTransport + protocol read arg

### LINKED ISSUE #536: Reconsider whether we need sub-protocol negotiation
Both WebSocket and QUIC have a subprotocol negotiation mechanism (QUIC mandates ALPN, WebSocket has Sec-WebSocket-Protocol).  When we designed WebTransport originally, we decided that between the application streams and the URL path/query, those are redundant.

Reasons why we may want to reconsider this:
- People design applications for QUIC, and they may rely on ALPN for things like version negotiation.  Supporting a similar mechanism in WebTransport would make porting those protocols easier.
- QUIC requires ALPN for protection against cross-protocol attacks, which may be a useful property here too.

Reasons why we may want to keep it as-is:
- Mitigating cross-protocol attacks might not be particularly useful.  While HTTP itself does support a similar mechanism (Content-Type/Accept), Web developers often assume that the URL implies the format it contains (i.e. if you have an endpoint called `/api/update-entry`, you know what can be POSTed onto it, and what format the response will be).

jan-ivar:
Fixes #536. Right now this seems only defined for HTTP/3, so that's what I went with. Is this what we want?

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webtransport/pull/598.html" title="Last updated on Sep 17, 2024, 3:13 AM UTC (6b82382)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/598/55eb163...jan-ivar:6b82382.html" title="Last updated on Sep 17, 2024, 3:13 AM UTC (6b82382)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -181,7 +181,7 @@ A [=WebTransport session=] has the following capabilities defined in [[!WEB-TRAN
 </table>
 
 To <dfn for=session>establish</dfn> a [=WebTransport session=] with an [=/origin=] |origin| and a
-|protocols| array, follow [[!WEB-TRANSPORT-OVERVIEW]]
+|subprotocols| array, follow [[!WEB-TRANSPORT-OVERVIEW]]
 [Section 4.1](https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-overview-06#section-4.1-2.2.1),
 with using |origin|, [=ASCII serialization of an origin|serialized=] and [=isomorphic encoded=],
 as the [:Origin:] header of the request.
--- a/index.bs
+++ b/index.bs
@@ -904,9 +904,9 @@ sequence&lt;{{WebTransportHash}}&gt; |serverCertificateHashes|, run these steps.
       1. Abort these steps.
     1. Set |transport|.{{[[State]]}} to `"connected"`.
     1. Set |transport|.{{[[Session]]}} to |session|.
-    1. If the connection is an HTTP/3 connection, set |transport|'s {{[[Reliability]]}} to `"supports-unreliable"`,
-       and set {{[[Protocol]]}} to the `WebTransport-Subprotocol` header field in the 2xx response to the
-       CONNECT request if present, following [[!WEB-TRANSPORT-HTTP3]]
+    1. If the connection is an HTTP/3 connection, set |transport|.{{[[Reliability]]}} to `"supports-unreliable"`,
+       and set |transport|.{{[[Protocol]]}} to the `WebTransport-Subprotocol` header field in the 2xx response to
+       the CONNECT request if present, following [[!WEB-TRANSPORT-HTTP3]]
        [Section 3.4](https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3/#section-3.4).
     1. If the connection is an HTTP/2 connection [[!WEB-TRANSPORT-HTTP2]], set |transport|'s {{[[Reliability]]}} to `"reliable-only"`.
     1. [=Resolve=] |transport|.{{[[Ready]]}} with undefined.
--- a/index.bs
+++ b/index.bs
@@ -181,8 +181,7 @@ A [=WebTransport session=] has the following capabilities defined in [[!WEB-TRAN
 </table>
 
 To <dfn for=session>establish</dfn> a [=WebTransport session=] with an [=/origin=] |origin| and a
-|protocols| array,
-follow [[!WEB-TRANSPORT-OVERVIEW]]
+|protocols| array, follow [[!WEB-TRANSPORT-OVERVIEW]]
 [Section 4.1](https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-overview-06#section-4.1-2.2.1),
 with using |origin|, [=ASCII serialization of an origin|serialized=] and [=isomorphic encoded=],
 as the [:Origin:] header of the request.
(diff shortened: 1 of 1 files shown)


jan-ivar:
Meeting:
- s/protocols/subprotocols/
- Make sure to preserve order in header field list.

index.bs
```diff
@@ -703,6 +708,11 @@ A {{WebTransport}} object has the following internal slots.
    <td class="non-normative">The number of concurrently open
    [=bidirectional=] streams the application anticipates the server creating, or null.
   </tr>
+  <tr>
+   <td><dfn>`[[Protocol]]`</dfn>
+   <td class="non-normative">A string indicating the subprotocol selected by the server,
```
martinthomson, line 713: Do we have a reference for "subprotocol"?  I know that it's a websocket concept, so maybe that can be used.

index.bs
```diff
@@ -749,6 +759,14 @@ agent MUST run the following steps:
    congestion control algorithms that optimize for |congestionControl|, as allowed by
    [[!RFC9002]] [Section 7](https://www.rfc-editor.org/rfc/rfc9002#section-7),
    then set |congestionControl| to `"default"`.
+1. Let |protocols| be {{WebTransport/constructor(url, options)/options}}'s
+   {{WebTransportOptions/protocols}}
+1. If any of the values in |protocols| occur more than once, fail to match
+   the requirements for elements that comprise the value of `WebTransport-Subprotocol`
+   fields as defined by the WebTransport protocol, or have an [=isomorphic encoded=]
+   length exceeding 512, [=throw=] a {{SyntaxError}} exception.
```
martinthomson, line 767: I think that you need to reject zero-length protocols explicitly.

index.bs
```diff
@@ -749,6 +759,14 @@ agent MUST run the following steps:
    congestion control algorithms that optimize for |congestionControl|, as allowed by
    [[!RFC9002]] [Section 7](https://www.rfc-editor.org/rfc/rfc9002#section-7),
    then set |congestionControl| to `"default"`.
+1. Let |protocols| be {{WebTransport/constructor(url, options)/options}}'s
+   {{WebTransportOptions/protocols}}
+1. If any of the values in |protocols| occur more than once, fail to match
+   the requirements for elements that comprise the value of `WebTransport-Subprotocol`
```
martinthomson, line 765: Can we drop the "sub" and use `WebTransport-Protocol` instead?

index.bs
jan-ivar: In WebSocket, the options-member is [protocols](https://websockets.spec.whatwg.org/#dom-websocket-websocket), but refers to each string as a _"subprotocol name"_ that has to _"match the requirements for elements that comprise the value of [`Sec-WebSocket-Protocol`](https://datatracker.ietf.org/doc/html/rfc6455#section-11.3.4) fields as defined by The WebSocket protocol. [[WSP]](https://websockets.spec.whatwg.org/#biblio-wsp)"_

image

WebSockets is mostly what I modeled this PR on. We could s/subprotocol/subprotocol name/ maybe if that helps?

I could also break out "subprotocol name" as a definition if that helps?

index.bs
jan-ivar: We seem to be going the other way naming-wise judging from other feedback I've received.

But here specifically it needs to match https://github.com/ietf-wg-webtrans/draft-ietf-webtrans-http3/pull/144 which is already merged.

index.bs
```diff
@@ -180,13 +180,17 @@ A [=WebTransport session=] has the following capabilities defined in [[!WEB-TRAN
...
-To <dfn for=session>establish</dfn> a [=WebTransport session=] with an [=/origin=] |origin|,
-follow [[!WEB-TRANSPORT-OVERVIEW]]
+To <dfn for=session>establish</dfn> a [=WebTransport session=] with an [=/origin=] |origin| and a
+|protocols| array, follow [[!WEB-TRANSPORT-OVERVIEW]]
 [Section 4.1](https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-overview-06#section-4.1-2.2.1),
 with using |origin|, [=ASCII serialization of an origin|serialized=] and [=isomorphic encoded=],
 as the [:Origin:] header of the request.
 When establishing a session, the client MUST NOT provide any [=credentials=].
 The resulting underlying transport stream is referred to as the session's <dfn>CONNECT stream</dfn>.
+Additionally, if the [=underlying connection=] is using HTTP/3 and the |protocols| array is non-empty,
```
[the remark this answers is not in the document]
jan-ivar: Agree this should be the goal. @nidhijaju or @vasilvv can you open this issue? But in the meantime, can we iterate here by merging what we have, and come back and fix this when it's available in the overview doc?

jan-ivar:
Meeting:
- @nidhijaju to open issue on HTTP/2 and overview doc
- then update this PR to reference overview doc and merge.

jan-ivar:
Meeting:
- protocols or subprotocols?

wilaw:
Could we call this "applicationProtocol"? This avoids overloading the multiple references to internet protocols (UDP, QUIC, WebTransport) and instead refers to the protocol we wish to describe, which is defined completely by the application using the WebTransport connection.

jan-ivar:
Meeting:
- Two votes (plus martin) for protocols instead of subprotocols
- or applicationProtocols
- If we change it we should change the IETF specs to match.
- Our spec says "protocol" a lot to refer to the underlying/IETF parts
- Leave it subprotocols for now and ask the IETF if they want to change it to protocols

jan-ivar:
IETF preferred "protocol" by overwhelming margin in a poll (19 protocol, 3 subprotocol, 9 applicationLevelProtocol)

wilaw:
Note https://github.com/ietf-wg-webtrans/draft-ietf-webtrans-overview/issues/15#issuecomment-2299876480, calling for harmonization around the use of "protocol".

jan-ivar:
The API is now "protocol" throughout, except prose still uses _"subprotocol"_ to disambiguate it from its own [§ 3. Protocol concepts](https://w3c.github.io/webtransport/#protocol-concepts) (HTTP3 does a [similar thing](https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3/#name-subprotocol-negotiation)).

jan-ivar:
s/subprotocol/protocol/ for variable names

nidhijaju (commented):
s/subprotocol/protocol based on last meeting

index.bs
```diff
@@ -1006,6 +1041,12 @@ Note: Setting {{WebTransport/anticipatedConcurrentIncomingUnidirectionalStreams}
 {{WebTransport/anticipatedConcurrentIncomingBidirectionalStreams}} does not guarantee
 the application will receive the number of streams it anticipates.
 
+: <dfn for="WebTransport" attribute>protocol</dfn>
+:: Once a [=WebTransport session=] has been established and the {{protocols}}
+   constructor option was used to provide a non-empty array, returns the subprotocol
```
nidhijaju, line 1046:
```suggestion
   constructor option was used to provide a non-empty array, returns the application-level protocol
```

PR #603, opened 2024-05-21
Add stats.atSendCapacity

### LINKED ISSUE #559: Quality of a bandwidth estimate
There might be utility in exposing some kind of "quality" indication of a bandwidth estimate, indicating how close it is to the "real" bandwidth of the channel.

There are few signals that congestion controller could expose as an API that I can think of:
- Has there ever been a bandwidth sample that is not application-limited?
- Is the connection out of slow start?

cc @ianswett

jan-ivar:
Fixes #559. For discussion.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webtransport/pull/603.html" title="Last updated on Oct 8, 2024, 11:22 PM UTC (44a80cb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/603/3363834...jan-ivar:44a80cb.html" title="Last updated on Oct 8, 2024, 11:22 PM UTC (44a80cb)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1534,8 +1534,10 @@ The dictionary SHALL have the following attributes:
    currently have an estimate, the member MUST be the `null` value.
    The member can be `null` even if it was not `null` in previous results.
 : <dfn for="WebTransportConnectionStats" dict-member>atSendCapacity</dfn>
-:: A value of false indicates the application is sending less data than available
-   network capacity, and the {{estimatedSendRate}} might be application limited.
+:: A value of false indicates the {{estimatedSendRate}} might be application limited,
+   meaning the application is sending significantly less data than the congestion
+   controller allows. A congestion controller might produce a poor estimate of the
+   available network capacity while it is application limited.
 
    A value of true indicates the application is sending data at network
    capacity, and the {{estimatedSendRate}} reflects the network capacity available
--- a/index.bs
+++ b/index.bs
@@ -1408,7 +1408,7 @@ dictionary WebTransportCloseInfo {
 };
 </pre>
 
-The dictionary SHALL have the following members:
+The dictionary SHALL have the following attributes:
 
 : <dfn for="WebTransportCloseInfo" dict-member>closeCode</dfn>
 :: The error code communicated to the peer.
@@ -1430,7 +1430,7 @@ dictionary WebTransportSendStreamOptions {
 };
 </pre>
 
-The dictionary SHALL have the following members:
+The dictionary SHALL have the following attributes:
 
 : <dfn for="WebTransportSendStreamOptions" dict-member>sendGroup</dfn>
 :: An optional {{WebTransportSendGroup}} to [=group=] this
@@ -1486,11 +1486,13 @@ dictionary WebTransportConnectionStats {
   DOMHighResTimeStamp rttVariation;
   DOMHighResTimeStamp minRtt;
   WebTransportDatagramStats datagrams;
-  WebTransportSendRateStats? sendRate;
+  unsigned long long? estimatedSendRate;
+  boolean isSendDataLimited;
+  boolean inSlowStart;
 };
 </pre>
 
-The dictionary SHALL have the following members:
+The dictionary SHALL have the following attributes:
 
 : <dfn for="WebTransportConnectionStats" dict-member>bytesSent</dfn>
 :: The number of bytes sent on the [=underlying connection=], including retransmissions.
@@ -1517,13 +1519,20 @@ The dictionary SHALL have the following members:
    [Section 5.3](https://www.rfc-editor.org/rfc/rfc9002#section-5.3).
 : <dfn for="WebTransportConnectionStats" dict-member>minRtt</dfn>
 :: The minimum round-trip time observed on the entire connection.
-: <dfn for="WebTransportConnectionStats" dict-member>datagrams</dfn>
-:: Contains statistics on datagram transmission over the [=underlying connection=].
-: <dfn for="WebTransportConnectionStats" dict-member>sendRate</dfn>
-:: Contains congestion control information such as an estimated send
-   rate. If the user agent does not currently have this information,
-   the member MUST be the `null` value. The member can be `null` even
-   if it was not `null` in previous results.
+: <dfn for="WebTransportConnectionStats" dict-member>estimatedSendRate</dfn>
+:: The estimated rate at which queued data will be sent by the user agent, in bits per second.
+   This rate applies to all streams and datagrams that share a [=WebTransport session=]
+   and is calculated by the congestion control algorithm (potentially chosen by
+   {{WebTransport/congestionControl}}). If the user agent does not
+   currently have an estimate, the member MUST be the `null` value.
+   The member can be `null` even if it was not `null` in previous results.
+: <dfn for="WebTransportConnectionStats" dict-member>isSendDataLimited</dfn>
+:: True if the rate at which queued data will be sent by the user agent on all streams
+   and datagrams that share a [=WebTransport session=], as calculated by the
+   congestion control algorithm, is considered application limited. Otherwise false.
+: <dfn for="WebTransportConnectionStats" dict-member>inSlowStart</dfn>
+:: True if the current congestion control algorithm implements "slow start" and the
+   [=WebTransport session=] is currently in this mode. Otherwise false.
 
 ## `WebTransportDatagramStats` Dictionary ##  {#web-transport-datagram-stats}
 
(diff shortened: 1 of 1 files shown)


jan-ivar:
These are modeled on [estimatedSendRate](https://w3c.github.io/webtransport/#dom-webtransportconnectionstats-estimatedsendrate), the lone stat specific to the [WebTransport session](https://w3c.github.io/webtransport/#protocol-webtransport-session), as opposed to all the other stats which are for the entire [underlying connection](https://w3c.github.io/webtransport/#underlying-connection).

I dunno if that was a mistake or not. We should discuss.

jan-ivar:
Meeting:
- take out "datagrams" from isServerLimited
- Maybe add `required` to all four?

index.bs
```diff
@@ -1557,6 +1556,40 @@ The dictionary SHALL have the following attributes:
...
+   {{WebTransport/congestionControl}}).
+: <dfn for="WebTransportSendRateStats" dict-member>isDataLimited</dfn>
+:: True if the rate at which queued data will be sent by the user agent on all streams
+   that share a [=WebTransport session=], as calculated by the
+   congestion control algorithm, is considered application limited. Otherwise false.
+   Datagrams MUST NOT be included in this calculation.
+: <dfn for="WebTransportSendRateStats" dict-member>isServerLimited</dfn>
+:: True if the rate at which queued data will be sent by the user agent on all streams
+   and datagrams that share a [=WebTransport session=], as calculated by the
+   congestion control algorithm, is considered limited by back pressure from the server. Otherwise false.
```
jan-ivar, line 1588:
```suggestion
:: True if the rate at which queued data will be sent by the user agent on all streams
   that share a [=WebTransport session=], as calculated by the
   congestion control algorithm, is considered limited by back pressure from the server. Otherwise false.
   Datagrams MUST NOT be included in this calculation.
```

index.bs
```diff
@@ -1557,6 +1556,41 @@ The dictionary SHALL have the following attributes:
...
+: <dfn for="WebTransportSendRateStats" dict-member>estimate</dfn>
+:: The estimated rate at which queued data will be sent by the user agent, in bits per second.
+   This rate applies to all streams and datagrams that share a [=WebTransport session=]
+   and is calculated by the congestion control algorithm (potentially chosen by
+   {{WebTransport/congestionControl}}).
+: <dfn for="WebTransportSendRateStats" dict-member>isDataLimited</dfn>
+:: True if the rate at which queued data will be sent by the user agent on all streams
+   that share a [=WebTransport session=], as calculated by the
+   congestion control algorithm, is considered application limited. Otherwise false.
+   Datagrams MUST NOT be included in this calculation.
```
jan-ivar, line 1584:
```suggestion
:: True if the rate at which queued data will be sent by the user agent on all streams
   and datagrams that share a [=WebTransport session=], as calculated by the
   congestion control algorithm, is considered application limited. Otherwise false.
```

jan-ivar:
Meeting:
- interest in going back to flat list of members

Right now we have:
```js
  stats.sendRate.estimate;
  stats.sendRate.isDataLimited;
  stats.sendRate.isServerLimited;
  stats.sendRate.inSlowStart;
```
- Add a second PR to consider a flat list of members instead:
```js
  stats.estimatedSendRate;
  stats.isSendingDataLimited;
  stats.isSendingServerLimited;
  stats.inSendinginSlowStart;
```

index.bs
```diff
@@ -1557,6 +1556,40 @@ The dictionary SHALL have the following attributes:
...
+   and is calculated by the congestion control algorithm (potentially chosen by
+   {{WebTransport/congestionControl}}).
+: <dfn for="WebTransportSendRateStats" dict-member>isDataLimited</dfn>
+:: True if the rate at which queued data will be sent by the user agent on all streams
+   and datagrams that share a [=WebTransport session=], as calculated by the
+   congestion control algorithm, is considered application limited. Otherwise false.
+: <dfn for="WebTransportSendRateStats" dict-member>isServerLimited</dfn>
+:: True if the rate at which queued data will be sent by the user agent on all streams
+   that share a [=WebTransport session=], as calculated by the
+   congestion control algorithm, is considered limited by back pressure from the server. Otherwise false.
```
jan-ivar, line 1587:
```suggestion
   congestion control algorithm, is considered limited by back pressure from the server. Otherwise false.
   This is based on the server receive window.
```

jan-ivar:
Personally, I'm actually partial to a flat hierarchy a bit, since a hierarchy doesn't really solve ambiguity that much due the `?.` operator in JS. Example:
 
This will throw if `stats` or `stats.sendRate` is undefined:
```js
stats.sendRate.isDataLimited;
```

This result in undefined (which might interpreted as false by websites):
```js
stats.sendRate?.isDataLimited;
```

jan-ivar:
@wilaw wrote in https://github.com/w3c/webtransport/issues/607#issuecomment-2206558189:
> In the prior meeting, a comment was made that some of the proposed 'sending' stats may not be known, and that coercing a boolean value could cause a problem. Is this a real problem, or can robust defaults be chosen that will not mislead applications?

(replying here about stats we're still adding here)

If unavailable stats are mostly a problem at connection start, then defaults reflecting initial state should work.

> stats.estimatedSendRate; //nullable
> stats.isSendingDataLimited; // default false

A connection is inherently data limited to start, so this would naturally default to (start out as) true, unless we can think of a name for the inverse state.

> stats.isSendingServerLimited; // default false

Is it possible for a session to be both data limited AND server limited? If I have two long-lived streams, one server limited (receiving back pressure from the server) and the other data limited, would both be true? How would I know which applies to which?

> stats.inSendingSlowStart; // default false
> stats.blockedFramesCount; // default 0

These seem fine to me.

jan-ivar:
Meeting:
- no objections to flat

index.bs
```diff
@@ -1494,7 +1494,8 @@ dictionary WebTransportConnectionStats {
   required DOMHighResTimeStamp rttVariation;
   required DOMHighResTimeStamp minRtt;
   required WebTransportDatagramStats datagrams;
-  required unsigned long long? estimatedSendRate;
+  required unsigned long long? estimatedSendRate = null;
```
jan-ivar, line 1497:
```suggestion
  unsigned long long? estimatedSendRate = null;
```

index.bs
```diff
@@ -1494,7 +1494,8 @@ dictionary WebTransportConnectionStats {
   required DOMHighResTimeStamp rttVariation;
   required DOMHighResTimeStamp minRtt;
   required WebTransportDatagramStats datagrams;
-  required unsigned long long? estimatedSendRate;
+  unsigned long long? estimatedSendRate = null;
+  required boolean isSendDataLimited = true;
```
jan-ivar, line 1498:
```suggestion
 boolean isSendDataLimited = true;
```

index.bs
```diff
@@ -1494,7 +1494,8 @@ dictionary WebTransportConnectionStats {
   required DOMHighResTimeStamp rttVariation;
   required DOMHighResTimeStamp minRtt;
   required WebTransportDatagramStats datagrams;
-  required unsigned long long? estimatedSendRate;
+  unsigned long long? estimatedSendRate = null;
+ boolean isSendDataLimited = true;
```
jan-ivar, line 1498:
```suggestion
  boolean isSendDataLimited = true;
```

jan-ivar:
Meeting:
- bikeshed: atSendCapacity = false
- need to keep sending to stay there
- Note: atSendCapacity is not intended to replace the back-pressure signal. It is intended to signal the confidence level that you are at the estimated send rate. Capacity may increase or decrease based on network conditions.

martinthomson:
Including what I typed into chat...

When false, it indicates that estimatedSendRate is not being fully utilized and could be an unreliable estimate of actual network capacity.

When true, it indicates that the data that the application sends is close to the estimatedSendRate.  As long as that send rate is sustained, estimatedSendRate will adapt to network conditions and be a fair estimate of what capacity is available to the application.

index.bs
```diff
@@ -1532,7 +1533,19 @@ The dictionary SHALL have the following attributes:
    {{WebTransport/congestionControl}}). If the user agent does not
    currently have an estimate, the member MUST be the `null` value.
    The member can be `null` even if it was not `null` in previous results.
-
+: <dfn for="WebTransportConnectionStats" dict-member>atSendCapacity</dfn>
+:: A value of false indicates the application is sending less data than available
+   network capacity, and the {{estimatedSendRate}} might be application limited.
```
jan-ivar, line 1538:
```suggestion
:: A value of false indicates the {{estimatedSendRate}} might be application limited,
   meaning the application is sending significantly less data than the congestion
   controller allows. A congestion controller might produce a poor estimate of the
   available network capacity while it is application limited.
```

PR #604, opened 2024-06-03
Give datagrams priority over streams but don't starve them

### LINKED ISSUE #451: Allow datagram send order relative to streams to be application defined
PR #438 defined an optional send order for streams. This allows relative prioritization between streams, however it does not define the relative prioritization between datagrams and streams. 

Consider the use case of a streaming protocol in which a bidi stream is established as a control channel and A/V data is transmitted via datagrams. The current UA implementation always gives datagrams a higher send priority over all streams. In the face of congestion, the A/V data would be sent ahead of any control data.

One solution to this problem would be to add an optional sendOrder attribute to the datagrams.writable.getWriter method:

```
const wt = new WebTransport(url);
const writer = wt.datagrams.writable.getWriter({sendOrder:3});
```

Another might be to add a datagramSendOrder entry to the WebTransportOptions dictionary

```
const wt = new WebTransport(url,{datagramSendOrder:3});
const writer = wt.datagrams.writable.getWriter();
```

If datagram sendOrder is left undefined, then the default prioritization can be either that:
1. Datagrams are always sent ahead of any queued stream data
2. Datagrams are round-robined with other queued stream data that also has an undefined sendOrder.

jan-ivar:
Fixes #451.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1730,6 +1730,7 @@ To <dfn for="WebTransportSendStream">write</dfn> |chunk| to a {{WebTransportSend
      and wait for the operation to complete.
      This sending MAY be interleaved with sending of previously queued streams and datagrams,
      as well as streams and datagrams yet to be queued to be sent over this transport.
+     Datagrams SHOULD be given priority over this sending, but not to the point of starving it.
 
      The user-agent MAY have a buffer to improve the transfer performance. Such a buffer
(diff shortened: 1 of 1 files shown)


jan-ivar:
Meeting:
- merge and keep discussing specifying this further.

PR #619, opened 2024-10-21
Add notes on receiver-side stream availability

### LINKED ISSUE #618: When should a client initiated unidirectional stream be available on the server side?
I am still trying to create working stream-limit tests with all two webtransport enabled browsers.
I ran into another issue: I misused creating uni- and bidirectional streams as a matter of communication/synchronization in the tests. Chrome and Firefox behave quite differently.
If I create a unidirectional stream with
```
await transport.createUnidirectionalStream()
```
Chromium immediately sends out a frame indicating the stream's creation, but Firefox does not (at least for 5 seconds).
I had to push one byte into the unidirectional stream to let the server know about the stream.
So the question is, is the behavior within spec? If not, I would file a bug with Firefox, and if yes, I think this can lead to trouble...

wilaw:
Fixes #618

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/pull/619.html" title="Last updated on Oct 21, 2024, 2:00 PM UTC (898287e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/619/3a324bf...898287e.html" title="Last updated on Oct 21, 2024, 2:00 PM UTC (898287e)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -994,11 +994,17 @@ these steps.
 : <dfn for="WebTransport" attribute>incomingBidirectionalStreams</dfn>
 :: Returns a {{ReadableStream}} of {{WebTransportBidirectionalStream}}s that have been
    received from the server.
+
+   Note: whether the incoming streams already have data on them will depend on server behavior. 
+
    The getter steps for the `incomingBidirectionalStreams` attribute SHALL be:
      1. Return [=this=]'s {{[[IncomingBidirectionalStreams]]}}.
 : <dfn for="WebTransport" attribute>incomingUnidirectionalStreams</dfn>
 :: A {{ReadableStream}} of unidirectional streams, each represented by a
    {{WebTransportReceiveStream}}, that have been received from the server.
+
+   Note: whether the incoming streams already have data on them will depend on server behavior. 
+
    The getter steps for `incomingUnidirectionalStreams` are:
      1. Return [=this=].{{[[IncomingUnidirectionalStreams]]}}.
 
@@ -1110,6 +1116,8 @@ the application will receive the number of streams it anticipates.
    stream.  Note that the mere creation of a stream is not immediately visible to the peer until
    it is used to send data.
 
+   Note: there is no expectation that the server will be aware of the stream until data is sent on it.
+
    When `createBidirectionalStream` is called, the user agent MUST run the
    following steps:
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -994,11 +994,17 @@ these steps.
 : <dfn for="WebTransport" attribute>incomingBidirectionalStreams</dfn>
 :: Returns a {{ReadableStream}} of {{WebTransportBidirectionalStream}}s that have been
    received from the server.
+
+   Note: whether the incoming streams already have data on them will depend on server behavior.
```
wilaw, line 998:
```suggestion
   Note: Whether the incoming streams already have data on them will depend on server behavior. 
```

index.bs
```diff
@@ -994,11 +994,17 @@ these steps.
...
+
+   Note: whether the incoming streams already have data on them will depend on server behavior. 
+
    The getter steps for the `incomingBidirectionalStreams` attribute SHALL be:
      1. Return [=this=]'s {{[[IncomingBidirectionalStreams]]}}.
 : <dfn for="WebTransport" attribute>incomingUnidirectionalStreams</dfn>
 :: A {{ReadableStream}} of unidirectional streams, each represented by a
    {{WebTransportReceiveStream}}, that have been received from the server.
+
+   Note: whether the incoming streams already have data on them will depend on server behavior.
```
wilaw, line 1006:
```suggestion
   Note: Whether the incoming streams already have data on them will depend on server behavior. 
```

index.bs
```diff
@@ -1110,6 +1116,8 @@ the application will receive the number of streams it anticipates.
    stream.  Note that the mere creation of a stream is not immediately visible to the peer until
    it is used to send data.
 
+   Note: there is no expectation that the server will be aware of the stream until data is sent on it.
```
wilaw, line 1119:
```suggestion
   Note: There is no expectation that the server will be aware of the stream until data is sent on it.
```

index.bs
```diff
@@ -1148,6 +1156,8 @@ the application will receive the number of streams it anticipates.
    that the mere creation of a stream is not immediately visible to the server until it is used
    to send data.
 
+   Note: there is no expectation that the server will be aware of the stream until data is sent on it.
```
wilaw, line 1159:
```suggestion
   Note: There is no expectation that the server will be aware of the stream until data is sent on it.
```

Issue #610, opened 2024-07-29
Relative prioritization of streams and datagrams


afrind:
Media over QUIC recently merged a PR defining how moq will use priorities (https://github.com/moq-wg/moq-transport/pull/470).  The scheme is based on the MoQ object model (track, group, object) and is independent of how that object model is mapped to QUIC or WebTransport constructs (streams and datagrams).  It is possible for a MoQ application to specify a track that maps to a stream be given higher priority than a track that maps to datagrams.  Jan-Ivar presented at IETF 120 and mentioned that WebTransport now gives priority to datagrams over streams unilaterally, which is incompatible with moq's scheme.

wilaw:
There is a sendOrder and sendGroup scheme in the WebTransport API which would meet the MoQ requirements, but it only applies to Streams. One way to satisfy the MoQ requirement would be to also put flows of Datagrams into sendGroups of their own and then allow the application to assign the relative priority of those against other stream-based sendGroups. This has been discussed in #451 and #515 but after much discussion was never adopted.

vasilvv:
I do wonder if we should just give datagrams a sendGroup/sendOrder, that would probably be better than the current wording.  Maybe we could even have multiple datagram queues with different priorities.

jan-ivar:
> It is possible for a MoQ application to specify a track that maps to a stream be given higher priority than a track that maps to datagrams. 

What's the use case? I understand putting audio in datagrams and video in stream-of-streams, and giving audio priority.

> WebTransport now gives priority to datagrams over streams unilaterally, which is incompatible with moq's scheme.

I suspect it's incompatible even before https://github.com/w3c/webtransport/pull/604 when things were implementation-defined, but agree we may have to rethink.

Chrome sends all datagrams > all streams whereas Firefox sends all streams > all datagrams. Neither seems moq compatible.

What are moq's requirements?

> sendOrder

I think we're talking about weights here (between groups), not send ordering (within groups). 

> Maybe we could even have multiple datagram queues with different priorities.

Sounds like https://github.com/w3c/webtransport/issues/419.

I'd like to first explore what the incompatibility is. At least for deterministic priorities, we've assumed the app can control the total datagram to stream ratio by how many total datagrams it queues per second. User agents could in theory base their send-buffer dequeuing strategy on its assessment of how quickly datagram send-queues build up, compared with other send-queues.

[...]

PR #620, opened 2024-10-22
Add datagrams.createWritable()

jan-ivar:
Partial fix for https://github.com/w3c/webtransport/issues/610.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webtransport/pull/620.html" title="Last updated on Feb 11, 2025, 12:43 AM UTC (a1b22e0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/620/f2bfc64...jan-ivar:a1b22e0.html" title="Last updated on Feb 11, 2025, 12:43 AM UTC (a1b22e0)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -363,39 +363,19 @@ A {{WebTransportDatagramsWritable}} object has the following internal slot.
  </tbody>
 </table>
 
-The user agent MAY update {{[[OutgoingMaxDatagramSize]]}} for any {{WebTransport}} object whose
-{{[[State]]}} is either `"connecting"` or `"connected"`.
-
  To <dfn export for="WebTransportDatagramsWritable" lt="create|creating">create</dfn> a
- {{WebTransportDatagramsWritable}} given a
- <dfn export for="WebTransportDatagramsWritable/create"><var>readable</var></dfn>, and
- a <dfn export for="WebTransportDatagramsWritable/create"><var>writable</var></dfn>,
- perform the following steps.
+ {{WebTransportDatagramsWritable}}, perform the following steps.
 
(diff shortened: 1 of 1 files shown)


jan-ivar:
Meeting:
- This breaks backwards compatibility. Maybe leave `wt.datagrams.writable` alone for now.
- Editors can integrate once this is done.

jan-ivar:
This needs a bit more surgery to the algorithms. Will ping folks for review when that's done.

index.bs
```diff
@@ -332,15 +332,118 @@ A [=WebTransport stream=] has the following signals:
...
 
 A <dfn interface>WebTransportDatagramDuplexStream</dfn> is a generic duplex stream.
 
 <pre class="idl">
 [Exposed=(Window,Worker), SecureContext]
 interface WebTransportDatagramDuplexStream {
+  WebTransportDatagramsWritable createWritable();
   readonly attribute ReadableStream readable;
-  readonly attribute WritableStream writable;
+  readonly attribute WebTransportDatagramsWritable writable;
```
nidhijaju, line 446: I think it's confusing to have both `CreateWritable()` and `writable`. If we call `createWritable()` two times and then access `writable`, do we get the WritableStream from the first `createWritable()` call or a third instance of a WritableStream?

jan-ivar:
> It would be useful to add an Example to section 14 showing the use of datagrams.createWriteable(), perhaps juxtaposed against the default writeable, to show that both can be used simultaneously.

I think it would make sense to show examples of sendOrder and sendGroup, which is the use case for `createWritable()`.

So I might defer examples to the second PR for #610.

wilaw:
Overtaken by #629

PR #627, opened 2025-02-10
Fix some bikeshed errors

jan-ivar:


jan-ivar:
Stricter bikeshed updates are blocking PRs. Merging to unblock.

PR #628, opened 2025-02-11
Add datagrams.createWritable({sendGroup, sendOrder})

### LINKED ISSUE #610: Relative prioritization of streams and datagrams

jan-ivar:
Fixes #610 (builds on #620)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -363,39 +363,19 @@ A {{WebTransportDatagramsWritable}} object has the following internal slot.
  </tbody>
 </table>
 
-The user agent MAY update {{[[OutgoingMaxDatagramSize]]}} for any {{WebTransport}} object whose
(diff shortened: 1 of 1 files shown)


jan-ivar:
Abandoned due to rebase difficulty

PR #644, opened 2025-04-22
Add exportKeyingMaterial method

### LINKED ISSUE #411: expose a TLS Keying Material Exporter / unique value for the TLS session
It would be useful to expose a TLS Key Exporter ([RFC 5707](https://www.rfc-editor.org/rfc/rfc5705)), or at least a value that's derived from the TLS master secret, and therefore unique to the TLS session.

This is a useful building block for protocols running on top of TLS:
1. The (somewhat misleadingly named) [HTTP Transport Authentication](https://datatracker.ietf.org/doc/draft-schinazi-httpbis-transport-auth/07/) draft uses a TLS Keying Material Exporter. Note that I'm not suggesting to implement exactly this draft, just something along these lines.
2. In the p2p setting, peers might run a secondary handshake on top of a WebTransport session. Having access to a value unique to the TLS session would allow one to bind the that handshake to the TLS session.

nidhijaju:
Fixed #411.

***
<a href="https://pr-preview.s3.amazonaws.com/nidhijaju/webtransport/pull/644.html" title="Last updated on Apr 23, 2025, 2:59 AM UTC (4a714e3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/644/6db6cae...nidhijaju:4a714e3.html" title="Last updated on Apr 23, 2025, 2:59 AM UTC (4a714e3)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -713,6 +713,7 @@ interface WebTransport {
   constructor(USVString url, optional WebTransportOptions options = {});
 
   Promise&lt;WebTransportConnectionStats&gt; getStats();
+  [NewObject] Promise&lt;ArrayBuffer&gt; deriveKeyingExporter(BufferSource label, optional BufferSource context);
   readonly attribute Promise&lt;undefined&gt; ready;
   readonly attribute WebTransportReliabilityMode reliability;
   readonly attribute WebTransportCongestionControl congestionControl;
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1206,6 +1207,26 @@ the application will receive the number of streams it anticipates.
...
+   1. Let |contextLength| be 0.
+   1. If |context| is given, set |contextLength| to |context|.[=BufferSource/byte length=].
+   1. Let |p| be a new promise.
+   1. Run the following steps [=in parallel=], but [=abort when=] |transport|'s
+      {{[[State]]}} becomes `"closed"` or `"failed"`, and instead
+      [=queue a network task=] with |transport| to [=reject=] |p| with an {{InvalidStateError}}:
+     1. Let |keyingExporter| be the result of deriving a TLS keying exporter, as defined
+        in [[!WEB-TRANSPORT-HTTP3]] [Section 4.7](https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3/#section-4.7), with
+        |labelLength|, |label|, |contextLength|, and if present, |context|.
+     1. [=Resolve=] |p| with |keyingExporter|.
```
vasilvv, line 1227: We should probably reject if either length does not fit into uint8, since that's the length limitation in the spec.

index.bs
```diff
@@ -1206,6 +1207,26 @@ the application will receive the number of streams it anticipates.
            1. [=Resolve=] |p| with |stats|.
      1. Return |p|.
 
+: <dfn for="WebTransport" method>deriveKeyingExporter(BufferSource label, optional BufferSource context)</dfn>
```
martinthomson, line 1210: Is this the right name.  `exportX` would be better, I think.

index.bs
nidhijaju: Done

index.bs
nidhijaju: Updated the name, but I'll leave the type change determination to @vasilvv based on the discussion in the issue.

martenrichter:
How should the api behave in case of http/2? (It only refers to the http/3 spec).

nidhijaju:
> How should the api behave in case of http/2? (It only refers to the http/3 spec).

https://github.com/ietf-wg-webtrans/draft-ietf-webtrans-http2/issues/113 outlines the proposed behavior, but the PR for the HTTP/2 spec is still pending, so this will be updated once that lands.

jan-ivar:
Meeting:
- byte array is fine

Neustradamus:
@nidhijaju: Thanks for your PR and @martinthomson, @vasilvv, @jan-ivar for review and @jan-ivar for merging!

hamishwillee:
FYI, this is going into FF155. MDN docs are here: https://github.com/mdn/content/pull/45411

If any of you are able/interested I would very much appreciate a sanity check.

Issue #643, opened 2025-04-18
Writing zero length Uint8Array to stream, what is the expected behavior?


martenrichter:
Because of
https://github.com/fails-components/webtransport/issues/365#issuecomment-2814483261
I am asking what the correct way to handle it is. (So sending a chunk of zero length, I believe it is also a question for datagram as well as for the streams.)
I would now implement that I would just ignore the attempt?
Or should we error?

jan-ivar:
Since send streams don't preserve chunk boundaries anyway, I'd expect zero bytes to be added to the send queue, and for the operation to complete, without impact on what goes over the wire.

For datagrams the answer might be different. Can we send zero-payload datagrams? If so, that might be what I'd expect? If we don't want that then a simple check against zero in the writeDatagrams algorithm might be appropriate?

martenrichter:
> Since send streams don't preserve chunk boundaries anyway, I'd expect zero bytes to be added to the send queue, and for the operation to complete, without impact on what goes over the wire.

That's also what I assume. 

But there is an educational problem. When I started working on webtransport, I assumed that byte boundaries are preserved (since I used websocket before, which does preserve byte boundaries, and first tests with small data also preserved the byte boundaries).  Developers often assume that byte boundaries are presumed when posting issues, which is natural if someone comes from WebSocket.  (and some stick very hard to their assumption) And actually, also the example section of the W3C may suggest it.

When looking at W3C and IETF specs, it is not really written out that chunk boundaries are not preserved; it is merely the absence of a mechanism that conveys that message (or I have overlooked it). In the MDN docs, there is also no hint that it differs in this respect.
So it would be nice, if it could be written out explicitly  (I know that is not really the style of the spec.), so that I can point to it.

> 
> For datagrams the answer might be different. Can we send zero-payload datagrams? If so, that might be what I'd expect? If we don't want that then a simple check against zero in the writeDatagrams algorithm might be appropriate?

Good question, I do not see anything that should prevent it. I can write a test in my test harness against my code, Chromium, and Firefox, and see what happens.

martenrichter:
Ok, I have tested a zero length datagram with Firefox and Chromium, the datagram is send out and received.
But, I use to comply with the spec with a ByobReader supplied by node.js and this one throws:
```
TypeError [ERR_INVALID_STATE]: Invalid state: chunk ArrayBuffer is zero-length or detached
    at ReadableByteStreamController.enqueue (node:internal/webstreams/readablestream:1177:13)
[...]

PR #651, opened 2025-05-05
Don't send and ignore reception of zero-byte datagrams

jan-ivar:
Partial fix for #643.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -455,7 +455,7 @@ To <dfn>sendDatagrams</dfn>, given a {{WebTransport}} object |transport| and a
 1. Let |maxSize| be |datagrams|.{{[[OutgoingMaxDatagramSize]]}}.
 1. While |queue| is not empty:
   1. Let |bytes|, |timestamp| and |promise| be |queue|'s first element.
-  1. If |bytes|'s length ≤ |maxSize|:
+  1. If |bytes|'s length > 0 and ≤ |maxSize|:
     1. If it is not possible to send |bytes| to the network immediately, then break this loop.
     1. [=session/Send a datagram=], with |transport|.{{[[Session]]}} and |bytes|.
   1. Remove the first element from |queue|.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -455,7 +455,7 @@ To <dfn>sendDatagrams</dfn>, given a {{WebTransport}} object |transport| and a
 1. Let |maxSize| be |datagrams|.{{[[OutgoingMaxDatagramSize]]}}.
 1. While |queue| is not empty:
   1. Let |bytes|, |timestamp| and |promise| be |queue|'s first element.
-  1. If |bytes|'s length ≤ |maxSize|:
+  1. If |bytes|'s length > 0 and ≤ |maxSize|:
```
martinthomson, line 458: This appears to just skip datagrams if they are the wrong size.  Why is the promise not rejected?

jan-ivar: Are you referring to ≥ maxSize, zero or both?

Re ≥ maxSize that's how it is currently specified [and](https://searchfox.org/mozilla-central/rev/4c065f1df299065c305fb48b36cdae571a43d97c/testing/web-platform/tests/webtransport/datagrams.https.any.js#204-205,208) [implemented](https://wpt.fyi/results/webtransport/datagrams.https.any.html).

martinthomson: Both, in this case.  I'm not seeing that in this part of the specification.  (Good that it is implemented that way.)  Is that specified somewhere else?

This loop seems to iterate over the queue, sending any packet that is an OK size.  Once it can't send, it breaks out of the loop and removes the first item from the queue (not those that it sent, just the first one; it also doesn't  check that the first packet was sent or not).  I don't understand how that is even approximately correct.

jan-ivar: > Once it can't send, it breaks out of the loop and removes the first item from the queue

It breaks out of the loop, not the if-statement.

image

martinthomson: Eww.  Still, I guess that's OK.  We just silently drop things that are the wrong size.  Though I suspect that maybe this is wrong, now that we're talking about empty datagrams being valid.

alvestrand:
IETF WebTransport overview says that a stream is a sequence of bytes:

PR #655, opened 2025-05-19
Update to new echo-server (also polyfill MSTP+MSTG)

jan-ivar:


samples/webcodecs-echo/index.html
```diff
@@ -186,6 +186,8 @@ <h2>WebTransport Echo with WebCodecs in Worker + RVFC</h2>
 <button id="stop">Stop</button>
 <br/></br>
 <script src="https://cdn.plot.ly/plotly-2.18.2.min.js"></script>
+<script src="https://jan-ivar.github.io/polyfills/mediastreamtrackprocessor.js"></script>
+<script src="https://jan-ivar.github.io/polyfills/mediastreamtrackgenerator.js"></script>
```
wilaw, line 190: Can these polyfills be copied locally instead of referencing an external repo? If someone downloads the code and installs it, it will constantly be calling back to your repo, which is a brittle dependency.

jan-ivar: The same probably applies to `plot.ly`, but the tradeoff would be maintaining multiple copies of these things. For easier maintenance I think I prefer demos like this to use the canonical versions. As long as github doesn't go down I don't mind it hitting my repo.

PR #663, opened 2025-05-28
More robust camera selection in webcodecs-echo sample

jan-ivar:


jan-ivar:
I've addressed the "Critical Issues". The "Logical & Functional Issues" seem a mixed bag and unrelated to this PR (like applyConstraints on the single track we care about is fine and superior to calling gUM again). The valid ones (like getting rid of `decHwAcceleration`)  might be better addressed separately.

jan-ivar:
I want to update the webcodecs sample to work in Safari next, so I'm merging this one to unblock it. Hope that's OK. Thanks @wilaw for the review!

PR #667, opened 2025-06-18
API for Reliable Resets

### LINKED ISSUE #654: Add reliable reset stream to the API
We know the underlying transport protocol has the ability to use RESET_STREAM_AT, and this is a useful function which should also be exposed to applications.  MoQ, for example, has identified places where it's beneficial to use RESET_STREAM_AT compared to plain RESET_STREAM.

One convenient API to express RESET_STREAM_AT is stream.checkpoint(), which the transport can use to advance the reliable offset used when sending reset (defaults to 0).

martinthomson:
This also goes through and cleans up all the mess around RESET_STREAM. It's not perfect: we are now using that term (RESET_STREAM) to refer to a RESET_STREAM_AT frame (h3) or a WT_RESET_STREAM capsule (h2).

Closes #654.

***
<a href="https://pr-preview.s3.amazonaws.com/martinthomson/webtransport/pull/667.html" title="Last updated on Jul 1, 2025, 12:20 AM UTC (b32af85)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/667/2c7fa24...martinthomson:b32af85.html" title="Last updated on Jul 1, 2025, 12:20 AM UTC (b32af85)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -39,6 +39,13 @@ Boilerplate: omit conformance
     "status": "Proposed Standard",
     "publisher": "IETF"
   },
+  "reliable-reset": {
+    "authors": ["Marten Seemann", "奥一穂"],
+    "href": "https://datatracker.ietf.org/doc/html/draft-ietf-quic-reliable-stream-reset",
+    "title": "QUIC Stream Resets with Partial Delivery",
+    "status": "Internet-Draft",
+    "publisher": "IETF"
+  },
   "web-transport-overview": {
     "authors": ["Victor Vasiliev"],
     "href": "https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-overview",
@@ -48,14 +55,14 @@ Boilerplate: omit conformance
   },
   "web-transport-http3": {
     "authors": ["Alan Frindell", "Eric Kinnear", "Victor Vasiliev"],
-    "href": "https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3/",
+    "href": "https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3",
     "title": "WebTransport over HTTP/3",
     "status": "Internet-Draft",
     "publisher": "IETF"
   },
   "web-transport-http2": {
     "authors": ["Alan Frindell", "Eric Kinnear", "Tommy Pauly", "Martin Thomson", "Victor Vasiliev", "Guowu Xie"],
-    "href": "https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http2/",
+    "href": "https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http2",
     "title": "WebTransport over HTTP/2",
     "status": "Internet-Draft",
     "publisher": "IETF"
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2518,6 +2542,22 @@ interface WebTransportWriter : WritableStreamDefaultWriter {
...
+   This uses the mechanism described in [[RELIABLE-RESET]].
+
+   Note: This does not guarantee delivery in the event that a connection fails,
+   only when a stream is [=stream/reset=].
+
+   When {{commit}} is called for |stream|, the user agent MUST run the following steps:
+   1. Let |transport| be |stream|.{{WebTransportSendStream/[[Transport]]}}.
+   1. [=Queue a network task=] with |transport|
+      to set |stream|.{{WebTransportSendStream/[[CommittedOffset]]}}
+      to the value of |stream|.{{WebTransportSendStream/[[BytesWritten]]}}.
```
jan-ivar, line 2559: Queuing a task doesn't seem needed and only adds uncertainty (values may have updated by then).
```suggestion
   1. Set |stream|.{{WebTransportSendStream/[[CommittedOffset]]}}
      to the value of |stream|.{{WebTransportSendStream/[[BytesWritten]]}}.
```

jan-ivar (changes requested):
Thanks, this looks great.

We've been calling it `checkpoint()` in discussions but `commit()` WFM!

One fix needed I think

index.bs
martinthomson: Can we do that?  Maybe the web platform is able to play fast and loose with thread safety, but I figured that this is a case where could afford to have a single owner for the value.  If we pretend that all values are atomic, then maybe that works.  I'll leave this to you to resolve, since I don't care that much.

jan-ivar: For browser-initiated actions, queuing a task is essential to get on the JS event loop, but this being a method call we're already on it.

Both `[[CommittedOffset]]` and `[[BytesWritten]]` are JS-thread only, the latter incremented [here](https://pr-preview.s3.amazonaws.com/martinthomson/webtransport/pull/667.html#ref-for-dom-webtransportsendstream-byteswritten-slot%E2%91%A0) on the task that resolves the write promise, so it cannot change from under us (we're either in a different task, or more typically based on the desired `writer.write(x).then(() => writer.commit())` pattern, on the microtask queue being emptied at the tail of the same task that resolved the promise).

jan-ivar: Though (this reminded me) there's some prose still missing to get the the `[[CommittedOffset]]` safely from the JS thread to the in-parallel step where the reset happens. I think we need to add this to [abort a WebTransportSendStream](https://pr-preview.s3.amazonaws.com/martinthomson/webtransport/pull/667.html#webtransportsendstream-abort), and maybe `close` as well.

jan-ivar: Yes, it's all the same event loop. I imagine back in the days it was a single task queue as well, but someone decided this wasn't complicated enough, when input could be triaged into multiple [task sources](https://html.spec.whatwg.org/multipage/webappapis.html#task-source), by some taxonomy.

martinthomson:
I've added the committed offset to the abort algorithm as it should have had previously.  The "Reset" operation that it invokes doesn't have a clean API, so I wasn't able to trace that any further though.  Over to you.

jan-ivar:
Note https://github.com/w3c/webtransport/pull/669 is cleaning this will likely need to a rebase, but this should be enough to get `editors can integrate` tomorrow, so I'm happy to merge that after. Thanks again!

jan-ivar:
Meeting:
- JI to update algorithm to reset w/ CommittedOffset and merge

PR #689, opened 2025-09-09
Add connection stats bytesAcknowledged & payloadBytes(Sent|Acknowledged)

### LINKED ISSUE #674: More stats
We have `bytesSent` that excludes UDP framing, but includes QUIC framing.  On other hand, we have `estimatedSendRate`, which, as far as I can tell, excludes the the QUIC framing overhead.  It would be nice to provide `appBytes*` versions that only include application payload (i.e. contents of STREAM/DATAGRAM frames).

I also noticed we don't have `bytesAcked`; did we discuss this before?  Nominally `bytes_sent = bytes_acked + bytes_lost + bytes_in_flight`, so you can kinda guess this by guessing `bytes_in_flight`, but it's really fragile.

jan-ivar:
Fixes #674.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1704,6 +1704,9 @@ Note: Any unavailable stats will be [=map/exists|absent=] from the {{WebTranspor
 <pre class="idl">
 dictionary WebTransportConnectionStats {
   unsigned long long bytesSent = 0;
+  unsigned long long bytesAcknowledged = 0;
+  unsigned long long payloadBytesSent = 0;
+  unsigned long long payloadBytesAcknowledged = 0;
   unsigned long long packetsSent = 0;
   unsigned long long bytesLost = 0;
   unsigned long long packetsLost = 0;
@@ -1723,6 +1726,20 @@ The dictionary SHALL have the following attributes:
 : <dfn for="WebTransportConnectionStats" dict-member>bytesSent</dfn>
 :: The number of bytes sent on the [=underlying connection=], including retransmissions.
    Does not include UDP or any other outer framing.
+: <dfn for="WebTransportConnectionStats" dict-member>bytesAcknowledged</dfn>
+:: The number of bytes acknowledged as received by the server using
+   QUIC's ACK mechanism on the [=underlying connection=].
+   Does not include UDP or any other outer framing.
+
+   Note: Typically trails {{WebTransportConnectionStats/bytesSent}} but can be permanently less due to packet loss.
+
+: <dfn for="WebTransportConnectionStats" dict-member>payloadBytesSent</dfn>
+:: The number of payload bytes sent on the [=underlying connection=], including retransmissions.
+   Excludes [[!QUIC]] framing, UDP or any other outer framing.
+: <dfn for="WebTransportConnectionStats" dict-member>payloadBytesAcknowledged</dfn>
+:: The number of payload bytes acknowledged as received by the server using
+   QUIC's ACK mechanism on the [=underlying connection=].
+   Excludes [[!QUIC]] framing, UDP and any other outer framing.
 : <dfn for="WebTransportConnectionStats" dict-member>packetsSent</dfn>
 :: The number of packets sent on the [=underlying connection=], including those that are determined to have been lost.
 : <dfn for="WebTransportConnectionStats" dict-member>bytesLost</dfn>


index.bs
```diff
@@ -1723,6 +1726,20 @@ The dictionary SHALL have the following attributes:
...
+   QUIC's ACK mechanism on the [=underlying connection=].
+   Does not include UDP or any other outer framing.
+
+   Note: Typically trails {{WebTransportConnectionStats/bytesSent}} but can be permanently less due to packet loss.
+
+: <dfn for="WebTransportConnectionStats" dict-member>payloadBytesSent</dfn>
+:: The number of payload bytes sent on the [=underlying connection=], including retransmissions.
+   Excludes [[!QUIC]] framing, UDP and any other outer framing.
+: <dfn for="WebTransportConnectionStats" dict-member>payloadBytesAcknowledged</dfn>
+:: The number of payload bytes acknowledged as received by the server using
```
jan-ivar, line 1740:
```suggestion
: <dfn for="WebTransportConnectionStats" dict-member>applicationBytesSent</dfn>
:: The number of application bytes sent on the [=underlying connection=], including retransmissions.
   Excludes [[!QUIC]] framing, UDP and any other outer framing.
: <dfn for="WebTransportConnectionStats" dict-member>applicationBytesAcknowledged</dfn>
:: The number of application bytes acknowledged as received by the server using
```

index.bs
```diff
@@ -1723,6 +1726,20 @@ The dictionary SHALL have the following attributes:
 : <dfn for="WebTransportConnectionStats" dict-member>bytesSent</dfn>
 :: The number of bytes sent on the [=underlying connection=], including retransmissions.
    Does not include UDP or any other outer framing.
+: <dfn for="WebTransportConnectionStats" dict-member>bytesAcknowledged</dfn>
+:: The number of bytes acknowledged as received by the server using
+   QUIC's ACK mechanism on the [=underlying connection=].
+   Does not include UDP or any other outer framing.
```
martinthomson, line 1732:
```suggestion
:: The number of bytes acknowledged as received by the server using
   QUIC's ACK mechanism on the [=underlying connection=].
   This counts the outer size of QUIC packets,
   but does not include UDP, IP, or any other outer framing.
```

index.bs
```diff
@@ -1723,6 +1726,20 @@ The dictionary SHALL have the following attributes:
...
+
+   Note: Typically trails {{WebTransportConnectionStats/bytesSent}} but can be permanently less due to packet loss.
+
+: <dfn for="WebTransportConnectionStats" dict-member>payloadBytesSent</dfn>
+:: The number of payload bytes sent on the [=underlying connection=], including retransmissions.
+   Excludes [[!QUIC]] framing, UDP and any other outer framing.
```
martinthomson: This concept is a little fiddly and will require special accounting.

If you want to measure the number of bytes that the application was able to usefully send/deliver/whatever, then you need to deduct stream and datagram headers.  That's special WT-specific accounting, and useful, but you need to be a little more precise about what "payload bytes" (or my preferred "application bytes").

All that said, I do think that the retransmissions thing is much, much harder to account for.

You can define "application bytes" right here:

```suggestion
:: The number of application bytes sent on the [=underlying connection=].
   <dfn>Application bytes</dfn> is counted as the sum of
   the bytes sent on streams, excluding stream headers,
   and the bytes sent in datagrams, excluding any session identifiers.
   The count of application bytes <pick-one>does not | does</pick-one>
   count any retransmissions of stream data.
```

martinthomson: FWIW, I prefer to avoid counting retransmissions.  Accounting for retransmissions at this layer is much more intrusive on the implementation: a webtransport implementation is removed from the QUIC implementation that does retransmissions.  You'd need very tight integration to do accurate accounting of retransmissions, because some stream bytes cannot be counted and only the webtransport code really knows which bytes those are.  Given that this crosses multiple stack layers (WT > HTTP/3 > streams > retransmission) that's quite a burden.

I realize that this leads to a value that the application could measure for itself, mostly (datagram accounting isn't likely to work out perfectly with our API), but I think that the value is still very useful.  The QUIC packet byte accounting is a better measure of the link performance.

index.bs
```diff
@@ -1722,7 +1725,23 @@ The dictionary SHALL have the following attributes:
 
 : <dfn for="WebTransportConnectionStats" dict-member>bytesSent</dfn>
 :: The number of bytes sent on the [=underlying connection=], including retransmissions.
-   Does not include UDP or any other outer framing.
+   This counts the size of QUIC packets,
+   but does not include UDP or any other outer framing.
```
jan-ivar, line 1729:
```suggestion
   but does not include UDP, IP, or any other outer framing.
```

index.bs
```diff
@@ -1722,7 +1725,23 @@ The dictionary SHALL have the following attributes:
 
 : <dfn for="WebTransportConnectionStats" dict-member>bytesSent</dfn>
 :: The number of bytes sent on the [=underlying connection=], including retransmissions.
-   Does not include UDP or any other outer framing.
+   This counts the size of QUIC packets,
```
wilaw, line 1728:
```suggestion
   This includes the size of QUIC packets,
```

index.bs
```diff
@@ -1722,7 +1725,23 @@ The dictionary SHALL have the following attributes:
...
+   but does not include UDP, IP, or any other outer framing.
+: <dfn for="WebTransportConnectionStats" dict-member>bytesAcknowledged</dfn>
+:: The number of bytes acknowledged as received by the server using
+   QUIC's ACK mechanism on the [=underlying connection=].
+   This counts the size of QUIC packets,
```
wilaw, line 1733:
```suggestion
   This includes the size of QUIC packets,
```

jan-ivar:
Meeting:
- Use case: how much can I shove down this pipe?
- What is the cost in application bytes? (points to including retransmissions)
- need to estimate throughput accurately (points to not including retransmissions) 
- potentially a lot of cross-layer work / additional scope
- @vasilvv to check what we need for moq

nidhijaju (commented):
Maybe we already covered this in a meeting and I missed it, but why do we need the two different versions of bytesSent/bytesAcked if we're consistent about how we measure it and estimatedSendRate? Are they both separately useful somehow?

PR #709, opened 2025-12-09
s/datagramsReadableMode/datagramsReadableType/ and beef up note

### LINKED ISSUE #706: Is DatagramsReadableMode necessary and should it use byob as its default value?
DatagramsReadableMode allows to control whether a datagrams readable is either a byte stream or a regular stream.
This is, AFAIK, the only API that does this pattern on the web.

AIUI, the concern is that empty datagrams may be useful for some web applications but would not for many applications.
DatagramsReadableMode allows applications to decide whether to be able to get them or not.

This leads to the following questions:

1. Would it be possible to build a single stream that would cover both use cases.
Ideally, web applications caring about empty datagrams would use a default reader. Others would use a byob reader for efficiency.
This likely require changes to the streams spec (either via opt-in or for all byte sources maybe).
Has this been investigated/discussed?

2. If DatagramsReadableMode proves to be needed, what should be the default value?

AIUI, implementations diverge on the support: Chrome only supports byob sources, Firefox only supports default sources (do I get it right?).
If so, this probably means that applications running on both Chrome and Firefox do not care about empty datagrams.

If we think that most applications will not care about empty datagrams, we might want to create datagram byte streams by default.
This is probably more convenient and good for compat.

jan-ivar:
Fixes #706.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webtransport/pull/709.html" title="Last updated on Dec 16, 2025, 7:50 PM UTC (4f9d2cf)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/709/99aa333...jan-ivar:4f9d2cf.html" title="Last updated on Dec 16, 2025, 7:50 PM UTC (4f9d2cf)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -337,7 +337,7 @@ A {{WebTransportDatagramsWritable}} object has the following internal slot.
   </tr>
   <tr>
    <td><dfn>`[[Transport]]`</dfn>
-   <td class="non-normative">A {{WebTransport}} which owns this {{WebTransportDatagramsWritable}}.
+   <td class="non-normative">The {{WebTransport}} that owns this {{WebTransportDatagramsWritable}}.
   </tr>
   <tr>
    <td><dfn>`[[SendGroup]]`</dfn>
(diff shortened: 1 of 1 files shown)


martinthomson (approved):
I might have gone further about the suitability, but what you have is accurate.

PR #745, opened 2026-03-19
Update api.csswg.org/bikeshed URLs to use spec-generator

kfranqueiro:
This updates CONTRIBUTING.md, which referenced the discontinued api.csswg.org/bikeshed HTTP API, to instead reference https://www.w3.org/publications/spec-generator/.

I've verified that both updated commands work as expected.


martinthomson (approved):
die-on=nothing is a bit much, but I'm OK with this.  (I also typically add -Ssf to curl invocations for POST.)

PR #764, opened 2026-06-03
Gate pooling on a strong networkPartitionKey

### LINKED ISSUE #760: Pooling should require a strong secondKey in fetch's network partition key
This is a tracker for (in case it needs to be solved here instead):
- https://github.com/whatwg/fetch/issues/1926

Pooling is a feature inherited from [fetch](https://fetch.spec.whatwg.org/#concept-connection-obtain). But WebTransport is greenfield and must disallow cross-origin connection sharing.

If not solved in fetch, we could update the WT [constructor](https://w3c.github.io/webtransport/#webtransport-constructor) to ignore the [allowPooling](https://w3c.github.io/webtransport/#dom-webtransportoptions-allowpooling) option if [network partition key](https://fetch.spec.whatwg.org/#determine-the-network-partition-key)[1] is null.

jan-ivar:
Fixes #760.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webtransport/pull/764.html" title="Last updated on Jun 8, 2026, 5:11 PM UTC (4b8a56d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/764/826fe29...jan-ivar:4b8a56d.html" title="Last updated on Jun 8, 2026, 5:11 PM UTC (4b8a56d)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1051,7 +1051,8 @@ To <dfn export>obtain a WebTransport connection</dfn>, given a [=network partiti
   1. Let |url| be |request|'s [=request/current URL=].
   1. Let |newConnection| be |transport|.{{[[NewConnection]]}}.
   1. If |newConnection| is "`no`" and |networkPartitionKey|[1] is not unique to
-     |request|'s [=request/origin=], then set |newConnection| to "`yes-and-dedicated`".
+     |request|'s [=request/origin=], then set |newConnection| and
+     |transport|.{{[[NewConnection]]}} to "`yes-and-dedicated`".
 
      Note: The [=network partition key=]'s second key is [=implementation-defined=].
      It is <em>unique to</em> an [=origin=] only when the user agent guarantees that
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1049,6 +1049,9 @@ To <dfn export>obtain a WebTransport connection</dfn>, given a [=network partiti
   1. Let |transport| be the {{WebTransport}} object associated with |request|.
   1. Let |url| be |request|'s [=request/current URL=].
   1. Let |newConnection| be |transport|.{{[[NewConnection]]}}.
+  1. If |newConnection| is "`no`" and |networkPartitionKey|[1] is null or a value that
+     is not unique to |request|'s [=request/origin=], then set |newConnection| to
```
martinthomson: This "not unique to" test, is that documented somewhere?

jan-ivar: For _"unique"_? No.

secondKey is _"[null or an implementation-defined value](https://fetch.spec.whatwg.org/#determine-the-network-partition-key)"_, so we need to specify some stricter minimum requirement.

The challenge is doing so without being too lax or too prescriptive. For example, checking against `null` seems insufficient, since a value of 1 would be just as terrible as null.

Requiring a value that is unique per origin seems a decent approximation of what's bandied about in https://github.com/whatwg/fetch/issues/1035. Not sure how to best word it.

jan-ivar:
Abandoned over https://github.com/w3c/webtransport/issues/760#issuecomment-4724593858.

PR #775, opened 2026-07-02
Remove redundant optionality check for exportKeyingMaterial's context arg

### LINKED ISSUE #774: exportKeyingMaterial: IDL marks context required, but the prose treats it as optional (and WebIDL then forces outputLength optional too)
The IDL for `WebTransport.exportKeyingMaterial()` declares all three arguments as required:

```webidl
[NewObject] Promise<Uint8Array> exportKeyingMaterial(
    BufferSource label, BufferSource context, unsigned long outputLength);
```

But the algorithm prose treats `context` as **optional**. From
<https://w3c.github.io/webtransport/#dom-webtransport-exportkeyingmaterial>:

> 3. Let *contextLength* be 0.
> 4. **If *context* is given**, set *contextLength* to *context*.[byte length].

"If context is given" only makes sense if `context` is an optional argument. As the IDL is
written, `BufferSource context` has no `optional` keyword and is therefore mandatory, so it is
always "given" — step 3 and the conditional in step 4 are dead code.

This is a self-contained inconsistency within the spec; it does not depend on any test or
implementation to demonstrate.

### This also makes the IDL invalid once `context` is fixed

WebIDL does not allow a required argument to follow an optional one — see the operation grammar
and <https://webidl.spec.whatwg.org/#dfn-optional-argument>. So the moment `context` is marked
`optional` to match step 4, `outputLength` (which follows it) must be `optional` as well.
There is no valid WebIDL in which `context` is optional and `outputLength` is required.

### Requested change

Mark both trailing arguments `optional`:

```webidl
[NewObject] Promise<Uint8Array> exportKeyingMaterial(
    BufferSource label,
    optional BufferSource context,
    optional unsigned long outputLength);
```

- `context` should be `optional` — directly matching step 4's "If context is given".
- `outputLength` must then also be `optional` (WebIDL rule above). The prose should additionally
  define the default used when `outputLength` is omitted — e.g. an explicit "if outputLength is
  not given, set it to N" step — since step 6 currently only rejects `outputLength` of 0 or
  above the implementation cap, and says nothing about the omitted case.

Alternatively, if the WG's intent is genuinely that all three arguments are required, then
step 4's "If context is given" is wrong and should be dropped (context always present), and the
default-`outputLength` question is moot. Either way the current text is internally inconsistent.

### Context

`outputLength` was introduced in #756; the `optional`/default question does not appear to have
been settled there. Firefox currently implements
`exportKeyingMaterial(BufferSource label, optional BufferSource context)` (it does not yet
implement `outputLength`), which is why it fails the idlharness `.length` subtest:

> WebTransport interface: operation exportKeyingMaterial(BufferSource, BufferSource, unsigned long)

That failure is driven entirely by the required-argument count in the IDL above.

jan-ivar:
Fixes #774.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webtransport/pull/775.html" title="Last updated on Jul 2, 2026, 9:49 PM UTC (8519d79)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/775/fff8e1b...jan-ivar:8519d79.html" title="Last updated on Jul 2, 2026, 9:49 PM UTC (8519d79)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1330,8 +1330,7 @@ the application will receive the number of streams it anticipates.
 
    1. Let |labelLength| be |label|.[=BufferSource/byte length=].
    1. If |labelLength| is more than 255, return [=a promise rejected with=] a {{RangeError}}.
-   1. Let |contextLength| be 0.
-   1. If |context| is given, set |contextLength| to |context|.[=BufferSource/byte length=].
+   1. Let |contextLength| be |context|.[=BufferSource/byte length=].
    1. If |contextLength| is more than 255, return [=a promise rejected with=] a {{RangeError}}.
    1. If |outputLength| is 0 or more than an [=implementation-defined=] value--
        which [=must=] be at least 4096--


jesup (approved):

jan-ivar:
Merging ahead of CfC

PR #785, opened 2026-08-31
Reword the notes on atomic writes

martinthomson:
Originally, this was just an attempt to fix some bad grammar, but I took it a bit further.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2724,16 +2724,16 @@ interface WebTransportWriter : WritableStreamDefaultWriter {
    [Section 4.4](https://datatracker.ietf.org/doc/html/rfc9308#section-4.4)).
 
    Note: {{atomicWrite}} can still reject after sending some data. Though it
-   provides atomicity with respect to flow control, other errors may occur.
-   {{atomicWrite}} does not prevent data from being split between packets
-   or being interleaved with other data. Only the sender learns if
-   {{atomicWrite}} fails due to lack of available flow control credit.
+   provides atomicity with respect to flow control, other network errors could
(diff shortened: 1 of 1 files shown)


martinthomson:
[Build error reason](https://lists.w3.org/Archives/Public/spec-prod/2026JulSep/0002.html)

index.bs
```diff
@@ -2724,16 +2724,16 @@ interface WebTransportWriter : WritableStreamDefaultWriter {
    [Section 4.4](https://datatracker.ietf.org/doc/html/rfc9308#section-4.4)).
 
    Note: {{atomicWrite}} can still reject after sending some data. Though it
-   provides atomicity with respect to flow control, other errors may occur.
-   {{atomicWrite}} does not prevent data from being split between packets
-   or being interleaved with other data. Only the sender learns if
-   {{atomicWrite}} fails due to lack of available flow control credit.
+   provides atomicity with respect to flow control, other network errors could
+   cause the write to fail.
+   Nor does {{atomicWrite}} prevent data from being split between packets
```
wilaw:
```suggestion
Furthermore,  {{atomicWrite}} does not prevent data from being split between packets
```

jan-ivar (approved):
With nit

index.bs
martinthomson: was the removal of the indent intentional?

```suggestion
   Furthermore, {{atomicWrite}} does not prevent data from being split between packets
```

wilaw: No, thanks for catching.

nidhijaju (approved):

PR #786, opened 2026-08-31
Wrap the exportKeyingMaterial definition

martinthomson:
Because this ends up needing horizontal scrolling otherwise and that is never a good thing.


martinthomson:
[Build error reason](https://lists.w3.org/Archives/Public/spec-prod/2026JulSep/0002.html)

jan-ivar (approved):

nidhijaju (approved):

PR #790, opened 2026-09-08
Resolve datagram backpressure when sink queue has room

### LINKED ISSUE #787: writeDatagrams only resolves back pressure once all buffered datagrams have been sent
[writeDatagrams](https://w3c.github.io/webtransport/#writedatagrams) is written to buffer datagrams in its sink (_"[in another process](https://pr-preview.s3.amazonaws.com/jan-ivar/webtransport/pull/784.html#dom-webtransportdatagramduplexstream-outgoingmaxbuffereddatagrams)"_), by waving datagrams through until filled:

> 9. If the length of writable.[[[OutgoingDatagramsQueue]]](https://w3c.github.io/webtransport/#dom-webtransportdatagramswritable-outgoingdatagramsqueue-slot) is less than datagrams.[[[OutgoingMaxBufferedDatagrams]]](https://w3c.github.io/webtransport/#dom-webtransportdatagramduplexstream-outgoingmaxbuffereddatagrams-slot), then [resolve](http://www.ecma-international.org/ecma-262/6.0/index.html#sec-promise-objects) promise with undefined.

But it appears to have a bug:

It only resolves back pressure (calls for more) once the last buffered _"[datagram is sent or discarded](https://w3c.github.io/webtransport/#dom-webtransportdatagramswritable-outgoingdatagramsqueue-slot)"_, rather than after 1 datagram is sent or discarded.

This seems like a mistake if the goal is to keep the send buffer full.

For example:
```js
wt.datagrams.outgoingMaxBufferedDatagrams = 10;
await pullProducer.pipeTo(wt.datagrams.createWritable());
```
This would only pull from `pullProducer` to replenish the send buffer once the 10th datagram has been sent or discarded.

Instead, we probably want to replenish the send buffer as soon as any datagram has been sent or discarded.

jan-ivar:
Fixes #787.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/webtransport/pull/790.html" title="Last updated on Sep 8, 2026, 10:56 PM UTC (eced479)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webtransport/790/681b520...jan-ivar:eced479.html" title="Last updated on Sep 8, 2026, 10:56 PM UTC (eced479)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -349,8 +349,13 @@ A {{WebTransportDatagramsWritable}} object has the following internal slot.
  <tbody>
   <tr>
    <td><dfn>`[[OutgoingDatagramsQueue]]`</dfn>
-   <td class="non-normative">A queue of tuples of an outgoing datagram, a timestamp and a promise
-   which is resolved when the datagram is sent or discarded.
+   <td class="non-normative">A queue of tuples of an outgoing datagram and a timestamp.
+  </tr>
+  <tr>
+   <td><dfn>`[[BackpressurePromise]]`</dfn>
+   <td class="non-normative">A promise which is resolved once
+   {{WebTransportDatagramsWritable/[[OutgoingDatagramsQueue]]}} has room for another datagram,
+   or null.
   </tr>
   <tr>
    <td><dfn>`[[Transport]]`</dfn>
@@ -375,6 +380,8 @@ A {{WebTransportDatagramsWritable}} object has the following internal slot.
  1. Let |stream| be a [=new=] {{WebTransportDatagramsWritable}}, with:
     : {{WebTransportDatagramsWritable/[[OutgoingDatagramsQueue]]}}
     :: an empty queue
+    : {{WebTransportDatagramsWritable/[[BackpressurePromise]]}}
+    :: null
     : {{WebTransportDatagramsWritable/[[Transport]]}}
     :: |transport|
     : {{WebTransportDatagramsWritable/[[SendGroup]]}}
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -448,19 +477,19 @@ with |transport| to run the following steps:
 1. If |duration| is null, then set |duration| to an [=implementation-defined=] value.
 1. Run the following steps [=in parallel=]:
   1. While |queue| is not empty:
-    1. Let |bytes|, |timestamp| and |promise| be |queue|'s first element.
+    1. Let |bytes| and |timestamp| be |queue|'s first element.
     1. If more than |duration| milliseconds have passed since |timestamp|, then:
        1. Remove the first element from |queue|.
-       1. [=Queue a network task=] with |transport| to [=resolve=] |promise| with undefined.
+       1. [=Dequeue an outgoing datagram=] with |transport| and |writable|.
```
nidhijaju, line 483: We call `Remove the first element from |queue|`, and then `Dequeue an outgoing datagram` calls `[=queue/Dequeue=]` a second time, so wouldn't this cause two elements to be removed from `[[OutgoingDatagramsQueue]]` for every single datagram sent or expired, dropping every alternate datagram?

index.bs
```diff
@@ -419,13 +426,19 @@ The <dfn>writeDatagrams</dfn> algorithm is given a |transport| and |writable| as
...
-1. Enqueue |chunk| to |writable|.{{[[OutgoingDatagramsQueue]]}}.
+1. Let |chunk| be a tuple of |bytes| and |timestamp|.
+1. [=queue/Enqueue=] |chunk| to |writable|.{{[[OutgoingDatagramsQueue]]}}.
 1. If the length of |writable|.{{[[OutgoingDatagramsQueue]]}} is less than
-   |datagrams|.{{[[OutgoingMaxBufferedDatagrams]]}}, then [=resolve=] |promise| with undefined.
-1. Return |promise|.
+   |datagrams|.{{[[OutgoingMaxBufferedDatagrams]]}}, then return
+   [=a promise resolved with=] undefined.
+1. Assert: |writable|.{{[[BackpressurePromise]]}} is null.
+1. Set |writable|.{{[[BackpressurePromise]]}} to a new promise.
```
nidhijaju, line 436: If the session closes or the stream is aborted while `[[BackpressurePromise]]` is pending, wouldn't clearing `[[OutgoingDatagramsQueue]]` leave the `[[BackpressurePromise]]` promise pending? Should we ensure `[[BackpressurePromise]]` is rejected or reset in the cleanup and abort algorithms so awaiting JS callers do not hang?

jan-ivar: This seems like an existing issue so I'll file a separate issue on this.

index.bs
```diff
@@ -436,6 +449,22 @@ timestamp and the expiration duration work well only when the web developer pays
...
 <div algorithm>
 
+To <dfn>dequeue an outgoing datagram</dfn>, given a {{WebTransport}} object |transport| and a
+{{WebTransportDatagramsWritable}} object |writable|, [=queue a network task=] with |transport|
+to run these steps:
+1. If |writable|.{{[[OutgoingDatagramsQueue]]}} is empty, then return.
+1. [=queue/Dequeue=] |writable|.{{[[OutgoingDatagramsQueue]]}}.
+1. If |writable|.{{[[BackpressurePromise]]}} is null, then return.
+1. If the length of |writable|.{{[[OutgoingDatagramsQueue]]}} is not less than
+   |transport|.{{[[Datagrams]]}}.{{[[OutgoingMaxBufferedDatagrams]]}}, then return.
```
jan-ivar, line 459: We might need to look at what happens when the max changess up or down

index.bs
jan-ivar: It's done this way because _queue_  [here](https://pr-preview.s3.amazonaws.com/jan-ivar/webtransport/pull/790.html#senddatagrams) is a copy, so we need to dequeue from both the copy and the original.

But perhaps we don't need it to be a copy....

jan-ivar:
Editors can integrate pending closing of threads!

index.bs
martinthomson: Two options:

1. Remove this step here.  Remove the assertion in the sending code.
2. Keep this step.  Change the code above to check the queue depth and resolve+clear the promise if the queue has space, *before* enqueuing anything.

Either way, changes to the queue length will result in the promise being present when sending.  The first option is simpler, but it means that a queue size increase is not immediately available until the network takes a datagram.  The second makes the space immediately available.  Probably not hugely consequential, but the second is slightly more correct.

index.bs
jan-ivar: We do need the copy, since it's accessed in parallel in the steps below. But I'll make it more clear with some better names.

index.bs
jan-ivar: I decided to address it here instead of filing a new issue. The algorithm is redesigned to never clear [[BackpressurePromise]] and to resolve it in cleanup (which is the only place we clear [[OutgoingDatagramsQueue]]). PTAL

Regarding `datagramWritable.abort()`, the streams spec design is that abort() waits until the sink has finished consuming a chunk, so its behavior seems intentional. If network drops then cleanup happens and we resolve [[BackpressurePromise]]. If network doesn't drop but flow control somehow never opens then it might delay abort until a datagram ages out. This seems by design.

index.bs
```diff
@@ -422,51 +428,77 @@ The <dfn>writeDatagrams</dfn> algorithm is given a |transport| and |writable| as
...
-       1. Remove the first element from |queue|.
-       1. [=Queue a network task=] with |transport| to [=resolve=] |promise| with undefined.
-    1. Otherwise, break this loop.
-  1. If |transport|.{{[[State]]}} is not `"connected"`, then return.
-  1. While |queue| is not empty:
-    1. Let |bytes|, |timestamp| and |promise| be |queue|'s first element.
+       1. [=queue/Dequeue=] |datagrams|.
+       1. Set |count| to |count| + 1.
+    1. Otherwise, [=iteration/break=] this loop.
+  1. If |count| is greater than 0, then [=dequeue outgoing datagrams=] with |transport|,
```
saschanaz: Should this dequeueing also happen in the network task to resolve the promise?

jan-ivar:
PR preview is back!

Also I've made the algorithm less restrictive to allow UA tuning.

Updating backpressure less often lets the send queue drain before the {{WritableStream}} is asked for more datagrams. Update it often enough to keep the queue full.

